PR #3, opened 2020-02-20
Improve toolchain installation help messaging

phamann:
### TL;DR
Improves the UX of the remediation error messages for Rust toolchain installation.

### Why?
We've noticed in various user testing sessions that the current text wasn't explicit enough and didn't differentiate the command output from the other help text.  It also seems a previous refactor to how remediation messages are printed (via `text.Output()`) was causing the whitespace in the (`\n\t` etc) to be stripped.

### Previous output:
Screenshot 2020-02-20 at 16 58 33

### New output:
Screenshot 2020-02-20 at 16 36 15


peterbourgon:
Will that command work reliably on all target systems?

PR #5, opened 2020-03-09
Interactive init

phamann:
### TL;DR
Refactors `compute init` to accept its input interactively and act as a setup wizard to provision a service, backend and domain.

### Why?
We have observed friction in user testing and feedback that a users has to create a service then attach a domain and backend before a compute package can be deployed and activated. If you are new to Fastly this may be confusing. Therefore, we think that provisioning all of this for you within a single command will reduce this on-boarding friction.

### What?
- Refactor command to take input interactively
- Refactor `fastly configure` to persist user email to global config file so that it can be read by `compute init` for default author email.
- Add a `NullProgress` writer which swallows all writes
- Add an `UndoStack` which we use to teardown any state we have created in the API if an error occurs.
- Add [`golang-petname`](https://github.com/dustinkirkland/golang-petname) dep to randomly generate nice domain names

### Example:
Screenshot 2020-03-09 at 12 23 28

### DIFF
--- a/pkg/common/file.go
+++ b/pkg/common/file.go
@@ -55,7 +55,7 @@ func CopyFile(src, dst string) (err error) {
 	if err != nil {
 		return fmt.Errorf("error reading source file: %w", err)
 	}
-	defer in.Close()
+	defer in.Close() // #nosec G307
 
 	// Create destination file for writing.
 	out, err := os.Create(dst)
--- a/pkg/common/undo.go
+++ b/pkg/common/undo.go
@@ -1,0 +1,43 @@
+package common
+
+// UndoFn is a function with no arguments which returns an error or nil.
+type UndoFn func() error
+
+// UndoStack models a simple undo stack which consumers can use to store undo
+// stateful functions, such as a function to teardown API state if something
+// goes wrong during procedural commands, for example deleting a Fastly service
+// after it's been created.
+type UndoStack struct {
+	states []UndoFn
+}
+
+// NewUndoStack constructs a new UndoStack.
+func NewUndoStack() *UndoStack {
+	s := make([]UndoFn, 0, 1)
+	stack := &UndoStack{
+		states: s,
+	}
+	return stack
+}
+
+// Pop method pops last added UndoFn element oof the stack and returns it.
+// If stack is empty Pop() returns nil.
+func (s *UndoStack) Pop() UndoFn {
+	n := len(s.states)
+	if n == 0 {
+		return nil
+	}
+	v := s.states[n-1]
+	s.states = s.states[:n-1]
+	return v
+}
+
+// Push method pushes an Undoer element onto the UndoStack.
+func (s *UndoStack) Push(elem UndoFn) {
+	s.states = append(s.states, elem)
+}
+
+// Len method returns the number of elements in the UndoStack.
+func (s *UndoStack) Len() int {
+	return len(s.states)
+}
--- a/pkg/common/undo.go
+++ b/pkg/common/undo.go
@@ -1,5 +1,10 @@
 package common
 
+import (
+	"fmt"
+	"io"
+)
+
 // UndoFn is a function with no arguments which returns an error or nil.
 type UndoFn func() error
 
@@ -40,4 +45,22 @@ func (s *UndoStack) Push(elem UndoFn) {
 // Len method returns the number of elements in the UndoStack.
 func (s *UndoStack) Len() int {
 	return len(s.states)
+}
+
+// RunIfError unwinds the stack if a non-nil error is passed, by serially
+// calling each UndoFn function state in FIFO order. If any UndoFn returns an
+// error, it gets logged to the provided writer. Should be deferrerd, such as:
+//
+//     undoStack := common.NewUndoStack()
+//     defer undoStack.RunIfError(w, err)
+//
+func (s *UndoStack) RunIfError(w io.Writer, err error) {
+	if err == nil {
+		return
+	}
+	for i := len(s.states) - 1; i >= 0; i-- {
+		if err := s.states[i](); err != nil {
+			fmt.Fprintf(w, "%w", err)
+		}
+	}
 }
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -95,21 +95,14 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		// Use a null progress writer whilst gathering input.
 		progress = text.NewNullProgress()
 	}
-
-	undoStack := common.NewUndoStack()
-
 	defer func() {
 		if err != nil {
 			progress.Fail() // progress.Done is handled inline
-			// Unwind the undo stack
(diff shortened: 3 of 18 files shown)


pkg/common/file.go
```diff
@@ -55,7 +55,7 @@ func CopyFile(src, dst string) (err error) {
 	if err != nil {
 		return fmt.Errorf("error reading source file: %w", err)
 	}
-	defer in.Close()
+	defer in.Close() // #nosec G307
```
peterbourgon: 🙄 IMO disable this check entirely. (Minor nit, not blocking this PR.)

pkg/common/undo.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Push method pushes an Undoer element onto the UndoStack.
+func (s *UndoStack) Push(elem UndoFn) {
+	s.states = append(s.states, elem)
+}
+
+// Len method returns the number of elements in the UndoStack.
+func (s *UndoStack) Len() int {
+	return len(s.states)
+}
```
peterbourgon: If you are going to go through the (appreciated!) trouble of defining an UndoStack type, IMO you should go all the way and also provide a helper method to execute it in case of error.

```go
func (s *UndoStack) RunIfError(err error) {
	if err == nil {
		return
	}
	for i := len(s.states)-1; i >= 0; i-- {
		s.states[i]()
	}
}
```

That means you could reduce the public API to merely `Push` and `RunIfError`, probably remove the UndoFn type and use plain `func()`, and use it in-situ like

```go
func f() (err error) {
	var undo common.UndoStack
	defer undo.RunIfError(err)
	
	undo.Push(func() { ... })
```

pkg/compute/init.go
```diff
@@ -4,32 +4,59 @@ import (
...
+
+const (
+	defaultTemplate       = "https://github.com/fastly/fastly-template-rust-default.git"
+	defaultTopLevelDomain = "edgecompute.app"
+	manageServiceBaseURL  = "https://manage.fastly.com/configure/services/"
+)
 
 var (
+	gitRepositoryRegEx        = regexp.MustCompile(`((git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?`)
+	domainNameRegEx           = regexp.MustCompile(`(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]`)
```
peterbourgon: I am afraid this is incomplete, but I don't have the strength/knowledge/intelligence to test it. Does it come from some known source? Is it possible to be abused?

pkg/compute/init.go
```diff
@@ -75,19 +127,143 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+		}
+		if c.author == "" {
+			c.author = defaultEmail
+		}
+	}
+
+	if c.from == "" {
+		text.Output(out, "%s", text.Bold("Template:"))
+		for i, template := range defaultTemplates {
+			text.Output(out, "[%d] %s", i, template.Name)
```
peterbourgon: IMO it would be nice to put the URL of the template in parenthesis here, so that the " . . . or type URL" portion of the prompt is visually reinforced.

pkg/compute/init.go
```diff
@@ -75,19 +127,143 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	_, err = c.Globals.Client.CreateBackend(&fastly.CreateBackendInput{
+		Service: service.ID,
+		Version: 1,
+		Name:    c.backend,
+		Address: c.backend,
+	})
+	if err != nil {
+		return fmt.Errorf("error creating backend: %w", err)
+	}
+	undoStack.Push(func() error {
```
peterbourgon: If you want undo functions to return an error — which is sensible — then logging each non-nil error when executing the undo stack would probably be the best bet.

pkg/compute/init.go
```diff
@@ -134,22 +310,39 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
 
 	progress.Done()
-	text.Success(out, "Initialized package %s to %s", m.Name, abspath)
+
+	text.Break(out)
+
+	fmt.Fprintf(out, "Initialized package %s to:\n\t%s\n\n", text.Bold(m.Name), text.Bold(abspath))
+	fmt.Fprintf(out, "Manage this service at:\n\t%s\n\n", text.Bold(fmt.Sprintf("%s%s", manageServiceBaseURL, service.ID)))
+	fmt.Fprintf(out, "To compile the package, run:\n\t%s\n\n", text.Bold("fastly compute build"))
+	fmt.Fprintf(out, "To deploy the package, run:\n\t%s\n\n", text.Bold("fastly compute deploy"))
```
peterbourgon: Is the idiom of emitting "output" from a larger process by doing one or more

> Prompt:
>&nbsp;&nbsp;&nbsp;&nbsp;**Output**

blocks at the conclusion of the process done elsewhere in the CLI? If so, is there an opportunity to make this style consistent?

pkg/text/progress.go
```diff
@@ -234,3 +234,33 @@ func (p *VerboseProgress) Done() {}
...
 // Fail implements the Progress interface. It's a no-op.
 func (p *VerboseProgress) Fail() {}
+
+//
+//
+//
+
+// NullProgress is an implementation of Progress which discards everything
+// written into it and produces no output.
+type NullProgress struct{}
```
peterbourgon: 🤘

pkg/common/undo.go
peterbourgon: If the undo funcs returned an error, RunIfError could take an io.Writer to which error results could be printed.

phamann: Added this helper method in https://github.com/fastly/cli/pull/5/commits/d4b73e21737f74615747e148b2902b1ccaca7ff2

pkg/compute/init.go
phamann: It seemed the most robust in the top SO response for this question, however I agree with you that it could be brittle and I'm sad the Go standard library doesn't offer something such as how you can parse and validate IP addresses.

pkg/compute/init.go
phamann: Added error logging as part of d4b73e2

pkg/compute/init.go
phamann: Added a `text.Description()` helper in https://github.com/fastly/cli/pull/5/commits/97bdd8200a757642fb81ed9ee0ca1a8ece6a0b2a

pkg/compute/init.go
phamann: Added in https://github.com/fastly/cli/pull/5/commits/78fd8c262bf2a19bd64543fc003bdbbff37d04b7

pkg/common/file.go
phamann: Will do in another PR and discuss there.

PR #8, opened 2020-03-18
Add useful next step links to compute deploy

phamann:
### TL;DR
Adds additional links to the output of `compute deploy`, to manage the service via the UI and the first domain to view.

### Why?
It's often come up in user testing sessions that the user doesn't want to do next after deploying or know the test domain to `curl` etc. This is an attempt to reduce that friction, it follows the same convention and uses the same output patterns that we introduced in #5 

### Example:
Screenshot 2020-03-17 at 15 31 46

### DIFF
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -145,7 +145,7 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		Service: serviceID,
 		Version: c.version,
 	}); err == nil {
-		text.Description(out, "View the service at", fmt.Sprintf("https://%s", domains[0].Name))
+		text.Description(out, "View this service at", fmt.Sprintf("https://%s", domains[0].Name))
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/compute/deploy.go
```diff
@@ -136,6 +136,18 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+
+	text.Break(out)
+
+	text.Description(out, "Manage this service at", fmt.Sprintf("%s%s", manageServiceBaseURL, serviceID))
+
+	if domains, err := c.Globals.Client.ListDomains(&fastly.ListDomainsInput{
+		Service: serviceID,
+		Version: c.version,
+	}); err == nil {
+		text.Description(out, "View the service at", fmt.Sprintf("https://%s", domains[0].Name))
```
peterbourgon:
```suggestion
		text.Description(out, "View this service at", fmt.Sprintf("https://%s", domains[0].Name))
```

peterbourgon: (Just for consistency)

PR #9, opened 2020-03-19
Add commands for S3 logging endpoints

phamann:
### TL;DR
Adds CRUD commands for Amazon S3 logging endpoints.

### DIFF
--- a/pkg/logging/bigquery/bigquery_test.go
+++ b/pkg/logging/bigquery/bigquery_test.go
@@ -115,7 +115,7 @@ func TestBigQueryList(t *testing.T) {
 	}
 }
 
-func TestBiogQueryDescribe(t *testing.T) {
+func TestBigQueryDescribe(t *testing.T) {
 	for _, testcase := range []struct {
 		args       []string
 		api        mock.API
--- a/pkg/logging/s3/create.go
+++ b/pkg/logging/s3/create.go
@@ -1,0 +1,84 @@
+package s3
+
+import (
+	"io"
+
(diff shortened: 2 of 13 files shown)


pkg/logging/bigquery/bigquery_test.go
```diff
@@ -115,7 +115,7 @@ func TestBigQueryList(t *testing.T) {
 	}
 }
 
-func TestBiogQueryDescribe(t *testing.T) {
+func TestBigQueryDescribe(t *testing.T) {
```
peterbourgon: 😅

pkg/logging/s3/create.go
```diff
@@ -0,0 +1,84 @@
...
+	case string(fastly.S3RedundancyReduced):
+		c.Input.Redundancy = fastly.S3RedundancyReduced
+	}
+
+	switch c.serverSideEncryption {
+	case string(fastly.S3ServerSideEncryptionAES):
+		c.Input.ServerSideEncryption = fastly.S3ServerSideEncryptionAES
+	case string(fastly.S3ServerSideEncryptionKMS):
+		c.Input.ServerSideEncryption = fastly.S3ServerSideEncryptionKMS
+	}
```
peterbourgon: Is the zero value of these fields meaningful when the Input goes to the Fastly API?

phamann: Are you asking due to the lack of a `default` case? The empty string zero value is correct/valid in the API if no value is passed via user.

PR #10, opened 2020-03-25
Fix up undoStack.RunIfError

jeddenlea:
First, with Go 1.14, the use of `%w` in a Printf context other than Errorf
causes a vet error.  `%w` is a special trigger to Errorf to create a
"wrapped" error.  What you'd want to do instead is either `%s` or `%v`.
However, given the nature of error strings, the behavior of simply
concatenating them together to `w` seems inappropriate.  If anything,
they should at least occur on seperate lines, so Fprintln seems more
appropriate.

Second, RunIfError's comment and invocation in Exec made apparent
another bug.  As written, the undo stack would never actually execute.
As directed, if you were to:

	defer undoStack.RunIfError(out, err)

...every value is evaluated at the point of `defer`, but the execution of
the function is what is actually deferred.  `err` will be `nil` there,
and thus the undo stack is never actually executed.

This simple fix there is to actually defer a closure.  When an error is
returned, it will then be observable from within the closure.

### DIFF
--- a/pkg/common/undo.go
+++ b/pkg/common/undo.go
@@ -52,7 +52,7 @@ func (s *UndoStack) Len() int {
 // error, it gets logged to the provided writer. Should be deferrerd, such as:
 //
 //     undoStack := common.NewUndoStack()
-//     defer undoStack.RunIfError(w, err)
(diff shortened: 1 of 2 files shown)


jeddenlea:
To be replaced....

PR #14, opened 2020-04-06
Add the ability to initialise a compute project from a specific branch

phamann:
### TL;DR
As requested by @katie-martin-fastly, this adds the ability to clone from a specific template repository branch when initialising a new Compute@Edge project via `compute init`. We then use this logic to ensure that the default template is pointing to a named branch. This ensures we maintain compatibility with the `fastly` guest crate API.

### Note
- I'm keeping the `--branch` flag hidden for the time being as we don't want to add confuse to the normal path of supplying templates repo urls. 
- I also cleaned up the `compute init` tests and fixed a rendering issue with author, but are small enough changes that don't warrant their own PR.

### DIFF
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -21,6 +21,7 @@ import (
 	"github.com/fastly/cli/pkg/text"
 	"github.com/fastly/go-fastly/fastly"
 	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
 type template struct {
@@ -30,6 +31,7 @@ type template struct {
 
 const (
 	defaultTemplate       = "https://github.com/fastly/fastly-template-rust-default.git"
+	defaultTemplateBranch = "0.2.0-alpha3"
 	defaultTopLevelDomain = "edgecompute.app"
 	manageServiceBaseURL  = "https://manage.fastly.com/configure/services/"
 )
@@ -54,6 +56,7 @@ type InitCommand struct {
 	description string
 	author      string
 	from        string
+	fromBranch  string
 	path        string
 	domain      string
 	backend     string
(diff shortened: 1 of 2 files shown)


pkg/compute/init.go
```diff
@@ -68,6 +71,7 @@ func NewInitCommand(parent common.Registerer, globals *config.Data) *InitCommand
 	c.CmdClause.Flag("description", "Description of the package").Short('d').StringVar(&c.description)
 	c.CmdClause.Flag("author", "Author of the package").Short('a').StringVar(&c.author)
 	c.CmdClause.Flag("from", "Git repository containing package template").Short('f').StringVar(&c.from)
+	c.CmdClause.Flag("from-branch", "Git branch name to clone from package template repository").Hidden().StringVar(&c.fromBranch)
```
peterbourgon, line 74: This feels like it overloads `--from`. What about `--branch` or `--remote-branch`?

phamann: Resolved in https://github.com/fastly/cli/pull/14/commits/541d1419a04dcd60e00d52a7294036342380db06

pkg/compute/init.go
```diff
@@ -259,10 +263,19 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	}
 	defer os.RemoveAll(tempdir)
 
+	var ref plumbing.ReferenceName
+	if c.from == defaultTemplate {
+		ref = plumbing.NewBranchReferenceName(defaultTemplateBranch)
+	}
+	if c.branch != "" {
+		ref = plumbing.NewBranchReferenceName(c.branch)
+	}
```
peterbourgon, line 272: Does `ref` need a non-blank default?

phamann: I tested this and read a bit of the go-git source, without you get the HEAD reference which was the same behaviour before this change.

PR #15, opened 2020-04-07
Normalize backend name during compute init

phamann:
### TL;DR
To reduce on-boarding friction `compute init` only asks for a backend address during the interactive initialisation, we then use the same value for both the `address` and `name` fields in the [API call](https://docs.fastly.com/api/config#backend_85c170418ee71191dbb3b5046aeb6c2c) to create the backend. However, we have since discovered these are not valid backend `name` values and therefore we need to normalize the address input before creation. 

### Notes
Backend names must match this regular expression (`^[a-zA-Z][a-zA-Z0-9_]*$`), i.e. they myst conform to these rules:
- cannot start with a number.
- can only contain alpha numeric ASCII characters or underscores.

If the input address is an IP address and therefore starts with a number, we prefix the string with "F_" which follows historic naming convention for backends in Fastly VCL services.

### DIFF
--- a/pkg/compute/compute_test.go
+++ b/pkg/compute/compute_test.go
@@ -887,6 +887,31 @@ func TestUploadPackage(t *testing.T) {
 	}
 }
 
+func TestNormalizeBackendName(t *testing.T) {
+	for _, testcase := range []struct {
+		input      string
+		wantOutput string
+	}{
+		{
+			input:      "www.example.com",
+			wantOutput: "www_example_com",
+		},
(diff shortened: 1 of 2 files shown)


phamann:
>If the normalized name is still invalid for whatever reason, the CreateBackend call will fail, correct?

Sadly this is not the case and I've raised with the downstream API team (cc @thommahoney)

phamann:
Closing as this issue has been resolved upstream.

PR #17, opened 2020-04-15
Use GoReleaser for build and release automation in CI

phamann:
### TL;DR
Replaces our custom make targets and CI GitHub actions that build and release binaries with [GoReleaser](https://goreleaser.com/). This greatly reduces complexity whilst increasing possibilities for future customisation. 

### Notes
We're planning to phase the roll out of this in these steps.
1. Replace current release and checksumming process with GoReleaser
1. Use goreleaser to generate a changelog.md
1. Use goreleaser to sign packages with gpg key
1. Use goreleaser to produce homebrew recipe
1. Use goreleaser to produce other package manager packages
1. Use goreleaser to signed and notarize MacOS binaries 

This PR is the first step :)

### DIFF
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -1,114 +1,31 @@
+name: Release
 on:
   push:
     tags:
       - 'v*'
-name: Release
 jobs:
-  release:
+  goreleaser:
     runs-on: ubuntu-latest
     steps:
-    - name: Create GitHub release
-      id: create_release
-      uses: actions/create-release@v1
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        tag_name: ${{ github.ref }}
-        release_name: ${{ github.ref }}
-        draft: true
-        prerelease: false
-    - name: Output upload URL to file
-      env:
-        UPLOAD_URL: ${{ steps.create_release.outputs.upload_url }}
-      run: |
-        echo $UPLOAD_URL > upload_url.txt
-    - name: Persist release upload URL
-      uses: actions/upload-artifact@v1
-      with:
-        name: release
-        path: upload_url.txt
-  upload:
-    needs: release
-    strategy:
-      matrix:
-        pair: [linux-amd64, linux-386, darwin-amd64, windows-amd64, windows-386]
-    runs-on: ubuntu-latest
-    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Unshallow
+      run: git fetch --prune --unshallow
     - name: Install Go
       uses: actions/setup-go@v1
       with:
         go-version: '1.14.x'
-    - name: Checkout code
-      uses: actions/checkout@v1
-    - name: Get version
+    - name: Set GOVERSION
       run: |
-        echo ::set-env name=VERSION::${GITHUB_REF/refs\/tags\//}
-    - name: Build release binaries
-      env:
-        PAIR: ${{ matrix.pair }}
-      shell: bash
+        echo "::set-env name=GOVERSION::$(go version)"
+    - name: Set GIT_TAG
       run: |
-        GOOS=$(echo $PAIR | cut -d'-' -f1)
-        GOARCH=$(echo $PAIR | cut -d'-' -f2)
-        if [[ $GOOS =~ "windows" ]]
-        then
-          env VERSION=$VERSION GOOS=$GOOS GOARCH=$GOARCH make release-exe
-        else
-          env VERSION=$VERSION GOOS=$GOOS GOARCH=$GOARCH make release
-        fi
-    - name: Get upload URL
-      uses: actions/download-artifact@v1
+        echo "::set-env name=GIT_TAG::${GITHUB_REF/refs\/tags\//}"
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v1
       with:
-        name: release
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -41,20 +41,6 @@ build:
...
-	go build -o fastly $(LDFLAGS) ./cmd/fastly
-	tar czf fastly_${VERSION}_$(shell go env GOOS)-$(shell go env GOARCH).tar.gz fastly
-	rm fastly
-
-.PHONY: release-exe
-release-exe:
-	go build -o fastly.exe $(LDFLAGS) ./cmd/fastly
-	tar czf fastly_${VERSION}_$(shell go env GOOS)-$(shell go env GOARCH).tar.gz fastly.exe
-	rm fastly.exe
+	GOVERSION=$(shell go version) goreleaser
```
mccurdyc, line 46: I'm curious if you ran this make target, would it duplicate a tag? Once from this goreleaser process and then once from the GitHub Action which is triggered on a tag push?

mccurdyc: Nevermind. I think I see what's going on. The GitHub Action will run the release on `v*` tags automatically, but the make target would be for a manual release that presumably wouldn't be tagged `v*`.

phamann: Good question, I mainly left this in for consistency with our previous setup and as you suggest to have a way to manually run the release if needed. Though I can see how it may add confusion, so may just remove.

phamann: Have now removed :+1:

.github/workflows/tag_release.yml
```diff
@@ -1,114 +1,28 @@
...
-    strategy:
-      matrix:
-        pair: [linux-amd64, linux-386, darwin-amd64, windows-amd64, windows-386]
+  goreleaser:
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Unshallow
+      run: git fetch --prune --unshallow
```
thommahoney, line 13: TIL `--unshallow` 👍

.github/workflows/tag_release.yml
```diff
@@ -1,114 +1,28 @@
...
-      with:
-        upload_url: ${{ env.UPLOAD_URL }}
-        asset_path: ./bin/fastly_${{ env.VERSION }}_SHA256SUMS
-        asset_name: fastly_${{ env.VERSION }}_SHA256SUMS
-        asset_content_type: text/plain
+        echo "::set-env name=GOVERSION::$(go version)"
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v1
+      with:
+        version: latest
```
thommahoney, line 24: any concern about goreleaser being unstable? I guess time will tell.

.github/workflows/tag_release.yml
```diff
@@ -1,114 +1,28 @@
...
+        args: release --rm-dist
```
thommahoney, line 25: We we adding `--rm-dist` as a safety mechanism? (I see that the dir should already be empty since we are gitignoring it)

.goreleaser.yml
```diff
@@ -0,0 +1,47 @@
+project_name: fastly
+release:
+  prerelease: auto
+before:
+  hooks:
+    - go mod tidy
```
thommahoney, line 6: I totally understand why we want to do this during the release (to make sizes smaller) but it seems like our test suite should run this and fail the build if there are extraneous dependencies. Thoughts?

.goreleaser.yml
```diff
@@ -0,0 +1,47 @@
...
+    - go mod download
+builds:
+  - <<: &build_defaults
+      main: ./cmd/fastly
+      ldflags:
+        - -s -w -X "github.com/fastly/cli/pkg/version.AppVersion={{ .Version }}"
+        - -X "github.com/fastly/cli/pkg/version.GitRevision={{ .ShortCommit }}"
+        - -X "github.com/fastly/cli/pkg/version.GoVersion={{ .Env.GOVERSION }}"
```
thommahoney, line 14: Are these standardized / defined somewhere else or are these parameters defined on these lines? Do you think it would be valuable to annotate the build number here so that we can find logs easier?

Edit: there's no such thing as a build number in GHA 😲

.github/workflows/tag_release.yml
phamann: This was copy/pasta from their GitHub actions docs. It doesn't really matter either way as its running in an ephemeral CI job and will get discarded after.

.goreleaser.yml
phamann: These are defined and used inside the `fastly version` command and are used to populate the output. See here for more detail: https://github.com/fastly/cli/blob/master/pkg/version/root.go#L11-L67

.github/workflows/tag_release.yml
phamann: This is the version of our repository that goreleaser will use as the release version, not the version of goreleaser itself. Sadly, the action doesn't allow you to pin to a version, which was flagged during our security audit.  

`latest` here means use the latest semver-like tag. However we could be more robust here and pass in the tag that triggered the build.

.goreleaser.yml
phamann: I like the idea of ensuring we have a clean go.mod and go.sum file via a CI test :+1:

.github/workflows/tag_release.yml
phamann: Have now updated to use the git tag which triggered the build.

PR #23, opened 2020-04-17
Add CHANGELOG.md and automate change log generation

phamann:
### TL;DR
This adds a `CHANGELOG.md` to the root of the repository which documents the change log for the project for each semver release we create. It also adds steps to automate the generation of said CHANGELOG.md (using [github_changelog_generator](https://github.com/github-changelog-generator/github-changelog-generator)) during the release CI workflow.

### Notes
- We've favoured to use [github_changelog_generator](https://github.com/github-changelog-generator/github-changelog-generator) as the automation tool, as it uses labels we apply to issues and PRs to categorise the log. Such as `enhancement`, `bug` etc. Which we prefer over semantic-release style structured commit messages. 
- In order to automate the generation the CI job needs to commit back the new changes to the master branch via CI. I'm still unsure on my thoughts on this. 
- We generate a smaller RELEASE_CHANGELOG.md file with only the changes for the current release and pass that to `goreleaser` to use as the release notes. This ensures consistency accross the two logs.


.github/workflows/tag_release.yml
```diff
@@ -18,11 +18,37 @@ jobs:
...
+    - name: Commit CHANGELOG.md
+      env:
+        GITHUB_TOKEN: ${{ secrets.github_token }}
+        COMMIT_MSG: |
+          Release ${GIT_TAG}
+          skip-checks: true
+      run: |
+        git config user.email "patrick@fastly.com"
+        git config user.name "phamann"
+        git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
```
phamann, line 51: I don't like having to hardcode a git user here, even though we are using the action time-limited token to actually perform the write here. Is there another service-account style Fastly git user we could use instead of me?

Makefile
```diff
@@ -1,5 +1,7 @@
 SHELL := /bin/bash -o pipefail
 
+LAST_SEMVER_TAG := $(shell git tag | sort -r --version-sort | egrep '^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$$' | head -n1)
```
thommahoney, line 3:
```suggestion
LAST_SEMVER_TAG := $(shell git tag | sort -r --version-sort | egrep '^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$' | head -n1)
```

Makefile
```diff
@@ -40,3 +42,27 @@ build:
...
+.PHONY:
+changelog:
+	github_changelog_generator -u fastly -p cli \
+		--no-pr-wo-labels \
+		--no-author \
+		--enhancement-label "**Enhancements:**" \
+		--bugs-label "**Bug fixes:**" \
+		--release-url "https://github.com/fastly/cli/releases/tag/%s" \
+		--exclude-labels documentation \
+		--exclude-tags-regex "v.*-.*"
```
thommahoney, line 55: This is for "pre" releases and the like?

.github/workflows/tag_release.yml
```diff
@@ -18,11 +18,37 @@ jobs:
...
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Set GIT_TAG
+      run: |
+        echo "::set-env name=GIT_TAG::${GITHUB_REF/refs\/tags\//}"
+    - name: Commit CHANGELOG.md
+      env:
+        GITHUB_TOKEN: ${{ secrets.github_token }}
+        COMMIT_MSG: |
+          Release ${GIT_TAG}
+          skip-checks: true
```
thommahoney, line 47: Does this work? The [Github docs](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-status-checks#skipping-and-requesting-checks-for-individual-commits) indicate that there need to be two newlines before this line.

.github/workflows/tag_release.yml
```diff
@@ -18,11 +18,37 @@ jobs:
...
+      run: |
+        git config user.email "patrick@fastly.com"
+        git config user.name "phamann"
+        git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+        git checkout master
+        git add .
+        git diff --quiet && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}"; git push origin master)
```
thommahoney, line 54: The master branch is protected as of this review comment. How do we plan to allow the CHANGELOG.md file to be updated with these settings in place?

.github/workflows/tag_release.yml
```diff
@@ -18,11 +18,37 @@ jobs:
...
+        GITHUB_TOKEN: ${{ secrets.github_token }}
+        COMMIT_MSG: |
+          Release ${GIT_TAG}
+          skip-checks: true
+      run: |
+        git config user.email "patrick@fastly.com"
+        git config user.name "phamann"
+        git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+        git checkout master
+        git add .
```
thommahoney, line 53: It'd be safer to specifically `add` the files that we expect to have changed instead of staging all changes.

Makefile
phamann: This is required makefile escaping for dollar signs, i.e. it actually results in a single dollar.

Makefile
phamann: Correct, we only want to document proper releases.

.github/workflows/tag_release.yml
phamann: I've tested commit logic on a test branch with branch protection enabled (https://github.com/fastly/cli/commit/864eb8aefb3ca991acdb0b16d22508a92968a742) and it seemed to work. However, this may have only worked because the token I created could have assumed my admin privileges for the repo. It's also not that clear from the actions docs as to whether the generated `GITHUB_TOKEN` assumes admin privileges: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret so we may have to test this in CI to determine the behaviour.

thommahoney: Worst-case scenario we just need to fix it. Let's move forward.

.github/workflows/tag_release.yml
```diff
@@ -18,11 +18,39 @@ jobs:
     - name: Set GOVERSION
       run: |
         echo "::set-env name=GOVERSION::$(go version)"
+    - name: Install Ruby
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: '2.7'
+    - name: Install github_changelog_generator
+      run: gem install github_changelog_generator
```
thommahoney, line 26: It might be worth using Bundler with a `Gemfile.lock` to keep the `github_changelog_generator` pinned to a particular version. Alternatively, installing a specific version on this line with `-v` will keep it stable. The only concern here is that the output format may break the consumers of the [*_]CHANGELOG.md files.

PR #25, opened 2020-04-20
compute/init: make space after Author prompt match other prompts

shawnps:
Other prompts have a space after the closing bracket:

Screen Shot 2020-04-20 at 11 50 09 AM


phamann:
Thanks for the contribution @shawnps. Please ignore the failing security `scan` CI check. This is only because you don't have write access and therefore the token isn't present in the build. We are going to move this check to the master branch soon, for this reason.

PR #26, opened 2020-04-20
Publish to homebrew

phamann:
### TL;DR
Adds the required goreleaser configuration to publish to a Homebrew tap formula to https://github.com/fastly/homebrew-tap during the release CI workflow. This allows users to install the CLI via `brew` using `brew install fastly/tap/fastly`. 

### Notes:
- We can submit the forumla to https://github.com/Homebrew/homebrew-core to facilitate the `brew install fastly` default style naming and invocation install.
- We will need to create a wider scoped GITHUB_TOKEN secret for the CI job to enable the job to push to the tap repo.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -35,6 +35,29 @@ archives:
     <<: *archive_defaults
     wrap_in_directory: false
     format: zip
+brews:
+  - name: fastly
+    ids: [nix]
+    github:
+      owner: fastly
+      name: homebrew-tap
+    skip_upload: auto
+    description: Fastly CLI
+    homepage: https://github.com/fastly/cli
+    folder: Formula
+    custom_block: |
+      head do
+        url "https://github.com/fastly/cli.git"
+        depends_on "go"
+      end
+    install: |
+      system "make" if build.head?
+      bin.install "fastly"
+      (bash_completion/"fastly.sh").write `#{bin}/fastly --completion-script-bash`
+      (zsh_completion/"_fastly").write `#{bin}/fastly --completion-script-zsh`
+    test: |
+      help_text = shell_output("#{bin}/fastly --help")
+      assert_includes help_text, "Usage:"
 checksum:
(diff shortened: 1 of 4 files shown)


.goreleaser.yml
```diff
@@ -39,6 +39,29 @@ archives:
...
+      owner: fastly
+      name: homebrew-tap
+    skip_upload: auto
+    description: Fastly CLI
+    homepage: https://github.com/fastly/cli
+    folder: Formula
+    custom_block: |
+      head do
+        url "https://github.com/fastly/cli.git"
+        depends_on "go"
```
thommahoney, line 55: we also depend on a working `git` binary. unsure if we should declare it as part of the homebrew formula.

phamann: Couple of things to clarify:
- We use [`go-git`](https://github.com/go-git/go-git) which is a pure Go implementation of git and not the local `git` binary. 
- This block is only used if the user is installing the `HEAD` and not a specific version, during which the binary is compiled via Go, its not a dependency in normal operation as we download prebuilt release assets.

jfoote:
Thanks for flagging this @phamann! Two comments on your notes:

> We can submit the forumla to https://github.com/Homebrew/homebrew-core to facilitate the brew install fastly default style naming and invocation install.

^ I'm not sure this will work. From the [GoReleaser docs](https://goreleaser.com/homebrew/):

> Important: Note that GoReleaser does not generate a valid homebrew-core formula. The generated formulas are meant to be published as homebrew taps, and in their current form will not be accepted in any of the official homebrew repositories.

Regarding secrets:

> We will need to create a wider scoped GITHUB_TOKEN secret for the CI job to enable the job to push to the tap repo.

Per the [github docs](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token), `GITHUB_TOKEN` access is limited to the repo the app is installed in. Would you use a personal access token instead? Please feel free to follow up in a private channel if you feel your response here might be sensitive.

PR #28, opened 2020-04-21
Don't check for updates when running local builds

pteichman:
This removes the update portion of TestVersion in pkg/version,
which appears to be tested well enough in pkg/update instead.


pteichman:
@thommahoney Yeah, I think that's covered well enough in the tests in pkg/update. It's less of a full app integration test but handles that case.

phamann:
@thom yup: https://github.com/fastly/cli/blob/master/pkg/update/check_test.go#L106

PR #37, opened 2020-04-23
Fix data race in fastly configure

phamann:
### TL;DR
Fixes a data race introduced in #30 by ensuring we call `progress.Fail()` if we error.

### DIFF
--- a/pkg/configure/root.go
+++ b/pkg/configure/root.go
@@ -40,6 +40,8 @@ func NewRootCommand(parent common.Registerer, configFilePath string, cf APIClien
 
 // Exec implements the command interface.
 func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
+	var err error
+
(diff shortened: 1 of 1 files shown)


pkg/configure/root.go
```diff
@@ -39,7 +39,7 @@ func NewRootCommand(parent common.Registerer, configFilePath string, cf APIClien
 }
 
 // Exec implements the command interface.
-func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
+func (c *RootCommand) Exec(in io.Reader, out io.Writer) (err error) {
```
pteichman, line 42: The `var err error` on line 68 now shadows this error variable, so that case will not trigger progress.Fail().

I generally don't like named return variables unless they support documentation readability from afar, and `error` is so ubiquitous in Go that that isn't really necessary. If it's just here for the side effect of getting the variable declared before the defer(), the intent here may be clearer as a `var err error` at the top of the function.

PR #41, opened 2020-04-24
Fix release asset inconsistency

phamann:
### TL;DR
Following on from #39, we discovered another inconsistency between the old release asset format and the new one. `fastly update` assumes the binary is in the root of the tar ball and not in a directory. Whilst the directory is a preferred user experience, this breaks the upgrade flow and will have to wait until a major release to fix. 

### Note 
This PR also attempts to fix the changelog generation by force pushing the commit ot bypass branch protection.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -28,7 +28,9 @@ archives:
     builds: [macos, linux]
     <<: &archive_defaults
       name_template: "{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
-    wrap_in_directory: true
+      files:
+        - none*
+    wrap_in_directory: false
     format: tar.gz
   - id: windows
     builds: [windows]


.github/workflows/tag_release.yml
```diff
@@ -46,15 +46,11 @@ jobs:
...
-          skip-checks: true
+        COMMIT_MSG: Release ${GIT_TAG}
       run: |
         git config user.email "oss@fastly.com"
         git config user.name "Release bot"
         git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
         git checkout master
         git add CHANGELOG.md
-        git diff --quiet && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}"; git push origin master)
+        git diff --quiet && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}"; git push --force origin master)
```
thommahoney, line 56: this has the potential of accidentally wiping out merges to master that happen in quick succession. I'm not sure that the `--force` flag actually bypasses branch protections either.

.goreleaser.yml
```diff
@@ -28,7 +28,9 @@ archives:
     builds: [macos, linux]
     <<: &archive_defaults
       name_template: "{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
-    wrap_in_directory: true
+      files:
+        - none*
```
thommahoney, line 32: lol I just read [the docs](https://goreleaser.com/archive/#packaging-only-the-binaries) about this... at least they are acknowledging that it's a hack.

PR #45, opened 2020-04-27
Publish scoop package manifest during release process

phamann:
### TL;DR
Publishes an app manifest for the Windows [Scoop](https://scoop.sh/) package manager during the release process. The manifest is published to a Fastly owned public repo: [github.com/fastly/scoop-cli](https://github.com/fastly/scoop-cli).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -68,6 +68,14 @@ nfpms:
     formats:
       - deb
       - rpm
+scoop:
+  bucket:
+    owner: fastly
+    name: scoop-cli
+  homepage: https://github.com/fastly/cli
+  skip_upload: auto
+  description: Fastly CLI
+  license: Apache 2.0
 checksum:
   name_template: "{{ .ProjectName }}_v{{ .Version }}_SHA256SUMS"
 snapshot:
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -60,6 +60,14 @@ brews:
...
       help_text = shell_output("#{bin}/fastly --help")
       assert_includes help_text, "Usage:"
+scoop:
+  bucket:
+    owner: fastly
+    name: scoop-cli
+  homepage: https://github.com/fastly/cli
+  skip_upload: auto
+  description: Fastly CLI
+  license: Apache 2.0
```
thommahoney, line 70: https://github.com/fastly/scoop-cli/pull/1/files#diff-04c6e90faac2675aa89e2176d2eec7d8R29 says "MIT"

.goreleaser.yml
```diff
@@ -60,6 +60,14 @@ brews:
     test: |
       help_text = shell_output("#{bin}/fastly --help")
       assert_includes help_text, "Usage:"
+scoop:
+  bucket:
+    owner: fastly
+    name: scoop-cli
+  homepage: https://github.com/fastly/cli
+  skip_upload: auto
```
thommahoney, line 68: this setting seems to only apply to homebrew, according to the docs, though perhaps they are missing this field for scoop.

.goreleaser.yml
thommahoney: nvm this License is for fastly/cli itself, obviously.

PR #48, opened 2020-04-29
Update release instructions

thommahoney:
As part of the 0.7.0 release, I noticed that the release instructions assume a working development environment as well as a GPG-configured `git`. I didn't have either ahead of the release so this PR updates the RELEASE.md with additional instructions.

Certainly open to linking elsewhere for the development environment dependencies instead of listing them explicitly in RELEASE.md.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,13 @@ fastly:
 	@go build -trimpath $(LDFLAGS) -o "$@" ./cmd/fastly
 
 .PHONY: all
-all: fmt vet staticcheck lint gosec test build install
+all: dependencies fmt vet staticcheck lint gosec test build install
+
+.PHONY: dependencies
+dependencies:
+	go get -v -u github.com/securego/gosec/cmd/gosec
+	go get -v -u honnef.co/go/tools/cmd/staticcheck
+	go get -v -u golang.org/x/lint/golint
 
 .PHONY: fmt
 fmt:
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,11 @@ ifndef CHANGELOG_GITHUB_TOKEN
 	@echo "WARNING: No \$$CHANGELOG_GITHUB_TOKEN in environment, set one to avoid hitting rate limit."
 	@echo ""
 endif
+ifeq ($(SEMVER_TAG),)
+	$(error "You must set $$SEMVER_TAG to your desired release semver version.")
+endif
 	github_changelog_generator -u fastly -p cli \
+		--future-release $(SEMVER_TAG) \
 		--no-pr-wo-labels \
 		--no-author \
 		--enhancement-label "**Enhancements:**" \
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ fastly:
 	@go build -trimpath $(LDFLAGS) -o "$@" ./cmd/fastly
 
 .PHONY: all
-all: dependencies fmt vet staticcheck lint gosec test build install
+all: fmt vet staticcheck lint gosec test build install
 
 .PHONY: dependencies
 dependencies:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,23 +6,23 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 1. Merge all PRs intended for the release into the `master` branch
 1. Checkout and update the master branch and ensure all tests are passing:
-	* `git checkout master`
-	* `git pull`
+    * `git checkout master`
+    * `git pull`
     * `make all`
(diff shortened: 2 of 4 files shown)


RELEASE.md
```diff
@@ -6,23 +6,27 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 1. Merge all PRs intended for the release into the `master` branch
 1. Checkout and update the master branch and ensure all tests are passing:
-	* `git checkout master`
-	* `git pull`
+    * `git checkout master`
+    * `git pull`
+    * install and update necessary dependencies:
+        * go get -v -u honnef.co/go/tools/cmd/staticcheck
+        * go get -v -u golang.org/x/lint/golint
+        * go get -v -u github.com/securego/gosec/cmd/gosec
```
phamann, line 14: As you mention, some options here:
- Add a "prerequisites" section the development section of the `README.md` which details these and we link to here.
- Alternatively we could have a Make target `install-dependencies` or similar would checks for existence of each binary and installs if not present and add to `make all`.

RELEASE.md
```diff
@@ -6,23 +6,27 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
     * `git add CHANGELOG.md && git commit -m "vX.X.X"`
 1. Send PR for the `CHANGELOG.md`
 1. Once approved and merged, checkout and update the `master` branch:
-	* `git checkout master`
-	* `git pull`
+    * `git checkout master`
+    * `git pull`
 1. Create a new tag for `master`:
-	* `git tag -s vx.x.x -m "vx.x.x"`
+    * `git tag vx.x.x -m "vx.x.x"`
```
phamann, line 27: Signed commits for the tag releases was something which was advised by our internal security team as a good best practice for releasing. Personally i'd prefer it if we adhered to this. So maybe we should add a comment to the top suggesting to configure git correctly?

RELEASE.md
thommahoney: I went with the second option, and replaced the installation steps in the GHA workflow to `make dependencies` instead. Let me know what you think.

RELEASE.md
thommahoney: I removed this commit from the branch 👍

Makefile
```diff
@@ -13,7 +13,13 @@ fastly:
...
 
 .PHONY: all
-all: fmt vet staticcheck lint gosec test build install
+all: dependencies fmt vet staticcheck lint gosec test build install
+
+.PHONY: dependencies
+dependencies:
+	go get -v -u github.com/securego/gosec/cmd/gosec
+	go get -v -u honnef.co/go/tools/cmd/staticcheck
+	go get -v -u golang.org/x/lint/golint
```
phamann, line 22: There are some issues installing the tooling binaries this way:
- As the project is a go module, these commands will update the `go.mod` and `go.sum` manifests and therefore leaving git in a dirty state. 
- If we don't first assert whether the binaries are already on the $PATH, they have the potential to overwrite an existing binary, which some developers may not be happy about. 

Unfortunately, there isn't a easy/clean way of installing a global binary in a Go modules world. The recommended practice for this is something along the lines of:
1. cd to a directory that is (a) not a subdirectory of any directory with a go.mod file, and (b) not in GOPATH
2. go get my/tool/url
3. mv $(go env GOPATH)/bin/tool $HOME/bin
4. cd back to the original directory

In practice, step 1 is usually `cd /tmp`. One issue with above is we assume `$HOME/bin` exists and is on the $PATH. So i'm slightly inclined to think that we leave this as an exercise up to the developer instead of trying to automate via a script, as most developers of the project will already have this tooling configured locally. 

Alternatively if you think we should automate, then I suggest we create a `scripts/install-tools.sh` or similar script to do the heavily lifting I described above.

thommahoney: These are all really good points. I've gone ahead and removed `dependencies` from the `all` target and updated the developer documentation with a note about how to install them.

RELEASE.md
```diff
@@ -13,8 +13,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
     * Apply necessary labels (`enchancement`, `bug`, `documentation` etc) to all PRs intended for the release that you wish to appear in the `CHANGELOG.md`
     * **Only add labels for relevant changes**
     * `git checkout -b vx.x.x` where `vx.x.x` is your target version tag
-    * `make changelog`
-    * `git add CHANGELOG.md && git commit -m "vX.X.X"`
+    * `CHANGELOG_GITHUB_TOKEN=xxxx SEMVER_TAG=vx.x.x make changelo`
```
thommahoney, line 16:
```suggestion
    * `CHANGELOG_GITHUB_TOKEN=xxxx SEMVER_TAG=vx.x.x make changelog`
```

PR #49, opened 2020-04-29
`os.Rename` can fail if src and dst are not in the same filesystem

deweerdt:
```
$ ./fastly update
\ Initializing...Current version: 0.0.0-unknown
Latest version: 0.7.0
✓ Initializing...
✓ Fetching latest release...
✗ Replacing binary...

ERROR: error moving latest binary in place: rename /tmp/fastly_0.7.0_1588172969483405643/fastly /home/def/p/fastly/cli/fastly: invalid cross-device link.
```

This PR falls back to copying the files, should `os.Rename` fail.

### DIFF
--- a/pkg/update/root.go
+++ b/pkg/update/root.go
@@ -30,6 +30,28 @@ func NewRootCommand(parent common.Registerer, v Versioner, client api.HTTPClient
 	return &c
 }
 
+// copy a file called `src` to `dst`. This is useful if os.Rename fails because the two paths
+// are on different filesystems
+func copyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+
+	out, err := os.Create(dst)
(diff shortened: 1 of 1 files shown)


pkg/update/root.go
```diff
@@ -68,8 +90,10 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 	}
 
 	if err := os.Rename(latestPath, currentPath); err != nil {
-		progress.Fail()
-		return fmt.Errorf("error moving latest binary in place: %w", err)
+		if err := copyFile(latestPath, currentPath); err != nil {
```
phamann, line 93: We already have a more thorough implementation of copy file in the `common` package. https://github.com/fastly/cli/blob/master/pkg/common/file.go#L21 so it would be good to use that instead here.

phamann:
Have implemented the changes with #72, so closing this :)

PR #50, opened 2020-04-29
Ensure compute deploy selects the most ideal version to clone/activate 

phamann:
### TL;DR
As reported by @deweerdt, the current implementation will cause an API error to be returned if the latest service version is locked/deactivated. This PR updates `compute deploy` to use a more robust way of selecting and cloning the "ideal latest version". It uses the following logic:

- Find the active version and return
- If no active version, find the latest locked version and return
- Otherwise return the latest version

### Note:
If we think this logic is useful / common, we should upstream it to `go-fastly`.

### DIFF
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -232,7 +232,7 @@ func (c *Client) UpdatePackage(serviceID string, v int, path string) error {
 // - If no active version, find the latest locked version and return
 // - Otherwise return the latest version
 //
-// Note: assumes that the provided list is chronoligically sorted, such as that
+// Note: assumes that the provided list is chronologically sorted, such as that
 // provided from fastly.ListVersions().
 func GetLatestIdealVersion(versions []*fastly.Version) (*fastly.Version, error) {
 	var active, locked, latest *fastly.Version
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -84,14 +84,19 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 
 	if c.version == 0 {
 		progress.Step("Fetching latest version...")
-		version, err := c.Globals.Client.LatestVersion(&fastly.LatestVersionInput{
+		versions, err := c.Globals.Client.ListVersions(&fastly.ListVersionsInput{
 			Service: serviceID,
 		})
 		if err != nil {
-			return fmt.Errorf("error getting latest service version: %w", err)
+			return fmt.Errorf("error getting listing service versions: %w", err)
 		}
 
-		if version.Active {
+		version, err := GetLatestIdealVersion(versions)
+		if err != nil {
+			return fmt.Errorf("error finding latest service version")
+		}
+
+		if version.Active || version.Locked {
 			progress.Step("Cloning latest version...")
 			version, err = c.Globals.Client.CloneVersion(&fastly.CloneVersionInput{
 				Service: serviceID,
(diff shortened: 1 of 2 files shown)


pkg/compute/deploy.go
```diff
@@ -84,11 +84,26 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
-		version, err := c.Globals.Client.LatestVersion(&fastly.LatestVersionInput{
+		versions, err := c.Globals.Client.ListVersions(&fastly.ListVersionsInput{
 			Service: serviceID,
 		})
 		if err != nil {
-			return fmt.Errorf("error getting latest service version: %w", err)
+			return fmt.Errorf("error getting listing service versions: %w", err)
+		}
+
+		// Get the latest non-locked version.
```
thommahoney, line 94: Since versions are (deceivingly) non-linear, we might want to reconsider this approach. ie. If I have three versions:

- version 1: never activated
- version 2: cloned from 1, never activated
- version 3: cloned from 1, locked due to prior activation, currently inactive

... then choosing version 2 is not a good choice since it may be missing configuration from version 3. I think the right choice in this case is to clone version 3, though it may be difficult to ascertain. Leaving this note here for now as I investigate further.

denen99: Couple options 

a) Treat active version as the thing we should always clone from , regardless of any future clones to avoid the problem Thom mentioned.  
b) if no active version, take highest inactive version that we can write to and use that as the version to deploy to.

thommahoney: This PR implements (b) but I think we want to clone the most recently updated locked version instead. If there are no locked versions then we should use the highest version number. Certainly open other options!

denen99: I see I think that makes sense.  I am not sure how besides activating you get the config into a state of "locked" though but if locked is the thing we want to clone from, +1 from me.

phamann: Right, I feel like we can address both of your comments if revert back to the original logic and also clone the latest version if locked. I.e.

```suggestion
if c.version == 0 {
		progress.Step("Fetching latest version...")
		version, err := c.Globals.Client.LatestVersion(&fastly.LatestVersionInput{
			Service: serviceID,
		})
		if err != nil {
			return fmt.Errorf("error getting latest service version: %w", err)
		}

		if version.Active || version.Locked {
			progress.Step("Cloning latest version...")
			version, err = c.Globals.Client.CloneVersion(&fastly.CloneVersionInput{
				Service: serviceID,
				Version: version.Number,
			})
			if err != nil {
				return fmt.Errorf("error cloning latest service version: %w", err)
			}
		}

		c.version = version.Number
	}
```

thommahoney: I spoke with Patrick via DM about this implementation. He has a path forward that will:

1. Choose the active version and clone it
2. If no active version exists, choose the latest updated locked version and clone that
3. If no locked version exists, choose the latest updated version and use it

denen99: ✅

phamann: Have implemented the above in https://github.com/fastly/cli/commit/41af96c5c385101ba61942ad471189efacb9f2b7

pkg/compute/deploy.go
```diff
@@ -84,14 +84,19 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 
 	if c.version == 0 {
 		progress.Step("Fetching latest version...")
-		version, err := c.Globals.Client.LatestVersion(&fastly.LatestVersionInput{
+		versions, err := c.Globals.Client.ListVersions(&fastly.ListVersionsInput{
 			Service: serviceID,
 		})
 		if err != nil {
-			return fmt.Errorf("error getting latest service version: %w", err)
+			return fmt.Errorf("error getting listing service versions: %w", err)
```
tyler, line 91:
```suggestion
			return fmt.Errorf("error listing service versions: %w", err)
```

pkg/compute/deploy.go
```diff
@@ -221,3 +226,39 @@ func (c *Client) UpdatePackage(serviceID string, v int, path string) error {
 
 	return nil
 }
+
+// GetLatestIdealVersion gets the most ideal service version using the following logic:
+// - Find the active version and return
+// - If no active version, find the latest locked version and return
+// - Otherwise return the latest version
+//
+// Note: assumes that the provided list is chronoligically sorted, such as that
```
tyler, line 235:
```suggestion
// Note: assumes that the provided list is chronologically sorted, such as that
```

thommahoney, line 235: I think this implementation is Good Enough™️  for now but relying on chronological ordering (an analog of creation time by way of auto-incrementing version number) may have unintended consequences.

Using `updated_at` for ordering explicitly is probably the more correct behavior—the assumption being that users would want to use the latest version that changed, and not necessarily the latest version that exists. Hard to say.

phamann: 👍 Agreed. Have implemented sorting by `UpdatedAt` in https://github.com/fastly/cli/pull/50/commits/4cfd2cef3a2d97e199f97b663fa5ad04ed56a245

PR #55, opened 2020-05-06
logging: adds syslog logging endpoint

mccurdyc:

* [Fastly API Docs](https://docs.fastly.com/api/logging#api-section-logging_syslog)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/88b210b97743b0bfece81dc3dbbd7e3254e22d97/fastly/syslog.go)

## Validation

```
$ make test
```

```
( 
    cd /tmp/cli && \
    git checkout mccurdyc/logging-syslog && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/syslog/create.go
+++ b/pkg/logging/syslog/create.go
@@ -1,0 +1,74 @@
+package syslog
+
+import (
+	"io"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+// CreateCommand calls the Fastly API to create Amazon Syslog logging endpoints.
+type CreateCommand struct {
+	common.Base
+	manifest manifest.Data
+	Input    fastly.CreateSyslogInput
+
+	// We must store all of the boolean flags seperatly to the input structure
+	// so they can be casted to go-fastly's custom `Compatibool` type later.
+	useTLS bool
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCommand {
+	var c CreateCommand
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("create", "Create an Syslog logging endpoint on a Fastly service version").Alias("add")
+
+	c.CmdClause.Flag("name", "The name of the Syslog logging object. Used as a primary key for API access").Short('n').Required().StringVar(&c.Input.Name)
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
+	c.CmdClause.Flag("version", "Number of service version").Required().IntVar(&c.Input.Version)
+
+	c.CmdClause.Flag("address", "A hostname or IPv4 address").Required().StringVar(&c.Input.Address)
+
+	c.CmdClause.Flag("port", "The port number").UintVar(&c.Input.Port)
+	c.CmdClause.Flag("use-tls", "Whether to use TLS for secure logging. Can be either true or false").BoolVar(&c.useTLS)
+	c.CmdClause.Flag("tls-ca-cert", "A secure certificate to authenticate the server with. Must be in PEM format").StringVar(&c.Input.TLSCACert)
+	c.CmdClause.Flag("tls-hostname", "Used during the TLS handshake to validate the certificate").StringVar(&c.Input.TLSHostname)
+	c.CmdClause.Flag("tls-client-cert", "The client certificate used to make authenticated requests. Must be in PEM format").StringVar(&c.Input.TLSClientCert)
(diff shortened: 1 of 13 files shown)


pkg/logging/syslog/create.go
```diff
@@ -0,0 +1,74 @@
...
+
+// CreateCommand calls the Fastly API to create Amazon Syslog logging endpoints.
+type CreateCommand struct {
+	common.Base
+	manifest manifest.Data
+	Input    fastly.CreateSyslogInput
+
+	// We must store all of the boolean flags seperatly to the input structure
+	// so they can be casted to go-fastly's custom `Compatibool` type later.
+	useTLS bool
```
phamann, line 22: It should be possible to achieve this using a Kingpin custom parser instead of having to cast the value in the exeec function logic. Have a look at this and let me know your thoughts: https://github.com/alecthomas/kingpin#custom-parsers

pkg/logging/syslog/update.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+// UpdateCommand calls the Fastly API to update Amazon Syslog logging endpoints.
```
phamann, line 14:
```suggestion
// UpdateCommand calls the Fastly API to update Syslog logging endpoints.
```

pkg/logging/syslog/describe.go
```diff
@@ -0,0 +1,66 @@
...
+	"io"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+// DescribeCommand calls the Fastly API to describe an Amazon Syslog logging endpoint.
```
phamann, line 14:
```suggestion
// DescribeCommand calls the Fastly API to describe a Syslog logging endpoint.
```

pkg/logging/syslog/list.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+// ListCommand calls the Fastly API to list Amazon Syslog logging endpoints.
```
phamann, line 15:
```suggestion
// ListCommand calls the Fastly API to list Syslog logging endpoints.
```

pkg/logging/syslog/create.go
mccurdyc: Thanks for sharing! I did look into this and I personally prefer the unexported field. I don't have much experience with the alecthomas/kingpin package, but it looks like the custom parser would have to be for the value that you are parsing _into_, in this case, a `fastly.CompatiBool`. With that said, it might be useful to have in a common package in the CLI for continued use of parsing CLI flags into `fastly.Compatibool`s. However, I still think I prefer the unexported field for a few reasons:

1. Less code to manage.
2. Introducing additional exported types/functions to a `common` package (or having duplicate, unexported implementations).
3. If we were to format it as recommended, the flag parsing would look inconsistent. For example,

```golang
	c.CmdClause.Flag("port", "The port number").UintVar(&c.Input.Port)
	CompatiBool(c.CmdClause.Flag("use-tls", ...
	c.CmdClause.Flag("tls-ca-cert",...
```

phamann: Makes sense, I'm happy to leave as-is for now and we can refactor into the `common` package later if we feel it makes sense. :+1:

pkg/logging/syslog/create.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCommand {
+	var c CreateCommand
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("create", "Create an Syslog logging endpoint on a Fastly service version").Alias("add")
```
phamann, line 30:
```suggestion
	c.CmdClause = parent.Command("create", "Create a Syslog logging endpoint on a Fastly service version").Alias("add")
```

pkg/logging/syslog/root.go
```diff
@@ -0,0 +1,28 @@
...
+type RootCommand struct {
+	common.Base
+	// no flags
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("syslog", "Manipulate Fastly service version Syslog logging endpoints.")
```
phamann, line 21:
```suggestion
	c.CmdClause = parent.Command("syslog", "Manipulate Fastly service version Syslog logging endpoints")
```

PR #60, opened 2020-05-08
Improve `compute build` rust compilation error messaging

phamann:
### TL;DR
Improves the error messaging when Rust compilation fails during `fastly compute build`. 

### Why?
We've had lots of user feedback that the current implementation is confusing as it just returns an error code and no remediation message or help text. It was also not apparent that you could run the command with the `--verbose` flag to get the full output from the child `cargo` process which contains the real cause of the failure. 

### What
We now return the full buffered `stderr` output from `cargo build` if the process fails which hopefully gives the user much more information to solve their problem.  We also pass the `--verbose` flag to the child process if supplied giving even more information if required.

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -153,7 +153,7 @@ func (r Rust) Verify(out io.Writer) error {
 
 // Build implments the Toolchain interface and attempts to compile the package
 // Rust source to a Wasm binary.
-func (r Rust) Build(out io.Writer) error {
+func (r Rust) Build(out io.Writer, verbose bool) error {
 	// Get binary name from Cargo.toml.
 	var m CargoManifest
 	if err := m.Read("Cargo.toml"); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/compute/rust.go
```diff
@@ -164,12 +164,27 @@ func (r Rust) Build(out io.Writer) error {
...
+	args := []string{
+		toolchain,
+		"build",
+		"--bin",
+		binName,
+		"--release",
+		"--target",
+		WasmWasiTarget,
+		"--color",
+		"always",
```
thommahoney, line 176: Not necessary for this PR, but if people start using this tool in CI, they may not want colors.

pkg/compute/rust.go
```diff
@@ -201,15 +216,21 @@ func (r Rust) Build(out io.Writer) error {
...
+	}
+
+	// Wait for the command to exit.
+	if err := cmd.Wait(); err != nil {
+		// If we're not in verbose mode return the bufferred stderr output
+		// from cargo as the error.
+		if !verbose && stderrBuf.Len() > 0 {
+			return fmt.Errorf("error during compilation process:\n%s", strings.TrimSpace(stderrBuf.String()))
+		}
+		return fmt.Errorf("error during compilation process")
```
thommahoney, line 233: This line swallows the `err` from `cmd.Wait()` but I think it's usually just "exited with status 1` so 🤷

pkg/compute/rust.go
phamann: Yup I thought about this too. I think we need to do a sweep on the commands in another PR to ensure we don't assume TTY, ANSI etc support.

pkg/compute/rust.go
phamann: Yup this was intentional as the message wasn't helpful at all as you suggest. 👍

PR #61, opened 2020-05-08
Move the fossa security scan to master CI jobs

phamann:
### TL;DR
The `scan` job  in our current PR Action workflow is preventing external contributions from succeeding the workflow, thus giving false positives and making us force the merge with admin privileges. This is due to the way Actions (correctly) suppresses secrets on fork based builds to prevent them from leaking, which is causing the scan job to fail as it requires a secret. Therefore, we've decided to run this job on master pushes instead. Its a tradeoff as if a PR introduces a vulnerability we won't find out until its merged, however at least they will  be caught before a release.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,0 +1,32 @@
+on:
+  push:
+    branches:
+      - master
+name: Test
+jobs:
+  scan:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.14.x
+    - name: Install fossa
+      run: |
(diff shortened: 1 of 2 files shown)


mccurdyc:
Adding [this document](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) for those interested in reading more about "Actions (correctly) suppresses secrets on fork based builds".

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

phamann:
>It is interesting that the check for the scan job is still being reported in the PR that removes it, but this looks good to me.

@ezkl good catch. That was because I hadn't disabled the branch protection rule on `master` for `scan`.

PR #66, opened 2020-05-08
Basic support for historical & realtime stats

pteichman:
This gives the Fastly CLI a foothold in the stats world. Output is very basic, but roughly matches the old Python CLI.

This requires as-yet unmerged changes to go-fastly, so it's not expected to pass the build test.


pkg/stats/cmd_realtime.go
```diff
@@ -0,0 +1,133 @@
...
+			Timestamp uint64            `json:"timestamp"`
+			Data      []json.RawMessage `json:"data"`
+		}
+
+		err := client.GetRealtimeStatsJSON(&fastly.GetRealtimeStatsInput{
+			Service:   service,
+			Timestamp: timestamp,
+		}, &envelope)
+		if err != nil {
+			fmt.Fprintf(out, "ERROR: fetching stats: %s", err)
```
phamann, line 87: We have a [helper](https://github.com/fastly/cli/blob/master/pkg/text/text.go#L136-L140) in the text package to format errors:

```suggestion
			text.Error(out, "fetching stats: %w", err)
```

pkg/stats/cmd_realtime.go
```diff
@@ -0,0 +1,133 @@
...
+	var timestamp uint64
+	for {
+		var envelope realtimeResponse
+
+		err := client.GetRealtimeStatsJSON(&fastly.GetRealtimeStatsInput{
+			Service:   service,
+			Timestamp: timestamp,
+		}, &envelope)
+		if err != nil {
+			fmt.Fprintf(out, "ERROR: fetching stats: %s", err)
```
phamann, line 111:
```suggestion
			text.Error(out, "ERROR: fetching stats: %w", err)
```

pkg/stats/cmd_regions.go
```diff
@@ -0,0 +1,39 @@
+package stats
```
phamann, line 1: As above, can we rename this to pkg/stats/regions.go?

pkg/stats/cmd_regions.go
```diff
@@ -0,0 +1,39 @@
...
+	c.Globals = globals
+	c.CmdClause = parent.Command("regions", "List stats regions")
+	return &c
+}
+
+// Exec implements the command interface.
+func (c *RegionsCommand) Exec(in io.Reader, out io.Writer) error {
+	resp, err := c.Globals.Client.GetRegions()
+	if err != nil {
+		return fmt.Errorf("regions: %w", err)
```
phamann, line 31:
```suggestion
		return fmt.Errorf("fetching regions: %w", err)
```

pkg/stats/cmd_regions.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	for _, region := range resp.Data {
+		fmt.Fprintf(out, "%s\n", region)
```
phamann, line 35:
```suggestion
		text.Output(out, "%s", region)
```

pkg/stats/root.go
```diff
@@ -0,0 +1,26 @@
...
+// RootCommand dispatches all "stats" commands.
+type RootCommand struct {
+	common.Base
+}
+
+// NewRootCommand returns a new top level "stats" command.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("stats", "View service stats (historical and realtime)")
```
phamann, line 19:
```suggestion
	c.CmdClause = parent.Command("stats", "View statistics (historical and realtime) for a Fastly service")
```

pkg/stats/template.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := mapstructure.Decode(block, &agg); err != nil {
+		return err
+	}
+
+	hitRate := 0.0
+	if agg.Hits > 0 {
+		hitRate = float64((agg.Hits - agg.Miss - agg.Errors)) / float64(agg.Hits)
+	}
+
+	// TODO: parse the JSON more strictly so this doesn't need to be dynamic.
```
phamann, line 50: Noting the TODO here, do you want to do this as part of this PR or at a later date?

pkg/stats/cmd_realtime.go
```diff
@@ -0,0 +1,133 @@
...
+			agg := block.Aggregated
+
+			// FIXME: These are heavy-handed compatibility
+			// fixes for stats vs realtime, so we can use
+			// fmtBlock for both.
+			agg["start_time"] = block.Recorded
+			delete(agg, "miss_histogram")
+
+			if err := fmtBlock(out, service, agg); err != nil {
+				fmt.Fprintf(out, "ERROR: formatting stats: %s", err)
```
phamann, line 126:
```suggestion
				text.Error(out, "formatting stats: %w", err)
```

pkg/stats/cmd_realtime.go
```diff
@@ -0,0 +1,133 @@
...
+		}, &envelope)
+		if err != nil {
+			fmt.Fprintf(out, "ERROR: fetching stats: %s", err)
+			continue
+		}
+		timestamp = envelope.Timestamp
+
+		for _, block := range envelope.Data {
+			out.Write(block)
+			out.Write([]byte{'\n'})
```
phamann, line 94:
```suggestion
			text.Break(out)
```

pkg/app/run.go
```diff
@@ -288,6 +299,11 @@ func Run(args []string, env config.Environment, file config.File, configFilePath
 		return fmt.Errorf("error constructing Fastly API client: %w", err)
 	}
 
+	globals.RTSClient, err = fastly.NewRealtimeStatsClientForEndpoint(token, fastly.DefaultRealtimeStatsEndpoint)
+	if err != nil {
+		return fmt.Errorf("error constructing Fastly realtime stats client: %w", err)
+	}
```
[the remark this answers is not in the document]
pteichman: This sounds fine to me but it's a little annoying in practice: it means getting `token` before any of the CLI is set up, and there's some token validation that would be a little weird before setting up the static bits of the CLI. Happy to go either way here, but that's enough churn that I wanted to check.

pkg/stats/cmd_historical.go
```diff
@@ -0,0 +1,140 @@
...
+
+func writeBlocksJSON(out io.Writer, service string, blocks []statsResponseData) error {
+	for _, block := range blocks {
+		if err := json.NewEncoder(out).Encode(block); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
pteichman: Maybe? So far the only things that need a block are in this package, and I don't think any of the format stuff is finalized enough to want to make it a shared routine.

I think we're probably going to have to split the objects for historical & realtime stats, so this may not be as reusable as it looks either.

pkg/stats/cmd_historical.go
```diff
@@ -0,0 +1,140 @@
...
+
+// NewHistoricalCommand is the "stats historical" subcommand.
+func NewHistoricalCommand(parent common.Registerer, globals *config.Data) *HistoricalCommand {
+	var c HistoricalCommand
+	c.Globals = globals
+
+	c.CmdClause = parent.Command("historical", "Query historical stats")
```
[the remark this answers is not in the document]
pteichman: I like View a little better.

pkg/stats/template.go
pteichman: Let's do this later. I need to think about whether we can unify these APIs at all.

pkg/stats/cmd_historical.go
```diff
@@ -0,0 +1,140 @@
...
+
+	var envelope statsResponse
+	err := c.Globals.Client.GetStatsJSON(&req, &envelope)
+	if err != nil {
+		return err
+	}
+
+	if envelope.Status != statusSuccess {
+		return fmt.Errorf("non-success response: %s", envelope.Msg)
+	}
```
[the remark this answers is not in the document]
pteichman: GetStatsJSON will return an error on non-2xx statuses, so by this point we know it's an HTTP success. This check is based on a conservative reading of the stats response docs: https://docs.fastly.com/api/stats#Format

Basically this is the check to see whether `msg` has anything useful for us.

pteichman:
Force pushed because I rebased the original commits; there are no changes to those aside from conflict fixes. Everything new is in the four new commits.

pteichman:
Meeting you halfway: this adds tests for the `regions` and `historical` commands, but leaves out `realtime` for now since it's an infinite loop and I want to put in the time to test it properly.

Force pushed to rebase on master: all of the new code is in the last commit.
And as before, this continues to fail its build until we land the go-fastly upgrade.

From here my plan is to minimize churn:
1) Get go-fastly up to date.
2) Rebase this thing on master & verify tests.
3) Merge.

mccurdyc:
@pteichman go-fastly should be up to date now :)

https://github.com/fastly/cli/pull/71

pteichman:
(another rebase; resolving conflicts with the new trunk)

pkg/stats/historical.go
```diff
@@ -0,0 +1,110 @@
...
+
+	c.CmdClause.Flag("format", "Output format (json)").EnumVar(&c.formatFlag, "json")
+
+	return &c
+}
+
+// Exec implements the command interface.
+func (c *HistoricalCommand) Exec(in io.Reader, out io.Writer) error {
+	service, source := c.manifest.ServiceID()
+	if source != manifest.SourceUndefined {
```
phamann, line 47: This would still allow for an empty service ID to be sent in the request. Therefore we should error if none passed and always set the input field.

```suggestion
if source == manifest.SourceUndefined {
    return errors.ErrNoServiceID
}
c.Input.Service = service
```

pteichman: That's intentional, an empty service is a valid request: it will dump all of the services that api token  can read.

phamann: :+1: Makes sense, however I now feel uncomfortable that the "dump all services" behaviour is essentially an undocumented side-effect and not a first-class discoverable feature - which it should be. 

Also the possibility of `source` being undefined is reduced for C@E services as the service-id is always persisted and read from the `fastly.toml`, which is why we have this pattern as it allows you to omit the `--service-id` flag within your project directory.

So a couple of ideas for the future, but don't need to block this PR being merged:
- We add an additional `--all-services` (or similar) flag, which makes it a first-class feature and allows people to intentionally invoke the behaviour. (my preferred option)
- We update the flag description for `--service-id` to document that if left blank will return all services.

PR #67, opened 2020-05-11
Verify fastly crate version during compute build.

phamann:
### TL;DR
Verifies that the `fastly` crate exists as a dependency and is up-to-date during `fastly compute build`. This allows us to ensure that the package program is compatible with the edge runtime environment at compilation time.


pkg/compute/rust.go
```diff
@@ -44,8 +51,23 @@ func (m *CargoManifest) Read(filename string) error {
 	return err
 }
 
+// CargoLock models the package confuiguration properties of a Rust Cargo
```
acfoltzer, line 54:
```suggestion
// CargoLock models the package configuration properties of a Rust Cargo
```

pkg/compute/rust.go
```diff
@@ -148,6 +170,51 @@ func (r Rust) Verify(out io.Writer) error {
...
+	}
+	latest, err := semver.Parse(l)
+	if err != nil {
+		return err
+	}
+
+	if version.LT(latest) {
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("fastly crate not up-to-date"),
+			Remediation: fmt.Sprintf("To fix this error, run the following command:\n\n\t$ %s\n", text.Bold("cargo update -p fastly")),
```
acfoltzer, line 214: I goofed here. Unfortunately this is not going to be a sufficient remediation when we introduce a semver-breaking change. For example, suppose we have

```toml
[dependencies]
fastly = "0.2.0-alpha4"
```

When I try to use cargo to update it, even specifying the precise version I want, I get:

```
% cargo update -p fastly --precise 0.3.2
    Updating crates.io index
error: failed to select a version for the requirement `fastly = "^0.2.0-alpha4"`
  candidate versions found which didn't match: 0.3.2, 0.3.1, 0.3.0
  location searched: crates.io index
required by package `fastly-template-rust-default v0.1.0 (/home/acfoltzer/src/fastly-template-rust-default)`
```

When I initially proposed using `cargo update`, I forgot that it only updates to versions that are compatible with what's in the `Cargo.toml`. I think the only robust remediation would be to ask them to update that file:

> To fix this error, edit `Cargo.toml` to set `fastly = "^0.3.2"`.

phamann: Ah yes, that makes sense. I've now updated the message per your suggestion.

Screenshot 2020-05-11 at 19 57 14

phamann:
@acfoltzer and I discussed this PR on slack, we realised that prompting a user to add `fastly = "^0.3.2" to their `Cargo.toml` wasn't a sufficient fix as the lock file also need updating before the verification runs again. Therefore, we've now added a dual step remediation message in these cases. However if their current version is within the semver minor range of the latest, then only a `cargo update -p fastly` is needed, the logic now caters for these edge cases and serves the appropriate message. I've also added a more robust test suit to ensure this new logic is covered. 

@peterbourgon When/If you have some spare time, I'd appreciate a quick review on this. Danke!

pkg/compute/rust.go
```diff
@@ -148,6 +170,89 @@ func (r Rust) Verify(out io.Writer) error {
...
+
+	var lock CargoLock
+	if err := lock.Read("Cargo.lock"); err != nil {
+		return fmt.Errorf("error reading Cargo.lock file: %w", err)
+	}
+
+	// Fetch the latest crate version from cargo.io API.
+	l, err := GetLatestCrateVersion(r.client, "fastly")
+	if err != nil {
+		return err
```
peterbourgon, line 193:
```suggestion
		return fmt.Errorf("error fetching latest crate version: %w", err)
```

pkg/compute/rust.go
```diff
@@ -148,6 +170,89 @@ func (r Rust) Verify(out io.Writer) error {
...
+	}
+	latest, err := semver.Parse(l)
+	if err != nil {
+		return err
```
peterbourgon, line 197:
```suggestion
		return fmt.Errorf("error parsing latest crate SemVer: %w", err)
```

pkg/compute/rust.go
```diff
@@ -148,6 +170,89 @@ func (r Rust) Verify(out io.Writer) error {
...
+			),
+		}
+	}
+
+	// Parse lockfile version to semver.
+	version, err := semver.Parse(crate.Version)
+	if err != nil {
+		return fmt.Errorf("error parsing Cargo.lock file: %w", err)
+	}
+	// Parse the lowest minor semver for the latest release.
```
peterbourgon, line 227:
```suggestion

	// Parse the lowest minor semver for the latest release.
```

pkg/compute/rust.go
```diff
@@ -148,6 +170,89 @@ func (r Rust) Verify(out io.Writer) error {
...
+			Inner: fmt.Errorf("fastly crate not up-to-date"),
+			Remediation: fmt.Sprintf(
+				"To fix this error, edit %s with:\n\n\t %s\n\nAnd then run the following command:\n\n\t$ %s\n",
+				text.Bold("Cargo.toml"),
+				text.Bold(fmt.Sprintf(`fastly = "^%s"`, latest)),
+				text.Bold("cargo update -p fastly"),
+			),
+		}
+	}
+
```
peterbourgon, line 255: Not a request for change, but I'm surprised this level of semantic version "parsing" is required by Verify...

pkg/compute/rust.go
```diff
@@ -264,3 +369,62 @@ func (r Rust) Build(out io.Writer, verbose bool) error {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("error fetching latest crate version %s", resp.Status)
+	}
+
+	defer resp.Body.Close()
```
peterbourgon, line 401:
```suggestion

	resp, err := client.Do(req)
	if err != nil {
		return "", err
	}
	defer resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		return "", fmt.Errorf("error fetching latest crate version: %s", resp.Status)
	}
```

pkg/compute/rust.go
phamann: Haha me too 😆

PR #73, opened 2020-05-14
logging: adds Scalyr logging endpoint support

mccurdyc:

* [Fastly API Docs](https://docs.fastly.com/api/logging#api-section-logging_scalyr)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/c2c2c62210800daab641161412fbe210004aac0a/fastly/scalyr.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-scalyr && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/scalyr/create.go
+++ b/pkg/logging/scalyr/create.go
@@ -1,0 +1,100 @@
+package scalyr
+
+import (
+	"io"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+// CreateCommand calls the Fastly API to create Scalyr logging endpoints.
+type CreateCommand struct {
+	common.Base
+	manifest manifest.Data
+
+	// required
+	EndpointName string // Can't shaddow common.Base method Name().
+	Token        string
+	Version      int
+
+	// optional
+	Format            common.OptionalString
+	FormatVersion     common.OptionalUint
+	ResponseCondition common.OptionalString
+	Placement         common.OptionalString
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCommand {
+	var c CreateCommand
+
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("create", "Create a Scalyr logging endpoint on a Fastly service version").Alias("add")
+
+	c.CmdClause.Flag("name", "The name of the Scalyr logging object. Used as a primary key for API access").Short('n').Required().StringVar(&c.EndpointName)
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
+	c.CmdClause.Flag("version", "Number of service version").Required().IntVar(&c.Version)
+
+	c.CmdClause.Flag("auth-token", "The token to use for authentication (https://www.scalyr.com/keys)").Required().StringVar(&c.Token)
+
+	c.CmdClause.Flag("format", "Apache style log formatting").Action(c.Format.Set).StringVar(&c.Format.Value)
+	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint. Can be either 2 (default) or 1").Action(c.FormatVersion.Set).UintVar(&c.FormatVersion.Value)
+	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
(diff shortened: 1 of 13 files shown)


pkg/logging/scalyr/create.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCommand {
+	var c CreateCommand
+
+	// Without this, parsing flags leads to panics.
+	// This is because if we try to scan directly into c.Input.<Field> without first
+	// creating a valid reference, flags will attempt to be scanned into null pointers.
+	// This will result in sending the zero value values for unset flags which is acceptable
+	// for Creates, but not Updates.
```
mccurdyc, line 29: Thoughts on the comment? Good? Necessary?

pkg/logging/scalyr/create.go
```diff
@@ -0,0 +1,71 @@
...
+		Token:             fastly.String(""),
+		ResponseCondition: fastly.String(""),
+		Placement:         fastly.String(""),
+	}
+
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("create", "Create a Scalyr logging endpoint on a Fastly service version").Alias("add")
+
+	c.CmdClause.Flag("name", "The name of the Scalyr logging object. Used as a primary key for API access").Short('n').Required().StringVar(c.Input.Name)
```
mccurdyc, line 43:
```suggestion
	c.CmdClause.Flag("name", "The name of the Scalyr logging object. Used as a primary key for API access").Short('n').Required().StringVar(&c.Input.Name)
```
(suggestion just to highlight inconsistency, it is not a valid change)

Note the inconsistency with other "super" fields.

akappen:
We just released region support for Scalyr logging. The API documentation has been updated if you'd like to include support for that in the CLI. Thanks!

pteichman:
Conflict fix is over at https://github.com/fastly/cli/commit/e556630c38c6067e41bd33b64090d3400857deae

mccurdyc:
@akappen Awesome! We will have to include in [`fastly/go-fastly`](https://github.com/fastly/go-fastly/blob/master/fastly/scalyr.go#L11) and then cut a new release and then update the dependency version used in the CLI and then update this PR.

mccurdyc:
@phamann I'm going to come back to this to address [this comment](https://github.com/fastly/cli/pull/73#issuecomment-630443690), but I would like your thoughts on the `CreateCommand.createInput() Create*Input` pattern as I will adopt this in the other open PRs and come back to Scalyr once I cut a new go-fastly release.

pkg/logging/scalyr/create.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, errors.ErrNoServiceID
+	}
+
+	input.Service = serviceID
+
+	if c.Version <= 0 {
+		return nil, fmt.Errorf("error invalid value for required field (version): %d", c.Version)
+	}
+
+	input.Version = c.Version
```
phamann, line 70: Whilst this is good, we don't validate negative numbers for version input anywhere else in the project. I'm also conscious of doing this kind of client-side validation when the API will do it anyway. 

The big revelation for me here is why are the version fields in go-fastly `uint`'s to avoid this completely 😱 (not your fault, just convention there I hadn't realised until now)

pkg/logging/scalyr/create.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if c.EndpointName == "" {
+		return nil, fmt.Errorf("error invalid value for required field (name): '%s'", c.EndpointName)
+	}
+
+	input.Name = &c.EndpointName
+
+	if c.Token == "" {
+		return nil, fmt.Errorf("error invalid value for required field (auth-token): '%s'", c.Token)
+	}
```
phamann, line 80: As per above, I'd prefer for this validation to happen at the API layer.

pkg/logging/scalyr/create.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &input, nil
+}
+
+// Exec invokes the application logic for the command.
+func (c *CreateCommand) Exec(in io.Reader, out io.Writer) error {
+	input, err := c.createInput()
+	if err != nil {
+		return err
+	}
```
phamann, line 108: I like this pattern 👍

pkg/logging/scalyr/api_test.go
```diff
@@ -0,0 +1,433 @@
...
+func TestScalyrCreate(t *testing.T) {
+	for _, testcase := range []struct {
+		args       []string
+		api        mock.API
+		want       *fastly.Scalyr
+		wantError  string
+		wantOutput string
+	}{
+		{
+			args:      []string{"logging", "scalyr", "create", "--service-id", "123", "--version", "1", "--name", "", "--auth-token", "abc"},
```
mccurdyc, line 28: @phamann I added this test case as explicitly-set zero values in cli args results in runtime panics, hence the validation at the CLI layer.

Maybe runtime panics for this behavior are acceptable over validation. I do think this is an edge case as it appropriately and nicely handles omitted required flags.

mccurdyc:
Force-pushed rebasing master.

mccurdyc:
Force-pushed squashing into a single commit.

mccurdyc:
Force-pushed rebasing master.

PR #82, opened 2020-05-19
mock: remove dependency on app package

mccurdyc:
Removes dependency on `app` package which improved convenience, but limited its use.

See https://github.com/fastly/cli/pull/73/files#diff-aa80104cc7a38c091d22ef0cab9d4853R55

Previously, this introduced a cyclical dependency and made it impossible to test the package without significant changes.


mccurdyc:
Force-pushed rebasing master.

PR #85, opened 2020-05-21
Describe subcommand consistent --name short flag -d -> -n

mccurdyc:
## Proposed Change(s)

* Consistently use of `-n` as the short flag for `--name` (previously was `-d` that got copy-pasted ubiquitously).

```shell
$ find . -type f -name "describe.go" -exec sed -i "s/\"name\", \(.*\).Short('d')/\"name\", \1.Short('n')/g" {} \;`
```

## Validation

* `make test`


mccurdyc:
Force-pushed squashing into single commit.

PR #87, opened 2020-05-22
Add ability to exclude files from build package

### LINKED ISSUE #24: unintended files included in upload package
Fastly CLI version v0.5.0 (0fba867)

`fastly compute build` generates a package that contains unintended files like `.main.rs.swp` when vim is open. Maybe it should respect `.gitignore`? No I will not switch to emacs.

```
$ tar -tzvf pkg/ecp-loadtest-graphql.tar.gz
-rw-r--r--  0 mteodoro staff     161 Apr  8 07:57 ecp-loadtest-graphql/fastly.toml
-rw-r--r--  0 mteodoro staff     321 Apr  8 07:48 ecp-loadtest-graphql/Cargo.toml
drwxr-x---  0 mteodoro staff       0 Apr  3 17:04 ecp-loadtest-graphql/bin/
-rw-r--r--  0 mteodoro staff 34663899 Apr 17 12:24 ecp-loadtest-graphql/bin/main.wasm
drwxr-x---  0 mteodoro staff        0 Apr 17 12:24 ecp-loadtest-graphql/src/
-rw-r--r--  0 mteodoro staff    16384 Apr 17 12:24 ecp-loadtest-graphql/src/.main.rs.swp
-rw-r--r--  0 mteodoro staff     8163 Apr  8 07:58 ecp-loadtest-graphql/src/main.rs
```

phamann:
### TL;DR
This adds two new features:

- Change the default behaviour of `compute build` to not include the `src` directory in the package archive and add a `--include-src` flag if desired.
- Add the ability to exclude files from the package archive via glob patterns defined in a `.fastlyignore` file in the root of the project.

 Fixes #24. 

### Notes for reviewers:
I have renamed the existing `compute` integration style tests to a new file `compute_integration_test.go` which tests the external command API and have created a new `compute_test.go` to test the unexported (but still significant) functions within the package. Although I did a `git mv` the diff isn't pleasant, so it's easier to review the new commit directly. https://github.com/fastly/cli/commit/4ce16bb129bd31e544553a63cd0fa5b49a4dc333


pkg/compute/build.go
```diff
@@ -129,17 +124,168 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+
+// fileNameWithoutExtension returns a filename with its extension stripped.
+func fileNameWithoutExtension(filename string) string {
+	base := filepath.Base(filename)
+	firstDot := strings.Index(base, ".")
+	if firstDot > -1 {
+		return base[:firstDot]
+	}
+	return base
+}
```
peterbourgon, line 222: Do you want `/path/to/whatever.tar.gz` to return `whatever` or `whatever.tar`?

pkg/compute/build.go
```diff
@@ -129,17 +124,168 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+				files = append(files, path)
+			}
+		}
+		return nil
+	}); err != nil {
+		return files, err
+	}
+
+	return files, nil
+}
```
peterbourgon, line 291: Total nit, take it or leave it, but consider:

```go
func getNonIgnoredFiles(base string, ignoredFiles []string) ([]string, error) {
	ignore := make(map[string]bool, len(ignoredFiles))
	for _, s := range ignoredFiles {
		ignore[s] = true
	}

	err := filepath.Walk(base, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if info.IsDir() {
			return nil
		}
		if ignore[path] {
			return nil
		}
		files = append(files, path)
		return nil
	})
	
	return files, err
}
```

pkg/compute/build.go
phamann: `whatever`

pkg/app/run_test.go
```diff
@@ -275,6 +275,7 @@ COMMANDS
 
     --name=NAME          Package name
     --language=LANGUAGE  Language type
+    --include-src        Include src directory in package
```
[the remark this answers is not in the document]
phamann: Our original intention was the source would be useful for operational debugging, so I envision someone would probably only use this flag if Fastly Support/Engineering have explicitly asked them too.

pkg/compute/build.go
phamann: Yeah, I considered a map though thought its construction would be slower that the sort, but it does give you `O(1)` lookups and does look cleaner. Will change :+1:

pkg/compute/build.go
peterbourgon: OK, then this implementation is good 👍

pkg/compute/build.go
phamann: Changed in https://github.com/fastly/cli/pull/87/commits/2e614bb6d4cf5628877e17b0a9a181ff24ee3f37

pkg/compute/build.go
```diff
@@ -52,6 +45,7 @@ func NewBuildCommand(parent common.Registerer, client api.HTTPClient, globals *c
 	c.CmdClause = parent.Command("build", "Build a Compute@Edge package locally")
 	c.CmdClause.Flag("name", "Package name").StringVar(&c.name)
 	c.CmdClause.Flag("language", "Language type").StringVar(&c.lang)
+	c.CmdClause.Flag("include-src", "Include src directory in package").BoolVar(&c.includeSrc)
```
thommahoney, line 48: I think the name of this flag should be changed so that it's less specific to Rust. eg. If we changed the name to `--include-source`, we can generically apply it to other languages in the future. The description would need to be changed as well eg. `Include source code in built package`

pkg/compute/compute_intergration_test.go
```diff
@@ -0,0 +1,1341 @@
+package compute_test
```
thommahoney, line 1: this file's name is misspelled: `s/intergration/integration/`

pkg/compute/build.go
```diff
@@ -129,17 +123,166 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	}
+
+	binFiles, err := getNonIgnoredFiles("bin", ignoreFiles)
 	if err != nil {
 		return err
 	}
+	files = append(files, binFiles...)
+
+	if c.includeSrc {
+		srcFiles, err := getNonIgnoredFiles("src", ignoreFiles)
```
thommahoney, line 144: If we decide to change the flag to be less rust-specific, we'll want to map what it means per supported language in the future, but this implementation can stay for now.

phamann: Have added TODO comment :+1:

PR #89, opened 2020-05-26
Adds Azure Blob Storage logging endpoint support

mccurdyc:

* [Fastly API Docs](https://developer.fastly.com/reference/api/logging/azureblob/)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/c2c2c62210800daab641161412fbe210004aac0a/fastly/blobstorage.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-azure && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -19,6 +19,7 @@ import (
 	"github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/cli/pkg/healthcheck"
 	"github.com/fastly/cli/pkg/logging"
+	"github.com/fastly/cli/pkg/logging/azureblob"
 	"github.com/fastly/cli/pkg/logging/bigquery"
 	"github.com/fastly/cli/pkg/logging/cloudfiles"
 	"github.com/fastly/cli/pkg/logging/digitalocean"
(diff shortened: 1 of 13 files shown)


mccurdyc:
@phamann I assumed that we wanted to follow the naming convention (i.e., "blobstorage") in go-fastly.

I would personally prefer something that identifies this as an "Azure Blob Storage" rather than just "blob storage" as I think "blob storage" has weighted meaning. But having consistency is also very important.

pkg/logging/blobstorage/root.go
```diff
@@ -0,0 +1,28 @@
...
+type RootCommand struct {
+	common.Base
+	// no flags
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("blobstorage", "Manipulate Fastly service version Azure Blob Storage logging endpoints")
```
phamann, line 21: It would be good if we can keep consistency with the API endpoints, as `blobstorage` is too generic:
```suggestion
	c.CmdClause = parent.Command("azureblob", "Manipulate Fastly service version Azure Blob Storage logging endpoints")
```

mccurdyc: > It would be good if we can keep consistency with the API endpoints, as `blobstorage` is too generic:

@phamann per you review comment, do you think everything (i.e., package, filenames, etc.) should be renamed? (I would prefer everything).

The only reason that I didn't was because it is just "BlobStorage" in [go-fastly](https://github.com/fastly/go-fastly/blob/master/fastly/blobstorage.go).

mccurdyc:
Force pushed addressing review comments (i.e., blobstorage -> azureblob) and squashing into a single commit.

1. `fd --exclude .git . -exec rename s/blobstorage/azureblob/ {} \;`
2. `rename blobstorage azureblob pkg/logging/blobstorage*`
3. `rename blobstorage azureblob pkg/logging/azureblob/*`

mccurdyc:
Force-pushed gofmt check fix.

mccurdyc:
Force-pushed rebasing master.

pkg/app/run.go
```diff
@@ -256,18 +257,25 @@ func Run(args []string, env config.Environment, file config.File, configFilePath
 	digitaloceanUpdate := digitalocean.NewUpdateCommand(digitaloceanRoot.CmdClause, &globals)
 	digitaloceanDelete := digitalocean.NewDeleteCommand(digitaloceanRoot.CmdClause, &globals)
 
-	statsRoot := stats.NewRootCommand(app, &globals)
```
mccurdyc, line 259: I didn't realize this was formatted this way when it got merged to master.

PR #93, opened 2020-05-26
Real Time Charts

simonwistow:
This builds on the work done on the real time stats display by generating an ASCII realtime dashboard.

It's a port of a standalone experiment I did a while back tied into the CLI framework. Feel free to resize the terminal and hit `q` or `ctrl-c` to quit. 

I figure at worst case it might spark some inspiration in someone and/or someone can critique my Go and I can improve. Best case scenario it's actually useful, even if it's just as a demo.

Potential todos:
- [ ] get some input on the design
- [ ] fix the number of items in the Sparklines so they fill the whole of the box (or find a way to center them)
- [ ] play around with using the Plot type instead of Sparkline for some of the graphs
- [ ] add new "tabs" with WAF and Video and Request Type (IPv6, HTTP 1/2/3 etc etc) dashboards
- [ ] Get more details such as Customer name/id and Service name and put it in the header
- [ ] Fetch a list of Datacenters and use that so to stabilize the per-dc display
- [ ] Figure out some way to do the Miss Histogram
- [ ] Allow switching between Aggregate Data and Per DC (or maybe even Per Region)
- [ ] Maybe experiment with [TermDash](https://github.com/mum4k/termdash) as an alternative charting library
- [ ] Allow configuring of the layout via a config file (à la [GrafTerm](https://github.com/slok/grafterm/blob/master/docs/cfg.md)). Later, if that config was stored in the Fastly API, the dashboard in the UI could also use it.

stats

### DIFF
--- a/pkg/stats/realtime.go
+++ b/pkg/stats/realtime.go
@@ -64,6 +64,7 @@ func loopJSON(client api.RealtimeStatsInterface, service string, out io.Writer) 
 			Data      []json.RawMessage `json:"data"`
 		}
 
+		// this is annoyingly similar to the same loop in loopText
 		err := client.GetRealtimeStatsJSON(&fastly.GetRealtimeStatsInput{
 			Service:   service,
 			Timestamp: timestamp,
@@ -82,6 +83,35 @@ func loopJSON(client api.RealtimeStatsInterface, service string, out io.Writer) 
 }
 
 func loopText(client api.RealtimeStatsInterface, service string, out io.Writer) error {
+	responses := make(chan realtimeResponse)
+	errors := make(chan error)
+	go fetchData(client, service, responses, errors)
+
+	for {
+		select {
+		case err := <-errors:
+			text.Error(out, "fetching stats: %w", err)
+			continue
+		case envelope := <-responses:
+			for _, block := range envelope.Data {
+				agg := block.Aggregated
+
+				// FIXME: These are heavy-handed compatibility
+				// fixes for stats vs realtime, so we can use
+				// fmtBlock for both.
+				agg["start_time"] = block.Recorded
+				delete(agg, "miss_histogram")
+
+				if err := fmtBlock(out, service, agg); err != nil {
+					text.Error(out, "formatting stats: %w", err)
+					continue
+				}
+			}
+		}
+	}
+}
+
+func fetchData(client api.RealtimeStatsInterface, service string, responses chan realtimeResponse, errors chan error) {
 	var timestamp uint64
 	for {
 		var envelope realtimeResponse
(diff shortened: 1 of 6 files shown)


pkg/stats/realtime.go
```diff
@@ -82,6 +87,74 @@ func loopJSON(client api.RealtimeStatsInterface, service string, out io.Writer)
...
+	}
+	defer view.Close()
+	view.Resize()
+	view.SetLayout()
+
+	go fetchData(client, service, responses, errors)
+
+	for {
+		select {
+		case err := <-errors:
```
pteichman, line 134: Ok, here it makes more sense to use a channel because you do want to `select` across your remote calls and `view` events (because the termui API is all channel based).

So I might do a combination of what was above:
* make one channel of `struct { realtimeResponse, error }` (to make future loop changes easier)
* have the various loopFoos create a `context.Context` using WithCancel so you get a CancelFunc
* make fetchData cancel-able by adding a `ctx context.Context` as its first argument (and check for cancelation in its loop)
* call the CancelFunc at all of the loopFoo exit points (defer is fine here too) so that fetchData won't survive any longer than loopFoo

pkg/stats/realtime.go
```diff
@@ -82,6 +87,74 @@ func loopJSON(client api.RealtimeStatsInterface, service string, out io.Writer)
...
+				if err := view.UpdateStats(block); err != nil {
+					text.Error(out, "charting stats: %w", err)
+					continue
+				}
+				view.Render()
+			}
+		case e := <-view.Events:
+			switch e.ID { // event string/identifier
+			case "q", "<C-c>": // press 'q' or 'C-c' to quit
+				return nil
```
pteichman, line 148: This return is one place the infinite looping fetchData will leak.

pkg/stats/view.go
```diff
@@ -0,0 +1,296 @@
...
+	IoChart      *widgets.Sparkline
+	IoData       []float64
+	LoggingLabel *widgets.Paragraph
+	LoggingGroup *widgets.SparklineGroup
+	LoggingChart *widgets.Sparkline
+	LoggingData  []float64
+}
+
+func NewView(service string) (*View, error) {
+	var view = View{}
```
pteichman, line 72: It's pretty rare to see an empty struct declared like this: I'd expect to see one of:
```
var view View
```
or combine this line and the next into
```
view := View{
    Service: service,
}
```

pkg/stats/view.go
```diff
@@ -0,0 +1,296 @@
...
+	view.Service = service
+
+	if err := ui.Init(); err != nil {
```
pteichman, line 75: Calling `ui.Init` here seems orthogonal to creating a new View, and if this doesn't Init() then NewView doesn't need to return an error at all: that's useful because it signals to other code that this can never fail.

Depending on what happens in `ui.Init` it may also unnecessarily prevent you from calling NewView twice.

pkg/stats/view.go
```diff
@@ -0,0 +1,296 @@
...
+			ui.NewCol(1.0/2, ui.NewRow(0.2, v.HitLabel), ui.NewRow(0.8, v.HitGroup)),
+		),
+		ui.NewRow(0.22,
+			ui.NewCol(1.0/2, ui.NewRow(0.2, v.IoLabel), ui.NewRow(0.8, v.IoGroup)),
+			ui.NewCol(1.0/2, ui.NewRow(0.2, v.LoggingLabel), ui.NewRow(0.8, v.LoggingGroup)),
+		),
+	)
+}
+
+func (v *View) Resize() {
```
pteichman, line 211: This might be more long-term flexible if it took width & height as function arguments rather than getting the current whole terminal dimensions from `ui`.

pkg/stats/view.go
```diff
@@ -0,0 +1,296 @@
...
+	termWidth, termHeight := ui.TerminalDimensions()
+	v.Grid.SetRect(0, 0, termWidth, termHeight)
+}
+
+func (v *View) Render() {
+	ui.Clear()
+	ui.Render(v.Grid)
+}
+
+func (v *View) UpdateStats(block realtimeResponseData) error {
```
pteichman, line 221: This should probably be a `*realtimeResponseData`: doing it this way copies the struct as it's passed, and it has enough of a mix of pointers and values internally that it could get confusing.

phamann:
Closing until we discuss more internally. Sorry, @simonwistow 😞

PR #96, opened 2020-05-27
Adds Google Cloud Pub/Sub logging endpoint support

mccurdyc:

* [Fastly API Docs](https://developer.fastly.com/reference/api/logging/google-pubsub/)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/c2c2c62210800daab641161412fbe210004aac0a/fastly/pubsub.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-pubsub && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/googlepubsub/root.go
+++ b/pkg/logging/googlepubsub/root.go
@@ -1,0 +1,28 @@
+package googlepubsub
+
+import (
+	"io"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/config"
+)
+
+// RootCommand is the parent command for all subcommands in this package.
+// It should be installed under the primary root command.
+type RootCommand struct {
+	common.Base
+	// no flags
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("googlepubsub", "Manipulate Fastly service version Google Cloud Pub/Sub logging endpoints")
+	return &c
+}
+
+// Exec implements the command interface.
+func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
+	panic("unreachable")
+}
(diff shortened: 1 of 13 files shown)


pkg/mock/api.go
```diff
@@ -814,6 +820,31 @@ func (m API) DeleteDatadog(i *fastly.DeleteDatadogInput) error {
 	return m.DeleteDatadogFn(i)
 }
 
+// CreatePubsub implements Interface.
+func (m API) CreatePubsub(i *fastly.CreatePubsubInput) (*fastly.Pubsub, error) {
+	return m.CreatePubsubFn(i)
+}
+
+// ListPubsub implements Interface.
```
phamann, line 828:
```suggestion
// ListPubsubs implements Interface.
```

pkg/logging/googlepubsub/root.go
```diff
@@ -0,0 +1,28 @@
...
+type RootCommand struct {
+	common.Base
+	// no flags
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("googlepubsub", "Manipulate Fastly service version Google Cloud Pub/Sub logging endpoints")
```
phamann, line 21: The API endpoint is called `pubsub`, however that is a bit too generic for my liking, so I'm happy to keep as `googlepubsub`. But noting the inconsistency we may introduce if we do.

mccurdyc: @phamann yeah, I dislike the inconsistency with go-fastly. But I think we are consistent in the CLI in regards to generic, names (e.g., "blobstorage" in go-fastly is "azureblob" in the CLI).

One of the reasons why I think it is important to disambiguate is because I could see us adding another logging endpoint with "blob storage" or "pub sub" in the name in the future.

mccurdyc: I do think if we go the route of being consistent with go-fastly in regards to googlepubsub, we should do the same for azureblob.

mccurdyc: I am going to leave as googlepubsub to remain consistent in the CLI.

PR #97, opened 2020-05-27
Update Syslog to use new Create/Update pattern

mccurdyc:
## Proposed Change(s)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-test-syslog && \
    make test && \
    rm -rf /tmp/cli \
)
```

## Notes

 * The missing `Hostname` and `IPV4` fields are because the `Address` field can be used instead.

### DIFF
--- a/pkg/logging/syslog/syslog_integration_test.go
+++ b/pkg/logging/syslog/syslog_integration_test.go
@@ -1,0 +1,449 @@
+package syslog_test
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/fastly/cli/pkg/app"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/mock"
+	"github.com/fastly/cli/pkg/testutil"
+	"github.com/fastly/cli/pkg/update"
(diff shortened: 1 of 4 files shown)


pkg/logging/syslog/update.go
```diff
@@ -102,10 +108,6 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 		input.NewName = c.NewName.Value
 	}
 
-	if c.NewName.Valid {
```
mccurdyc, line 105: This is a duplicate. See above.

pkg/logging/syslog/syslog_integration_test.go
```diff
@@ -0,0 +1,449 @@
+package syslog_test
```
mccurdyc, line 1: This is a direct `mv syslog_test.go syslog_integration_test.go`.

Note that fields like `Hostname` and `IPV4` exist, but can't be set.

This is addressed in a separate branch.

pkg/logging/syslog/update.go
phamann: Nice catch!

mccurdyc:
Force-pushed rebasing master.

PR #98, opened 2020-05-27
Update S3 to use new Create/Update pattern

mccurdyc:
## Proposed Change(s)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-test-s3 && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/s3/s3_integration_test.go
+++ b/pkg/logging/s3/s3_integration_test.go
@@ -1,0 +1,453 @@
+package s3_test
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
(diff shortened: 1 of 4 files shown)


pkg/logging/s3/s3_integration_test.go
```diff
@@ -0,0 +1,453 @@
+package s3_test
```
mccurdyc, line 1: This is a direct `mv s3_test.go s3_integration_test.go`.

mccurdyc:
Force-pushed rebasing master.

PR #101, opened 2020-06-01
Update S3 & BigQuery root docs to be consistent

mccurdyc:
## Proposed Change(s)

Update S3 and BigQuery root command docs to be consistent with all other logging
endpoints and docs in general, by removing trailing '.' (period).

## Validation

```bash
make test
```


mccurdyc:
Force-pushed rebasing master.

PR #103, opened 2020-06-01
Update Sumologic to use new Create/Update pattern

mccurdyc:
## Proposed Change(s)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-test-sumologic && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/sumologic/sumologic_integration_test.go
+++ b/pkg/logging/sumologic/sumologic_integration_test.go
@@ -1,0 +1,386 @@
+package sumologic_test
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/fastly/cli/pkg/app"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/mock"
+	"github.com/fastly/cli/pkg/testutil"
+	"github.com/fastly/cli/pkg/update"
+	"github.com/fastly/go-fastly/fastly"
+)
+
+func TestSumologicCreate(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/logging/sumologic/sumologic_integration_test.go
```diff
@@ -0,0 +1,386 @@
+package sumologic_test
```
mccurdyc, line 1: This is a direct mv sumologic_test.go sumologic_integration_test.go

pkg/logging/sumologic/update.go
```diff
@@ -16,10 +16,12 @@ type UpdateCommand struct {
...
-	Input fastly.GetSumologicInput
-
-	NewName common.OptionalString
+	// required
+	EndpointName string // Can't shaddow common.Base method Name().
+	Version      int
 
+	// optional
+	NewName           common.OptionalString
 	Address           common.OptionalString
```
mccurdyc, line 25: @phamann TODO: this field is extraneous, but not removing as part of this change.

https://developer.fastly.com/reference/api/logging/sumologic/

mccurdyc: https://github.com/fastly/cli/pull/103/commits/2a0acd2de3e716ffe19c138bab1041bc0ca8252e

phamann: :+1:

mccurdyc: I'll pull into a separate PR just for clarity.

mccurdyc: https://github.com/fastly/cli/pull/104

PR #104, opened 2020-06-01
Updates Sumologic to remove extraneous Address field

mccurdyc:
I noted [here](https://github.com/fastly/cli/pull/103#discussion_r433282335) that this field is extraneous and is not exposed by the Fastly API ([docs](https://developer.fastly.com/reference/api/logging/sumologic/)), but is however exposed by the [fastly/go-fastly client library](https://github.com/fastly/go-fastly/blob/8c5f5b72363930ec9ba4de2ae53a5f7751a8b68a/fastly/sumologic.go#L16).

I validated that the API code does not expose this field.

### DIFF
--- a/pkg/logging/sumologic/update.go
+++ b/pkg/logging/sumologic/update.go
@@ -22,7 +22,6 @@ type UpdateCommand struct {
 
 	// optional
 	NewName           common.OptionalString
-	Address           common.OptionalString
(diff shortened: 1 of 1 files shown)


mccurdyc:
Force-pushed rebasing master.

mccurdyc:
Force-pushed rebasing master.

PR #107, opened 2020-06-01
Update Logentries to use new Create/Update pattern

mccurdyc:
## Proposed Change(s)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-test-logentries && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/logentries/create.go
+++ b/pkg/logging/logentries/create.go
@@ -15,11 +15,19 @@ import (
 type CreateCommand struct {
 	common.Base
 	manifest manifest.Data
-	Input    fastly.CreateLogentriesInput
 
-	// We must store all of the boolean flags seperatly to the input structure
-	// so they can be casted to go-fastly's custom `Compatibool` type later.
-	useTLS bool
+	// required
+	EndpointName string // Can't shaddow common.Base method Name().
+	Version      int
+
+	// optional
+	Port              common.OptionalUint
+	UseTLS            common.OptionalBool
+	Token             common.OptionalString
+	Format            common.OptionalString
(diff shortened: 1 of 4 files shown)


pkg/logging/logentries/create.go
```diff
@@ -15,11 +15,19 @@ import (
 type CreateCommand struct {
 	common.Base
 	manifest manifest.Data
-	Input    fastly.CreateLogentriesInput
 
-	// We must store all of the boolean flags seperatly to the input structure
-	// so they can be casted to go-fastly's custom `Compatibool` type later.
-	useTLS bool
```
mccurdyc, line 22: Yay! No more this.

pkg/logging/logentries/logentries_integration_test.go
```diff
@@ -0,0 +1,389 @@
+package logentries_test
```
mccurdyc, line 1: Straight `mv logentries_test.go logentries_integration_test.go`

pkg/logging/logentries/update.go
```diff
@@ -85,10 +90,6 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 		input.NewName = c.NewName.Value
 	}
 
-	if c.NewName.Valid {
```
mccurdyc, line 88: Removed a duplicate set.

phamann: Nice catch!

mccurdyc:
Force-pushed rebasing master.

PR #108, opened 2020-06-01
Update GCS to use new Create/Update pattern

mccurdyc:
## Proposed Change(s)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-test-gcs && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/gcs/create.go
+++ b/pkg/logging/gcs/create.go
@@ -15,7 +15,24 @@ import (
 type CreateCommand struct {
 	common.Base
 	manifest manifest.Data
-	Input    fastly.CreateGCSInput
+
+	// required
+	EndpointName string // Can't shaddow common.Base method Name().
+	Version      int
+	Bucket       string
+	User         string
+	SecretKey    string
+
+	// optional
+	Path              common.OptionalString
+	Period            common.OptionalUint
+	GzipLevel         common.OptionalUint8
+	Format            common.OptionalString
+	FormatVersion     common.OptionalUint
+	MessageType       common.OptionalString
+	ResponseCondition common.OptionalString
+	TimestampFormat   common.OptionalString
+	Placement         common.OptionalString
 }
 
 // NewCreateCommand returns a usable command registered under the parent.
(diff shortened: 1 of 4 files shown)


pkg/logging/gcs/gcs_integration_test.go
```diff
@@ -0,0 +1,436 @@
+package gcs_test
```
mccurdyc, line 1: Straight `mv gcs_test.go gcs_integration_test.go`

pkg/logging/gcs/create.go
```diff
@@ -25,36 +42,90 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
...
+
+	c.CmdClause.Flag("period", "How frequently log files are finalized so they can be available for reading (in seconds, default 3600)").Action(c.Period.Set).UintVar(&c.Period.Value)
+	c.CmdClause.Flag("path", "The path to upload logs to (default '/')").Action(c.Path.Set).StringVar(&c.Path.Value)
+	c.CmdClause.Flag("gzip-level", "What level of GZIP encoding to have when dumping logs (default 0, no compression)").Action(c.GzipLevel.Set).Uint8Var(&c.GzipLevel.Value)
+	c.CmdClause.Flag("format", "Apache style log formatting").Action(c.Format.Set).StringVar(&c.Format.Value)
+	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint. Can be either 2 (the default, version 2 log format) or 1 (the version 1 log format). The logging call gets placed by default in vcl_log if format_version is set to 2 and in vcl_deliver if format_version is set to 1").Action(c.FormatVersion.Set).UintVar(&c.FormatVersion.Value)
+	c.CmdClause.Flag("message-type", "How the message should be formatted. One of: classic (default), loggly, logplex or blank").Action(c.MessageType.Set).StringVar(&c.MessageType.Value)
```
mccurdyc, line 58: Making note that `message-type` already existed in Creates, but is not yet exposed by the [Go client library](https://github.com/fastly/go-fastly/blob/master/fastly/gcs.go#L160) for Updates.

This will be fixed in a subsequent round of changes addressing these missing pieces.

phamann: Might be worth adding a `TODO` comment in the update file so it can be caught?

PR #111, opened 2020-06-02
Adds ResponseCondition and Placement fields to BigQuery Creates

mccurdyc:
## Proposed Change(s)

* Adds `ResponseCondition` and `Placement` fields to BigQuery Creates.

## Relevant Link(s)/Documentation

* [API Docs](https://developer.fastly.com/reference/api/logging/bigquery/)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/34ee78b782c1a7bd2427224fee6ad8c3fb584c2e/fastly/bigquery.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-bq-add-fields && \
    make test && \
    rm -rf /tmp/cli \
)
```


pkg/logging/bigquery/create.go
```diff
@@ -87,8 +89,12 @@ func (c *CreateCommand) createInput() (*fastly.CreateBigQueryInput, error) {
 		input.FormatVersion = c.FormatVersion.Value
 	}
 
-	if c.FormatVersion.Valid {
```
phamann, line 90: Huh we had a duplicate in here? Good catch!

mccurdyc:
Force-pushed rebasing master.

PR #113, opened 2020-06-04
Adds MessageType field to GCS Updates

mccurdyc:
## Proposed Change(s)

* Add MessageType field to GCS Updates.

* [Fastly API Docs](https://developer.fastly.com/reference/api/logging/gcs/)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/fe7cd620035a0ab3980f93d9180df0db7c4f93c4/fastly/gcs.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/gcs-messagetype && \
    make test && \
    rm -rf /tmp/cli \
)
```


mccurdyc:
Force-pushed rebasing master.

PR #114, opened 2020-06-04
Adds PublicKey field to S3

mccurdyc:
## Proposed Change(s)

* Add PublicKey field to all S3 CRUD operations.

* [Fastly API Docs](https://developer.fastly.com/reference/api/logging/s3/)
* [fastly/go-fastly](https://github.com/fastly/go-fastly/blob/fe7cd620035a0ab3980f93d9180df0db7c4f93c4/fastly/s3.go)

## Validation

```
$ make test
```

```
(
    cd /tmp/cli && \
    git checkout mccurdyc/logging-s3-publickey && \
    make test && \
    rm -rf /tmp/cli \
)
```

### DIFF
--- a/pkg/logging/s3/s3_test.go
+++ b/pkg/logging/s3/s3_test.go
@@ -1,6 +1,7 @@
 package s3
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/fastly/cli/pkg/common"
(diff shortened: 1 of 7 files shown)


pkg/logging/s3/s3_test.go
```diff
@@ -193,27 +198,10 @@ func createCommandMissingServiceID() *CreateCommand {
 
 func updateCommandNoUpdates() *UpdateCommand {
 	return &UpdateCommand{
-		Base:                         common.Base{Globals: &config.Data{Client: nil}},
```
mccurdyc, line 196: Having these values explicitly set in the test didn't identify this when it was missing.

https://github.com/fastly/cli/pull/114/files#diff-9952ff67a00a160332b383c2329448aeR114

mccurdyc:
Force-pushed rebasing master.

PR #119, opened 2020-06-05
v0.12.0

mccurdyc:
### Proposed Change(s)

* Updates the changelog for upcoming v0.12.0 release.

PR #120, opened 2020-06-05
docs/release: Noting non-determinism of changelog

mccurdyc:
Seems like the GitHub API is not deterministic and picks up commits differently, so just running the changelog generator until I get the expected.

More conversation here - https://github.com/fastly/cli/pull/119

I'm thinking it may be related to - github-changelog-generator/github-changelog-generator#580


mccurdyc:
Force-pushed rebasing master.

PR #124, opened 2020-06-09
Refactor fastly crate verification to only enforce fastly-sys version

phamann:
### TL;DR
Refactors the crate verification introduced in #67 to only fail the build if a user has an out of range `fastly-sys` transient dependency and not an out of range `fastly` crate. This should reduce the frequency in which a user sees this message.

### Why?
Now that the ABI work has been abstracted out of the `fastly` crate to `fastly-sys`, we can refactor the CLI verification so that in only blocks builds when there is a semver-incompatible new version of the `fastly-sys` crate available. Which will be updated less frequently than the `fastly` crate itself. 

The simplest remediation to offer in this case is to recommend an upgrade to the latest fastly crate, since users are unlikely to be depending on fastly-sys directly. We could get fancy about the suggested remediations in case we decide to backport dependency upgrades to older major versions of the fastly crate, but let's stick to the simpler remediation for now.

We decided to have keep the version check for new versions of the `fastly` crate, but only emit an info message rather than blocking the build when a new one is available.

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -34,8 +34,9 @@ const (
 // package which we are interested in and is embedded within CargoManifest and
 // CargoLock.
 type CargoPackage struct {
-	Name    string `toml:"name"`
-	Version string `toml:"version"`
+	Name         string         `toml:"name" json:"name"`
+	Version      string         `toml:"version" json:"version"`
+	Dependencies []CargoPackage `toml:"-" json:"dependencies"`
 }
 
 // CargoManifest models the package confuiguration properties of a Rust Cargo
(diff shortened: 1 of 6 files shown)


pkg/compute/rust.go
```diff
@@ -170,88 +183,57 @@ func (r Rust) Verify(out io.Writer) error {
...
 	}
 
-	// Parse the lowest minor semver for the latest release.
-	// I.e. v0.3.2 becomes v0.3.0.
-	latestMinor, err := semver.Parse(fmt.Sprintf("%d.%d.0", latest.Major, latest.Minor))
-	if err != nil {
-		return fmt.Errorf("error parsing Cargo.lock file: %w", err)
+	// If fastly-sys version is on a lower than the latest, error with dual
+	// remediation steps.
+	if fastlySysVersion.LT(*latestFastlySys) {
```
acfoltzer, line 216: I tried reading the docs for this semver library and couldn't tell whether this is a strict less-than or a compatibility-respecting less-than. For example, we would want this to return false for `"0.3.4".LT("0.3.5")` but not `"0.3.4".LT("0.4.0")`

phamann: Good catch. I've refactored to use a different semver library which supports constraint ranges. Let me know what you think.

pkg/compute/rust.go
```diff
@@ -205,15 +206,22 @@ func (r Rust) Verify(out io.Writer) error {
 		return fmt.Errorf("error fetching latest crate version: %w", err)
 	}
 
+	// Create a semver contraint to be within the latest minor range or above.
+	// TODO(phamann): Update this to major when fastly-sys hits 1.x.x.
+	fastlySysConstraint, err := semver.NewConstraint(fmt.Sprintf("~%d.%d.0", latestFastlySys.Major(), latestFastlySys.Major()))
```
acfoltzer, line 211: Perhaps?
```suggestion
	fastlySysConstraint, err := semver.NewConstraint(fmt.Sprintf("~%d.%d.0", latestFastlySys.Major(), latestFastlySys.Minor()))
```

phamann: 🙈

pkg/compute/rust.go
```diff
@@ -170,88 +184,64 @@ func (r Rust) Verify(out io.Writer) error {
...
-				text.Bold(fmt.Sprintf(`fastly = "^%s"`, latest)),
-				text.Bold("cargo update -p fastly"),
-			),
-		}
+	// If fastly crate version is lower than the latest, suggest user should
+	// update, but don't error.
+	if fastlyVersion.LessThan(latestFastly) {
+		text.Break(out)
+		text.Info(out, fmt.Sprintf(
+			"a newer version of the fastly crate is avaiable, edit %s with:\n\n\t %s\n\nAnd then run the following command:\n\n\t$ %s\n",
```
acfoltzer, line 239: Trying to emphasize they don't have to do this immediately if the `fastly-sys` dependency is okay.
```suggestion
			"an optional upgrade for the fastly crate is available, edit %s with:\n\n\t %s\n\nAnd then run the following command:\n\n\t$ %s\n",
```

PR #125, opened 2020-06-09
Allow compute services to be initialised from an existing service ID

phamann:
### TL;DR
Refactors  `compute init` to accept a `--service-id` flag which will be used instead of provisioning a new service. This is an important feature request to bridge the UX flow between the UI -> CLI, i.e. this allows a user to first create a service via the UI and then move to the CLI to initialise the package.

### DIFF
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -52,6 +52,7 @@ var (
 // InitCommand initializes a Compute@Edge project package on the local machine.
 type InitCommand struct {
 	common.Base
+	serviceID   string
 	name        string
 	description string
 	author      string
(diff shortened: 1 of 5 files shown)


pkg/compute/init.go
```diff
@@ -67,6 +68,7 @@ func NewInitCommand(parent common.Registerer, globals *config.Data) *InitCommand
 	var c InitCommand
 	c.Globals = globals
 	c.CmdClause = parent.Command("init", "Initialize a new Compute@Edge package locally")
+	c.CmdClause.Flag("service-id", "Existing service ID to initialize from").Short('s').StringVar(&c.serviceID)
```
thommahoney, line 71: `Existing service ID to initialize from` reads ambiguously to me. Am I using this service as the basis for a new service? Perhaps "Existing service ID to use. By default, this command creates a new service."

pkg/compute/init.go
```diff
@@ -108,6 +110,19 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	undoStack := common.NewUndoStack()
 	defer func() { undoStack.RunIfError(out, err) }()
 
+	var service *fastly.Service
+	if c.serviceID != "" {
+		service, err = c.Globals.Client.GetService(&fastly.GetServiceInput{
```
thommahoney, line 115: The rest of this file assumes that the version is `1` but that might not be true. We may even need to clone an active service version. Thoughts?

phamann: 😱 Great catch! Yes, this is a very big oversight on my behalf, will fix.

PR #129, opened 2020-06-15
v0.13.0 changelog

phamann:
### TL;DR
- Updates changelog for v0.13.0
- Fixes `make changelog` PWD logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,8 +63,8 @@ install:
 
 .PHONY: changelog
 changelog:
-	@$(CURRENT_DIR)/scripts/changelog.sh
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -63,8 +63,8 @@ install:
 
 .PHONY: changelog
 changelog:
-	@$(CURRENT_DIR)/scripts/changelog.sh
+	@$(shell pwd)/scripts/changelog.sh
```
phamann, line 66: @mccurdyc FYI. Assuming `CURRENT_DIR` may be specific to you environment or Arch linux? Just checking you're ok with using `pwd` here instead?

PR #133, opened 2020-06-25
Bump Rust default template version to v0.4.0

phamann:
### TL:DR
Bumps the Rust default template which is cloned during `fastly compute init` to v0.4.0

### Note
We want to move away from hard coding this value and may move to dynamic configuration via an API request in the future.


phamann:
Merging as admin.

PR #139, opened 2020-07-08
Compare package hashsum during deployment

phamann:
### TL;DR
Refactors `fastly compute deploy` to use `gofastly.UpdatePackage()` methods now that we have upstreamed the API endpoints to `go-fastly` API client. Also adds new functionality to prevent a deploy if the package hashsum matches the package that already exists on the service.

### DIFF
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -1,8 +1,10 @@
 package compute
 
 import (
+	"crypto/sha512"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"sort"
 
@@ -20,7 +22,7 @@ type DeployCommand struct {
 	common.Base
 	manifest manifest.Data
 	path     string
-	version  int
+	version  common.OptionalInt
 }
 
 // NewDeployCommand returns a usable command registered under the parent.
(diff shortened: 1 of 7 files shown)


pkg/compute/deploy.go
```diff
@@ -92,34 +83,58 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+		Version: c.version,
+	}); err == nil {
+		hashSum, err := getHashSum(c.path)
+		if err != nil {
+			return fmt.Errorf("error getting package hashsum: %w", err)
+		}
+
+		if hashSum == p.Metadata.HashSum {
+			progress.Done()
+			text.Info(out, "Skipping package deployment, local and service version hashsums are identical. (service %v, version %v) ", serviceID, c.version)
```
thommahoney, line 113: Perhaps the customer doesn't care how we're determining that the local and remote packages are the same. ie. We can skip the bit about the "hashsum?" Seems fine either way, though.

How likely is it that the package will be the same?

phamann: :+1: Have removed the "hashsum" portion of the message.

> How likely is it that the package will be the same?

Not that likely, however this ensures running `fastly compute deploy` twice is idempotent and at the very least should hint to users that they may have forgotten to  run `fastly compute build` after changing some code.

PR #143, opened 2020-07-17
Fix bug where name was not added to the manifest

fgsch:
Use the default name if the user press enter at the name prompt.

### DIFF
--- a/pkg/compute/compute_integration_test.go
+++ b/pkg/compute/compute_integration_test.go
@@ -273,6 +273,19 @@ func TestInit(t *testing.T) {
 				"Updating package manifest...",
 			},
 		},
+		{
+			name:       "with default name inferred from directory",
+			args:       []string{"compute", "init"},
+			configFile: config.File{Token: "123"},
+			api: mock.API{
+				GetTokenSelfFn:  tokenOK,
+				GetUserFn:       getUserOk,
+				CreateServiceFn: createServiceOK,
+				CreateDomainFn:  createDomainOK,
+				CreateBackendFn: createBackendOK,
+			},
+			manifestIncludes: `name = "fastly-build`,
+		},
 	} {
 		t.Run(testcase.name, func(t *testing.T) {
 			// We're going to chdir to an init environment,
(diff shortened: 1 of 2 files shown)


fgsch:
Not sure how I can write a test but haven't checked in depth yet.

pkg/compute/init.go
```diff
@@ -151,13 +151,14 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	c.path = abspath
 
 	if name == "" {
-		name = filepath.Base(c.path)
-		fmt.Fprintf(progress, "--name not specified, using %s\n\n", name)
```
phamann, line 155: Would be good to keep the verbose output here.

fgsch: None of the other options that have defaults show this and the message is incorrect since this is not necessarily the name you will end up using so I think removing this is better.

pkg/compute/compute_integration_test.go
```diff
@@ -255,7 +255,10 @@ func TestInit(t *testing.T) {
 				CreateDomainFn:  createDomainOK,
 				CreateBackendFn: createBackendOK,
 			},
-			manifestIncludes: `authors = ["test@example.com"]`,
+			manifestIncludes: []string{
+				`authors = ["test@example.com"]`,
+				`name = "fastly-build"`,
```
phamann: As this is an `AssertStringContains` string contains match and we're only matching on the prefix of the name, you'll need to drop the closing `"` here for this to pass.

fgsch: Yup. I noticed after I pushed it. Anyways, I've reverting that change and add another test as you originally suggested.

fgsch:
> Looks like you've got some failing tests after: [50360af](https://github.com/fastly/cli/commit/50360af1ac70c1447e129e25a5c2260d5391f6d4)

Yup, I will let you know when this is ready for another review. Until then please ignore this pr.

fgsch:
Updated and ready for another review.

PR #156, opened 2020-09-24
Bump supported Rust toolchain version to 1.46

phamann:
### TL;DR
Bumps the supported Rust toolchain version for Compute@Edge programs to the stable `1.46.0` version.

### Notes
We intend to move away from fixed versions to a min-max range for more flexibility in the near future.


katie-martin-fastly (approved):
LGTM! :heavy_check_mark: 

I've created a matching PR over in the default template, and I'll open another for the beacon termination template shortly.

PR #160, opened 2020-10-23
Add AssemblyScript support to compute init and build commands

phamann:
### TL;DR
Adds beta AssemblyScript toolchain support to the Compute@Edge commands `fastly compute init` and `fastly compute build`. It uses `npm` to validate the package and environment and the `asc` toolchain for compilation.

### What
- Refactor `pkg/compute/init.go` and `pkg/compute/build.go` to support multiple languages.
- Extract common streaming execution logic used by toolchain compilers to `pkg/common/exec`.
- Extend the `Toolchain` interface to  better cater for  multiple languages. 
- Refactor `pkg/compute/rust.go` to support new toolchain interface additions.
- Add new AssemblyScript struct which implements `Toolchain` interface.
- Add support to clone startkit templates from git tags instead of branches.
- Tests!


pkg/compute/build.go
```diff
@@ -23,6 +23,12 @@ const IgnoreFilePath = ".fastlyignore"
...
 type Toolchain interface {
+	Name() string
+	DisplayName() string
+	StarterKits() []StarterKit
+	SourceDirectory() string
+	IncludeFiles() []string
+	Initialize(out io.Writer) error
 	Verify(out io.Writer) error
 	Build(out io.Writer, verbose bool) error
 }
```
phamann, line 34: I don't like the use of interface methods here to model getters for simple values. Would this be better modelled as a `Language` struct with an embedded `Toolchain` interface? Such as:

```suggestion
type Toolchain interface {
	Initialize(out io.Writer) error
	Verify(out io.Writer) error
	Build(out io.Writer, verbose bool) error
}

type Language struct {
    Name string
    DisplayName string
    StarterKits []StarterKit
    IncludeFiles []string
    
    Toolchain
}

func NewLangauge(name, displayName, string, kits []starterKits, includeFiles []string, toolchain Toolchain) *Language {
    // ... etc
}
```

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// AssemblyScript is an implments Toolchain for the AssemblyScript lanaguage.
```
thommahoney, line 16: this comment is mildly confusing

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+	// needed to assert that the correct versions of the asc compiler and
+	// @fastly/as-compute package are installed. We only check whether the
+	// binary exists on the users $PATH and error with installation help text.
+	fmt.Fprintf(out, "Checking if npm is installed...\n")
+
+	p, err := exec.LookPath("npm")
+	if err != nil {
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("`npm` not found in $PATH"),
+			Remediation: fmt.Sprintf("To fix this error, install Node.sj and npm by visiting:\n\n\t$ %s", text.Bold("https://nodejs.org/")),
```
thommahoney, line 63: spelling

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+			Remediation: fmt.Sprintf("To fix this error, run the following command:\n\n\t$ %s", text.Bold("npm install --save-dev assemblyscript")),
+		}
+	}
+
+	fmt.Fprintf(out, "Found asc at %s\n", path)
+
+	return nil
+}
+
+// Initialize implments the Toolchain interface and initializes a newly cloned
```
thommahoney, line 124: spelling

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("package.json not found"),
+			Remediation: fmt.Sprintf("To fix this error, run the following command:\n\n\t$ %s", text.Bold("npm init")),
+		}
+	}
+
+	fmt.Fprintf(out, "Found package.json at %s\n", fpath)
+
+	// Call npm install.
+	cmd := common.NewStreamingExec("npm", []string{"install"}, []string{}, false, out)
```
thommahoney, line 163: perhaps this deserves some user messaging? eg. "installing package dependencies..."

thommahoney (commented):
I like the new abstraction around languages and toolchains 👍

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return fmt.Errorf("getting package.json path: %w", err)
+	}
+
+	if !common.FileExists(fpath) {
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("package.json not found"),
+			Remediation: fmt.Sprintf("To fix this error, run the following command:\n\n\t$ %s", text.Bold("npm init")),
+		}
+	}
```
mccurdyc: suggestion - What about pulling this chunk out into a separate function that `Initialize` and `Verify` could use?

This would also reduce having to update error messages, etc. in two places and make this logic easier to test.

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+
+	fmt.Fprintf(out, "Found package.json at %s\n", fpath)
+
+	// Call npm install.
+	cmd := common.NewStreamingExec("npm", []string{"install"}, []string{}, false, out)
+	if err := cmd.Exec(); err != nil {
+		return err
+	}
+
+	return nil
```
mccurdyc, line 168: nitpick - 
```suggestion
	return cmd.Exec()	
```

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	binDir := filepath.Join(pwd, "bin")
+	if err := common.MakeDirectoryIfNotExists(binDir); err != nil {
+		return fmt.Errorf("error making bin directory: %w", err)
+	}
+
+	npmdir, err := getNpmBinPath()
+	if err != nil {
+		return err
+	}
```
mccurdyc, line 187: suggestion - thoughts on pulling these "metadata"-y bits into some centralized "initialization" function that gets called once and sets some "metadata" fields on like a config struct or something so that we don't have to have these checks in multiple places.

An additional benefit of this would be that the functions are more focused on what they are supposed to be doing --- e.g., "building" --- and less on validating the environment.

---

A valid counter argument to this would be then you have dependencies on the metadata fields being set and the functions --- e.g., `Build` --- are not entirely self-contained.

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,230 @@
...
+	// Disabling as the variables come from trusted sources.
+	/* #nosec */
+	cmd := exec.Command("npm", "link", "--json", "--depth", "0", name)
+	if err := cmd.Start(); err != nil {
+		return false
+	}
+	if err := cmd.Wait(); err != nil {
+		return false
+	}
+	return true
```
mccurdyc, line 231: suggestion - I think this could be simplified to
```suggestion
return exec.Command("npm", "link", "--json", "--depth", "0", name).Run() == nil
```

mccurdyc: I don't know that I'm a fan of condensing into a single line, but maybe
```suggestion
	cmd := exec.Command("npm", "link", "--json", "--depth", "0", name)
return cmd.Run() == nil
```

pkg/compute/build.go
mccurdyc: `NewLangauge` -> `NewLanguage` in your suggestion :)

I personally like the Language struct idea because this `NewLanguage` function could be where the "metadata"-y bits that I was talking about could be initialized and verified.

pkg/compute/compute_integration_test.go
```diff
@@ -346,8 +360,8 @@ func TestInit(t *testing.T) {
 	}
 }
 
-func TestBuild(t *testing.T) {
-	if os.Getenv("TEST_COMPUTE_BUILD") == "" {
+func TestBuildRust(t *testing.T) {
+	if os.Getenv("TEST_COMPUTE_BUILD_RUST") == "" && os.Getenv("TEST_COMPUTE_BUILD") == "" {
```
mccurdyc, line 364: Was `TEST_COMPUTE_BUILD_RUST` something that already existed previously for testing?

pkg/compute/compute_integration_test.go
```diff
@@ -469,6 +483,93 @@ func TestBuild(t *testing.T) {
 	}
 }
 
+func TestBuildAssemblyScript(t *testing.T) {
+	if os.Getenv("TEST_COMPUTE_BUILD_ASSEMBLYSCRIPT") == "" && os.Getenv("TEST_COMPUTE_BUILD") == "" {
```
mccurdyc, line 487: Personally, I think it would be kind of annoying to always minimally have to supply two environment variables for testing.

I definitely don't think we want to be string comparing some value in `TEST_COMPUTE_BUILD`, but rather using the other language-specific boolean envs and dropping `TEST_COMPUTE_BUILD` since that would be implied.

mccurdyc: I also don't really have the context for what either of these do, so I could be overlooking something.

mccurdyc (commented):
LGTM - a couple suggestions and I need to refresh, but didn't want to lose my comments.

I'm paranoid :)

pkg/compute/compute_integration_test.go
phamann: No this has been added so you can run the Rust tests in isolation. See other comment RE `TEST_COMPUTE_BUILD`.

pkg/compute/compute_integration_test.go
phamann: >Personally, I think it would be kind of annoying to always minimally have to supply two environment variables for testing.

The logic here may be confusing. You only need to supply `TEST_COMPUTE_BUILD` if you want to run all of the build tests i.e. rust and assemblyscript. Which is what is already happening in CI.

This logic skips the test if both `TEST_COMPUTE_BUILD_RUST` and `TEST_COMPUTE_BUILD` are set. i.e. it gives us the flexibility to only run a single toolchain test or all of them.

pkg/compute/assemblyscript.go
phamann: I went back and forwards between extracting the shared check in the AS Initialize() and Verify() methods. However opted for the duplication to allow them to remain completely independent of each other, as they’re called by different CLI commands.

pkg/common/exec.go
```diff
@@ -0,0 +1,91 @@
...
+	var errStdout, errStderr error
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		_, errStdout = io.Copy(stdout, stdoutIn)
+		wg.Done()
+	}()
+
+	_, errStderr = io.Copy(stderr, stderrIn)
```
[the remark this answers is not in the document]
phamann: Good catch!

pkg/compute/build.go
peterbourgon: @phamann I like your alternative — structs collect data, and interfaces collect behavior, so your type Language struct seems to best model what's actually going on.

pkg/common/file.go
```diff
@@ -80,3 +80,23 @@ func CopyFile(src, dst string) (err error) {
...
+	fi, err := os.Stat(path)
+	switch {
+	case err == nil && fi.IsDir():
+		return nil
+	case err == nil && !fi.IsDir():
+		return fmt.Errorf("%s already exists as a regular file", path)
+	case os.IsNotExist(err):
+		if err := os.MkdirAll(path, 0750); err != nil {
+			return err
+		}
```
peterbourgon, line 96:
```suggestion
		return os.MkdirAll(path, 0750)
```

peterbourgon: Is 0750 defined as a const somewhere?

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,222 @@
...
+		return fmt.Errorf("error getting current working directory: %w", err)
+	}
+	binDir := filepath.Join(pwd, "bin")
+	if err := common.MakeDirectoryIfNotExists(binDir); err != nil {
+		return fmt.Errorf("error making bin directory: %w", err)
+	}
+
+	npmdir, err := getNpmBinPath()
+	if err != nil {
+		return err
```
peterbourgon, line 182: Suggestion: if one path in a function annotates errors, all paths should, e.g.

```suggestion
		return fmt.Errorf("getting npm path: %w", err)
```

pkg/compute/assemblyscript.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return nil
+}
+
+func getNpmBinPath() (string, error) {
+	path, err := exec.Command("npm", "bin").Output()
+	if err != nil {
+		return "", fmt.Errorf("error getting npm bin path: %w", err)
```
peterbourgon, line 210: If you took my earlier suggestion, this could then just `return "", err`.

pkg/compute/build.go
```diff
@@ -99,6 +105,8 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 
 	var toolchain Toolchain
 	switch lang {
+	case "assemblyscript":
+		toolchain = &AssemblyScript{}
```
peterbourgon, line 109: Just to reinforce the earlier point, an empty `struct{}` used only to hang methods that implement some interface is a (very minor!!) yellow flag that the design could be improved.

pkg/compute/compute_integration_test.go
```diff
@@ -955,6 +1056,49 @@ func makeBuildEnvironment(t *testing.T, fastlyManifestContent, cargoManifestCont
...
+	p := make([]byte, 8)
+	n, err := rand.Read(p)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rootdir = filepath.Join(
+		os.TempDir(),
+		fmt.Sprintf("fastly-build-%x", p[:n]),
+	)
```
peterbourgon, line 1071: This was my dumb code initially 🤦 which can be replaced with

```suggestion
	rootdir = ioutil.TempDir("", "fastly-build-")
```

pkg/common/exec.go
```diff
@@ -0,0 +1,91 @@
...
+	/* #nosec */
+	cmd := exec.Command(s.command, s.args...)
+	cmd.Env = append(os.Environ(), s.env...)
+
+	// Pipe the child process stdout and stderr to our own output writer.
+	var stdoutBuf, stderrBuf bytes.Buffer
+	stdoutIn, _ := cmd.StdoutPipe()
+	stderrIn, _ := cmd.StderrPipe()
+	stdout := io.MultiWriter(s.output, &stdoutBuf)
+	stderr := io.MultiWriter(s.output, &stderrBuf)
```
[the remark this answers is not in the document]
peterbourgon: The exit code of any os/exec.Command is captured in the command, but it's up to the caller to act on it, it wouldn't be a function of how the stdout/stderr are managed (if that's what you're implying?)..

pkg/common/exec.go
```diff
@@ -0,0 +1,93 @@
...
+	// Construct the command with given arguments and environment.
+	//
+	// gosec flagged this:
+	// G204 (CWE-78): Subprocess launched with variable
+	// Disabling as the variables come from trusted sources.
+	/* #nosec */
+	cmd := exec.Command(s.command, s.args...)
+	cmd.Env = append(os.Environ(), s.env...)
+
+	// Pipe the child process stdout and stderr to our own output writer.
```
peterbourgon, line 49: As I understand it, the requirements here are

1. Child process stdout and stderr should stream to s.output in real time
2. In case of error and --verbose, stderr should be captured and printed separately

Is that true? Assuming so, I think we might be able to avoid the pipes altogether..

```
cmd := exec.Command(s.command, s.args...)
cmd.Env = append(os.Environ(), s.env...)

var stderrBuf bytes.Buffer
cmd.Stdout = s.output
cmd.Stderr = io.MultiWriter(s.output, stderrBuf)

if err := cmd.Run(); err != nil {
    if s.verbose && stderrBuf.Len() > 0 {
        return ...
    }
    return ...
}

return nil
```

WDYT?

pkg/common/exec.go
```diff
@@ -0,0 +1,93 @@
...
+	if errStdout != nil {
+		return fmt.Errorf("error streaming stdout output from child process: %w", errStdout)
+	}
+	if errStderr != nil {
+		return fmt.Errorf("error streaming stderr output from child process: %w", errStderr)
+	}
+
+	// Wait for the command to exit.
+	if err := cmd.Wait(); err != nil {
+		if !s.verbose && stderrBuf.Len() > 0 {
```
peterbourgon, line 86: I think this condition might be inverted, i.e. do you actually want `if s.verbose` here?

phamann: Nope. In non-verbose mode we use a `common.QuietProgress` writer which swallows anything henece why we need to capture the error and return so it can be written again, whereas in verbose it would have already printed the error. See: https://github.com/fastly/cli/blob/master/pkg/compute/build.go#L55-L60

pkg/common/exec.go
phamann: Nice. Have implemented and tested in: https://github.com/fastly/cli/pull/160/commits/b6776d3b6105ed6df7cc9e03c2d421392933d602

pkg/compute/compute_integration_test.go
phamann: Nice. Have implemented in https://github.com/fastly/cli/pull/160/commits/b6776d3b6105ed6df7cc9e03c2d421392933d602

pkg/compute/build.go
phamann: Ok @mccurdyc @peterbourgon Have implemented this in https://github.com/fastly/cli/pull/160/commits/5d59eb8165d13bb13d31b08e488f7080b0db7ae8 let me know your thoughts?

peterbourgon (approved):

mccurdyc (approved):
I like this!

PR #161, opened 2020-10-26
Support sasl kafka endpoint options in Fastly CLI

kellymclaughlin:
Add the following options to the Kafka logging subcommands in order to support SASL:

* `parse-log-kevvals` - Indicates the log format should be parsed as key-value pairs
* `max-batch-size` -  The maximum size of a log batch in bytes
* `auth-method` - one of: `plain`, `scram-sha256`, `scram-sha512`
* `username` - Username for authentication
* `password` - Password for authentication


mccurdyc:
@kellymclaughlin I haven't reviewed yet, but just noting I _think_ this will require us to update the go-fastly version in the Go module to get the latest version with these new fields, right?

https://github.com/fastly/go-fastly/pull/227

pkg/logging/kafka/create.go
```diff
@@ -61,14 +68,19 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
...
 	c.CmdClause.Flag("placement", "Where in the generated VCL the logging call should be placed, overriding any format_version default. Can be none or waf_debug").Action(c.Placement.Set).StringVar(&c.Placement.Value)
 	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
+	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).BoolVar(&c.ParseLogKeyvals.Value)
+	c.CmdClause.Flag("max-batch-size", "The maximum size of the log batch in bytes").Action(c.RequestMaxBytes.Set).UintVar(&c.RequestMaxBytes.Value)
+	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified.").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
+	c.CmdClause.Flag("auth-method", "SASL authentication method. Valid values are: plain, scram-sha-256, scram-sha-512").Action(c.AuthMethod.Set).StringVar(&c.AuthMethod.Value)
```
phamann, line 74: This could be a enum var:

```suggestion
	c.CmdClause.Flag("auth-method", "SASL authentication method.").Action(c.AuthMethod.Set).EnumVar(&c.AuthMethod.Value, "plain", "scram-sha-256", "scram-sha-512")
```

pkg/logging/kafka/create.go
```diff
@@ -124,6 +136,42 @@ func (c *CreateCommand) createInput() (*fastly.CreateKafkaInput, error) {
...
+			return nil, fmt.Errorf("the --auth-method, --username,  and --password flags must be present when using the --use-sasl flag")
+
+		}
+
+		if c.AuthMethod.Value != "plain" &&
+			c.AuthMethod.Value != "scram-sha-256" &&
+			c.AuthMethod.Value != "scram-sha-512" {
+
+			return nil, fmt.Errorf("invalid SASL authentication method: %s. Valid values are plain, scram-sha-256, and scram-sha-512", c.AuthMethod.Value)
+		}
```
phamann, line 158: This won't be required if we use the `EnumVar` arg type above, as the flag parser handles the validation for you.

pkg/logging/kafka/update.go
```diff
@@ -64,6 +73,14 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
 	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint. Can be either 2 (default) or 1").Action(c.FormatVersion.Set).UintVar(&c.FormatVersion.Value)
 	c.CmdClause.Flag("placement", "Where in the generated VCL the logging call should be placed, overriding any format_version default. Can be none or waf_debug").Action(c.Placement.Set).StringVar(&c.Placement.Value)
 	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
+	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).BoolVar(&c.ParseLogKeyvals.Value)
+	c.CmdClause.Flag("no-parse-log-keyvals", "Disable parsing of key-value pairs within the log format.").Action(c.NoParseLogKeyvals.Set).BoolVar(&c.NoParseLogKeyvals.Value)
```
phamann, line 77: We can collapse these two flags into a [`NegatableBool`](https://pkg.go.dev/gopkg.in/alecthomas/kingpin.v3-unstable#Clause.NegatableBool) which generates both variants of the bool for you `--parse-log-keyvals`, `--no-parse-log-keyvals`.

```suggestion
	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).NegatableBool(&c.ParseLogKeyvals.Value)
```

pkg/logging/kafka/update.go
```diff
@@ -160,6 +182,65 @@ func (c *UpdateCommand) createInput() (*fastly.UpdateKafkaInput, error) {
 		input.Placement = fastly.String(c.Placement.Value)
 	}
 
+	if c.ParseLogKeyvals.Valid && c.NoParseLogKeyvals.Valid {
+		return nil, fmt.Errorf("--parse-key-logvals and --no-parse-key-logvals are mutually exclusive")
+	}
```
phamann, line 187: No longer needed with NegatableBool.

```suggestion

```

pkg/logging/kafka/update.go
```diff
@@ -64,6 +73,14 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
...
 	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
+	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).BoolVar(&c.ParseLogKeyvals.Value)
+	c.CmdClause.Flag("no-parse-log-keyvals", "Disable parsing of key-value pairs within the log format.").Action(c.NoParseLogKeyvals.Set).BoolVar(&c.NoParseLogKeyvals.Value)
+	c.CmdClause.Flag("max-batch-size", "The maximum size of the log batch in bytes").Action(c.RequestMaxBytes.Set).UintVar(&c.RequestMaxBytes.Value)
+	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified.").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
+	c.CmdClause.Flag("disable-sasl", "Disable SASL authentication.").Action(c.DisableSASL.Set).BoolVar(&c.DisableSASL.Value)
+	c.CmdClause.Flag("auth-method", "SASL authentication method. Valid values are: plain, scram-sha-256, scram-sha-512.").Action(c.AuthMethod.Set).StringVar(&c.AuthMethod.Value)
```
phamann, line 81: As per the create command. This can be an enum:
```suggestion
	c.CmdClause.Flag("auth-method", "SASL authentication method.").Action(c.AuthMethod.Set).EnumVar(&c.AuthMethod.Value, "plain", "scram-sha-256", "scram-sha-512")
```

pkg/logging/kafka/update.go
```diff
@@ -160,6 +182,65 @@ func (c *UpdateCommand) createInput() (*fastly.UpdateKafkaInput, error) {
...
+			return nil, fmt.Errorf("the --auth-method, --username,  and --password flags must be present when using the --use-sasl flag")
+
+		}
+
+		if c.AuthMethod.Value != "plain" &&
+			c.AuthMethod.Value != "scram-sha-256" &&
+			c.AuthMethod.Value != "scram-sha-512" {
+
+			return nil, fmt.Errorf("invalid SASL authentication method: %s. Valid values are plain, scram-sha-256, and scram-sha-512", c.AuthMethod.Value)
+		}
```
phamann, line 218: No longer needed:
```suggestion
```

phamann (changes requested):
Thanks for the contribution @kellymclaughlin, looking good overall just want to clean up some of the flag types.  

Also made me realise that Kingpin really needs a `flag.RequiredIfSet(*otherFlag)` flag clause to decorate flag dependencies to avoid us having to write the validation logic for these relationships.

pkg/logging/kafka/kafka_integration_test.go
```diff
@@ -355,6 +378,11 @@ Version: 1
 		Format version: 2
 		Response condition: Prevent default logging
 		Placement: none
+		Parse log key-values: false
+		Max batch size: 0
+		SASL authentication method: `+`
+		SASL authentication username: `+`
+		SASL authentication password: `+`
```
mccurdyc, line 385: question - I'm confused by the `\`+\`` can you explain?

Is it to append a trailing whitespace or something?

I'm assuming that your editor is trimming it or something?

pkg/logging/kafka/kafka_test.go
```diff
@@ -129,6 +204,113 @@ func TestUpdateKafkaInput(t *testing.T) {
...
+				TLSClientKey:      fastly.String("-----BEGIN PRIVATE KEY-----bar"),
+				ParseLogKeyvals:   fastly.CBool(true),
+				RequestMaxBytes:   fastly.Uint(11111),
+				AuthMethod:        fastly.String("scram-sha-512"),
+				User:              fastly.String("user1"),
+				Password:          fastly.String("12345"),
+			},
+		},
+		{
+			name: "verify disabling SASL",
```
mccurdyc, line 237: We may be able to disregard the following because the CLI will require that credentials _always_ be supplied when SASL is enabled, but we should probably confirm that you can't enable, disable and re-renabled without supplying credentials in the API.

---

**TL;DR - we should ensure that when we disable SASL, we removed the stored SASL creds i.e., make sure these zero values make it through to the backend.**

Thanks for including this test!! It got me thinking...

I just want someone else to confirm my thoughts and that the following is true on the following:

1. enable SASL (i.e., send credentials to Fastly)
2. disable SASL 
   * This will "delete" the SASL credentials, right?
3. re-enable SASL, but don't supply credentials.

This may require us to confirm in go-fastly that sending empty string values to go-fastly will actually make it through to the Fastly backend.

---

Basically, I want to ensure that we don't have a possible security issue where if SASL was _ever_ enabled, it couldn't be disabled and then re-enabled without credentials.

pkg/logging/kafka/update.go
mccurdyc: I just wanted to post this because it took me a while to find the docs on this since the library is not a direct import in this file

https://pkg.go.dev/gopkg.in/alecthomas/kingpin.v3-unstable#Clause.NegatableBool

mccurdyc (commented):
I agree with @phamann's comments, but also added a couple comments of my own.

Otherwise, things look good to me.

kellymclaughlin:
> Thanks for the contribution @kellymclaughlin, looking good overall just want to clean up some of the flag types.

> Also made me realise that Kingpin really needs a flag.RequiredIfSet(*otherFlag) flag clause to decorate flag dependencies to avoid us having to write the validation logic for these relationships.

Thanks for looking over this @phamann! I just came across the `NegatableBool` yesterday evening myself and I didn't know about `EnumVar`. I've made those changes. 

As far as Kingpin, I would have definitely enjoyed a `RequiredIfSet` option. I also found the builtin validation stuff in the library not usable in our context which is too bad. I also noticed we're relying on the `v3-unstable` branch of Kingpin and I see that the author has ceased contributing to the library and has created an entirely new library for args parsing. This is my first foray into the cli code, but I wonder if there has been any consideration of moving to a new args parsing library that is more actively supported and offer us some better features? If there is ever discussion of that I'd be happy to be involved and be part of a divide and conquer effort to get it done.

pkg/logging/kafka/kafka_test.go
kellymclaughlin: Agreed, the code to ensure that we remove the stored creds is [here](https://github.com/fastly/cli/pull/161/files#diff-5084619e19b627b734d13406b8130df2ef2018ffa290659bf4d2a1219ee51cdfR189-R193). Right now the API doesn't establish a relationship among any of the Kafka configuration fields and I reflected that in my `go-fastly` changes. I think ideally we'd implement this logic once in the API versus in each consumer, but for now I wanted to deal with it here until I worked out the best way to deal with it more generally.

pkg/logging/kafka/kafka_integration_test.go
kellymclaughlin: Yes, this was evidently my editor doing this. I've removed this now.

pkg/logging/kafka/create.go
```diff
@@ -124,6 +136,42 @@ func (c *CreateCommand) createInput() (*fastly.CreateKafkaInput, error) {
...
+		if c.User.Valid {
+			input.User = fastly.String(c.User.Value)
+		}
+
+		if c.Password.Valid {
+			input.Password = fastly.String(c.Password.Value)
+		}
+	}
+
+	if !c.UseSASL.Value && (c.AuthMethod.Value != "" || c.User.Value != "" || c.Password.Value != "") {
```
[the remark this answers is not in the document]
kellymclaughlin: Good question. It's not necessary. The reason I added this was trying avoid unexpected results and unnecessary debugging. My reasoning was if a user gives the auth method, username, or password they probably meant to enable SASL and by accepting the *invalid* set of options we've possibly added to the confusion. The hope is by enforcing this at creation time we're saving them and possibly ourselves some time. Happy to remove it though if you think it's not worth it.

pkg/logging/kafka/create.go
kellymclaughlin: Done

pkg/logging/kafka/create.go
kellymclaughlin: Done

pkg/logging/kafka/update.go
kellymclaughlin: Done

pkg/logging/kafka/update.go
kellymclaughlin: Done

pkg/logging/kafka/update.go
kellymclaughlin: Done

pkg/logging/kafka/update.go
kellymclaughlin: Done

pkg/logging/kafka/create.go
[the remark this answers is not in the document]
jaredmorrow: It is probably worth giving the user feedback when the options they used make no sense.  My $0.02 it should stay.  The more Go Way to do it would be to move this invalid case as far up in the function as possible.

pkg/logging/kafka/create.go
```diff
@@ -124,6 +136,35 @@ func (c *CreateCommand) createInput() (*fastly.CreateKafkaInput, error) {
...
 
+	if c.ParseLogKeyvals.Valid {
+		input.ParseLogKeyvals = fastly.CBool(c.ParseLogKeyvals.Value)
+	}
+
+	if c.RequestMaxBytes.Valid {
+		input.RequestMaxBytes = fastly.Uint(c.RequestMaxBytes.Value)
+	}
+
+	if c.UseSASL.Valid && c.UseSASL.Value {
```
jaredmorrow, line 147: Move this error conditional towards the top of the function to fail early.

pkg/logging/kafka/create.go
```diff
@@ -124,6 +136,35 @@ func (c *CreateCommand) createInput() (*fastly.CreateKafkaInput, error) {
...
+
+		input.AuthMethod = fastly.String(c.AuthMethod.Value)
+
+		if c.User.Valid {
+			input.User = fastly.String(c.User.Value)
+		}
+
+		if c.Password.Valid {
+			input.Password = fastly.String(c.Password.Value)
+		}
```
jaredmorrow, line 161: Can be outside of the surrounding `if c.UseSASL.Valid && c.UseSASL.Value {` conditional.

pkg/logging/kafka/update.go
```diff
@@ -160,6 +178,50 @@ func (c *UpdateCommand) createInput() (*fastly.UpdateKafkaInput, error) {
...
+		if c.AuthMethod.Value != "" {
+			fmt.Printf("AuthMethod: %s\n", c.AuthMethod.Value)
+		}
+		if c.User.Value != "" {
+			fmt.Printf("AuthMethod: %s\n", c.User.Value)
+		}
+		if c.Password.Value != "" {
+			fmt.Printf("AuthMethod: %s\n", c.Password.Value)
+		}
+		return nil, fmt.Errorf("the --auth-method, --username, and --password options are only valid when the --use-sasl flag is specified")
```
jaredmorrow, line 222: Nit: Same comment as in the kafka/create file.  If possible the successful case should fall through the function, so the failures should be at the top of the function.

jaredmorrow (changes requested):
Some small nits related to https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/logging/kafka/create.go
kellymclaughlin: Done

pkg/logging/kafka/create.go
kellymclaughlin: Done

pkg/logging/kafka/update.go
kellymclaughlin: Done

pkg/logging/kafka/create.go
```diff
@@ -61,19 +68,32 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
 	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint. Can be either 2 (default) or 1").Action(c.FormatVersion.Set).UintVar(&c.FormatVersion.Value)
 	c.CmdClause.Flag("placement", "Where in the generated VCL the logging call should be placed, overriding any format_version default. Can be none or waf_debug").Action(c.Placement.Set).StringVar(&c.Placement.Value)
 	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
+	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).BoolVar(&c.ParseLogKeyvals.Value)
```
phamann, line 71:
```suggestion
	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format").Action(c.ParseLogKeyvals.Set).BoolVar(&c.ParseLogKeyvals.Value)
```

pkg/logging/kafka/create.go
```diff
@@ -61,19 +68,32 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
...
+	c.CmdClause.Flag("max-batch-size", "The maximum size of the log batch in bytes").Action(c.RequestMaxBytes.Set).UintVar(&c.RequestMaxBytes.Value)
+	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified.").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
```
phamann, line 73:
```suggestion
	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
```

pkg/logging/kafka/create.go
```diff
@@ -61,19 +68,32 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
...
+	c.CmdClause.Flag("auth-method", "SASL authentication method. Valid values are: plain, scram-sha-256, scram-sha-512").Action(c.AuthMethod.Set).HintOptions("plain", "scram-sha-256", "scram-sha-512").EnumVar(&c.AuthMethod.Value, "plain", "scram-sha-256", "scram-sha-512")
+	c.CmdClause.Flag("username", "SASL authentication username. Required if --auth-method is specified.").Action(c.User.Set).StringVar(&c.User.Value)
```
phamann, line 75:
```suggestion
	c.CmdClause.Flag("username", "SASL authentication username. Required if --auth-method is specified").Action(c.User.Set).StringVar(&c.User.Value)
```

pkg/logging/kafka/create.go
```diff
@@ -61,19 +68,32 @@ func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCom
...
+	c.CmdClause.Flag("password", "SASL authentication password. Required if --auth-method is specified.").Action(c.Password.Set).StringVar(&c.Password.Value)
```
phamann, line 76:
```suggestion
	c.CmdClause.Flag("password", "SASL authentication password. Required if --auth-method is specified").Action(c.Password.Set).StringVar(&c.Password.Value)
```

pkg/logging/kafka/update.go
```diff
@@ -64,6 +71,12 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
 	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint. Can be either 2 (default) or 1").Action(c.FormatVersion.Set).UintVar(&c.FormatVersion.Value)
 	c.CmdClause.Flag("placement", "Where in the generated VCL the logging call should be placed, overriding any format_version default. Can be none or waf_debug").Action(c.Placement.Set).StringVar(&c.Placement.Value)
 	c.CmdClause.Flag("response-condition", "The name of an existing condition in the configured endpoint, or leave blank to always execute").Action(c.ResponseCondition.Set).StringVar(&c.ResponseCondition.Value)
+	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format.").Action(c.ParseLogKeyvals.Set).NegatableBoolVar(&c.ParseLogKeyvals.Value)
```
phamann, line 74:
```suggestion
	c.CmdClause.Flag("parse-log-keyvals", "Parse key-value pairs within the log format").Action(c.ParseLogKeyvals.Set).NegatableBoolVar(&c.ParseLogKeyvals.Value)
```

pkg/logging/kafka/update.go
```diff
@@ -64,6 +71,12 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
...
+	c.CmdClause.Flag("max-batch-size", "The maximum size of the log batch in bytes").Action(c.RequestMaxBytes.Set).UintVar(&c.RequestMaxBytes.Value)
+	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified.").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
```
phamann, line 76:
```suggestion
	c.CmdClause.Flag("use-sasl", "Enable SASL authentication. Requires --auth-method, --username, and --password to be specified").Action(c.UseSASL.Set).BoolVar(&c.UseSASL.Value)
```

pkg/logging/kafka/update.go
```diff
@@ -64,6 +71,12 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
...
+	c.CmdClause.Flag("auth-method", "SASL authentication method. Valid values are: plain, scram-sha-256, scram-sha-512").Action(c.AuthMethod.Set).HintOptions("plain", "scram-sha-256", "scram-sha-512").EnumVar(&c.AuthMethod.Value, "plain", "scram-sha-256", "scram-sha-512")
+	c.CmdClause.Flag("username", "SASL authentication username. Required if --auth-method is specified.").Action(c.User.Set).StringVar(&c.User.Value)
```
phamann, line 78:
```suggestion
	c.CmdClause.Flag("username", "SASL authentication username. Required if --auth-method is specified").Action(c.User.Set).StringVar(&c.User.Value)
```

pkg/logging/kafka/update.go
```diff
@@ -64,6 +71,12 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
...
+	c.CmdClause.Flag("password", "SASL authentication password. Required if --auth-method is specified.").Action(c.Password.Set).StringVar(&c.Password.Value)
```
phamann, line 79:
```suggestion
	c.CmdClause.Flag("password", "SASL authentication password. Required if --auth-method is specified").Action(c.Password.Set).StringVar(&c.Password.Value)
```

phamann (changes requested):
Thanks @kellymclaughlin. This is looking good now. Just some edits to make the flag descriptions consistent with the rest of the project. You'll need to regenerate the app help for the tests to pass once accepted. Then we're good to go.

phamann (approved):
LGTM 🎉 

Thanks for the contribution @kellymclaughlin and apologies its taken so long to get over the line.

pkg/app/run_test.go
```diff
@@ -2565,6 +2565,18 @@ COMMANDS
                                    The name of an existing condition in the
                                    configured endpoint, or leave blank to always
                                    execute
+        --parse-log-keyvals        Parse key-value pairs within the log format
```
Integralist, line 2568: @kellymclaughlin Is there a reason this doesn't match the `--[no-]parse-log-keyvals` later on in this file?

Integralist (commented):
One very minor comment, also it would be good to have these new fields documented. I've left a comment on the most recent jira ticket here for reference: [here](https://jira.corp.fastly.com/browse/LOG-1074?focusedCommentId=516232&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-516232)

pkg/app/run_test.go
kellymclaughlin: The reason for that difference is that at creation time there is no need to be able to disable this setting (*i.e.* there's no ambiguity at create time) whereas when updating specifying `--no-parse-log-keyvals` explicitly allows one to disable the setting for an endpoint that had previously enabled it.

Integralist: Gotcha. Makes sense. Thanks for clarifying.

go.sum
```diff
@@ -32,6 +32,8 @@ github.com/emirpasic/gods v1.12.0 h1:QAUIPSaCu4G+POclxeqb3F+WPpdKqFGlw36+yOzGlrg
 github.com/emirpasic/gods v1.12.0/go.mod h1:YfzfFFoVP/catgzJb4IKIqXjX78Ha8FMSDh3ymbK86o=
 github.com/fastly/go-fastly v1.16.0 h1:RMHvkzZ52J60+jSPK+6BLodIsx4OBlaoB18XmL7C0+Y=
 github.com/fastly/go-fastly v1.16.0/go.mod h1:jILbTQnU/K/7XHQNzQWd1O7hbXIcp6dKrxfRWqU6xfk=
```
Integralist, line 34: @kellymclaughlin I just pulled your branch and ran `make all` and found that the `go.sum` was updated for me to remove the v1.16.0 dependency. Are you OK to update this PR with that change?

kellymclaughlin: @Integralist Ok, I've pushed that change.

Integralist: Thanks @kellymclaughlin - just to let you know that Patrick and I are still chipping away at the Go-Fastly v2 merge/release (things got complicated 😬) but we expect to have it finished today.

kellymclaughlin: :+1: It looks like something has changed in the last day or so on the Github side with setting environment variables for the CI actions and it's now causing the checks to fail. Let me know if you would like me to try to run that down for this PR.

Integralist: Yeah we had to fix that here: https://github.com/fastly/terraform-provider-fastly/pull/340/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R55

Integralist:
Looks like there is an `add-path` used as well somewhere and that I believe also needs to use the new github environment files? https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

kellymclaughlin:
Third time's a charm, but got the CI checks happy again.

Integralist (approved):

PR #165, opened 2020-11-03
Constrain Version Upgrade Suggestion

Integralist:
**Problem**: when developing locally, `./fastly version` would suggest an upgrade to a pre-release version.
**Solution**: version upgrades should only be for standard semver releases (e.g. `v1.1.1` not `v.1.1.1-rc-1`).
**Notes**: customers should (in theory) only be using an 'official' release and won't expect pre-release.

### DIFF
--- a/pkg/version/root.go
+++ b/pkg/version/root.go
@@ -70,9 +70,9 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 	return nil
 }
 
-// PreRelease determines if the given app version is a pre-release.
+// IsPreRelease determines if the given app version is a pre-release.
 //
 // NOTE: this is indicated by the presence of a hyphen, e.g. v1.0.0-rc.1
-func PreRelease(version string) bool {
-    result strings.Contains(version, "-")
+func IsPreRelease(version string) bool {
+	return strings.Contains(version, "-")
 }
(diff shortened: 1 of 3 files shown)


pkg/version/root.go
```diff
@@ -68,3 +69,16 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
...
+//
+// NOTE: this is indicated by the presence of a hyphen, e.g. v1.0.0-rc.1
+func PreRelease(version string) bool {
+	var result bool
+
+	if i := strings.Index(version, "-"); i >= 0 {
+		result = true
+	}
+
+	return result
```
phamann, line 83: Is there a reason why you chose this over `strings.Contains()`?

```suggestion
    result strings.Contains(version, "-")
```

pkg/version/root.go
```diff
@@ -68,3 +69,16 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 	fmt.Fprintf(out, "Built with %s\n", GoVersion)
 	return nil
 }
+
+// PreRelease determines if the given app version is a pre-release.
+//
+// NOTE: this is indicated by the presence of a hyphen, e.g. v1.0.0-rc.1
+func PreRelease(version string) bool {
```
phamann, line 76: Personal style, but I normally prefix predicate functions like this with `IsFoo` so the conditional reads nicer.

```suggestion
func IsPreRelease(version string) bool {
```

phamann (approved):
👍 LGTM - comments are minor nits and not blockers.

pkg/version/root.go
Integralist: no reason. should have used `Contains` so good shout 👍🏻

PR #169, opened 2020-11-18
Migrate to Go-Fastly 2.0.0

Integralist:
**Problem**: multiple lines of tagged releases for Go-Fastly is making adding new features via our various API clients much harder.
**Solution**: migrate cli to using Go-Fastly 2.0.0


Integralist:
@phamann this is ready for you (see DM also).

pkg/api/interface.go
```diff
@@ -16,6 +16,10 @@ type HTTPClient interface {
 
 // Interface models the methods of the Fastly API client that we use.
 // It exists to allow for easier testing, in combination with Mock.
+//
+// TODO(integralist):
+// there are missing methods such as GetVersion from this list so review in
```
phamann, line 21:
```suggestion
// There are missing methods such as GetVersion from this list so review in
```

pkg/domain/update.go
```diff
@@ -51,24 +53,24 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 		return err
 	}
 
-	c.updateInput.Service = d.ServiceID
-	c.updateInput.Version = d.Version
+	c.updateInput.ServiceID = d.ServiceID
+	c.updateInput.ServiceVersion = d.ServiceVersion
+
+	// set original value, and then afterwards check if we can use the flag value
```
phamann, line 59:
```suggestion
	// Set original value, and then afterwards check if we can use the flag value.
```

pkg/service/update.go
```diff
@@ -18,6 +18,13 @@ type UpdateCommand struct {
 	manifest    manifest.Data
 	getInput    fastly.GetServiceInput
 	updateInput fastly.UpdateServiceInput
+
+	// TODO(integralist):
+	// ensure consistency in capitalization
+	// should be lowercase to avoid ambiguity in common.Command interface
```
phamann, line 24:
```suggestion
	// TODO(integralist):
	// Ensure consistency in capitalization, should be lowercase to avoid 
	// ambiguity in common.Command interface.
```

pkg/service/update.go
```diff
@@ -40,8 +47,16 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 	}
 	c.getInput.ID = serviceID
 
+	// TODO(integralist):
+	// validation such as this should become redundant once Go-Fastly is
```
phamann, line 51:
```suggestion
	// Validation such as this should become redundant once Go-Fastly is
```

pkg/service/update.go
```diff
@@ -40,8 +47,16 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
...
+	// consistently implementing validation (which itself should be redundant
+	// once each backend API is 100% confirmed as validating client inputs).
+	//
+	// as it stands we have multiple clients duplicating logic which should exist
```
phamann, line 55:
```suggestion
	// As it stands we have multiple clients duplicating logic which should exist
```

pkg/serviceversion/update.go
```diff
@@ -25,8 +29,14 @@ func NewUpdateCommand(parent common.Registerer, globals *config.Data) *UpdateCom
 	c.manifest.File.Read(manifest.Filename)
 	c.CmdClause = parent.Command("update", "Update a Fastly service version")
 	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
-	c.CmdClause.Flag("version", "Number of version you wish to update").Required().IntVar(&c.Input.Version)
-	c.CmdClause.Flag("comment", "Human-readable comment").Required().StringVar(&c.Input.Comment)
+	c.CmdClause.Flag("version", "Number of version you wish to update").Required().IntVar(&c.updateInput.ServiceVersion)
+
+	// TODO(integralist):
+	// make 'comment' field mandatory once we roll out a new release of Go-Fastly
```
phamann, line 35:
```suggestion
	// Make 'comment' field mandatory once we roll out a new release of Go-Fastly
```

pkg/serviceversion/update.go
```diff
@@ -36,13 +46,32 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
...
+	if !c.comment.WasSet {
+		return fmt.Errorf("error parsing arguments: required flag --comment not provided")
+	}
+
+	s, err := c.Globals.Client.GetService(&c.getInput)
+	if err != nil {
+		return err
+	}
+
+	// set original value, and then afterwards check if we can use the flag value
```
phamann, line 61:
```suggestion
	// Set original value, and then afterwards check if we can use the flag value.
```

pkg/serviceversion/update.go
```diff
@@ -36,13 +46,32 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
...
+	c.updateInput.ServiceID = serviceID
+	c.updateInput.Comment = &s.Comment
+
+	// update field value as required
```
phamann, line 65:
```suggestion
	// Update field value as required.
```

phamann (approved):
👍 LGTM - Thank you so much for this very tedious, but very important upgrade!

My only comments are minor and non-blocking:
- A personal preference, but naming of `optional.WasSet` seems weird to me and makes the type less portable for use outside of flag actions. `IsValid` or `HasValue` may have been better choices.
- I formatted lots of comments per: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences. Most of them were inline and thus not an issue, but I prefer the consistency.
- Would be good to change the `comment` fields in go-fastly asap so they don't become breaking changes.

PR #170, opened 2020-11-19
v0.19.0

Integralist:


CHANGELOG.md
```diff
@@ -205,6 +212,7 @@
 
 **Enhancements:**
 
+- Bump default Rust template to v0.3.0 [\#32](https://github.com/fastly/cli/pull/32)
```
phamann, line 215: Is this move correct or we're non-determinism of the changelog tool?

phamann (approved):
👍 LGTM - other than minor changelog comment

CHANGELOG.md
Integralist: yeah the changelog definitely isn't showing any changes from what I've committed so I think someone made the mistake of manually changing that line so it wasn't under 'enhancement' (which I can understand, but due to the automated nature of the tool leaves confusion when manually changing things).

PR #172, opened 2020-11-21
rust: support cargo workspaces

### LINKED ISSUE #171: Build failure with Cargo workspaces
**Version**

```
$ fastly version
Fastly CLI version 0.19.0 (98cb4ff)
Built with go version unknown
```

**What happened**

Build failure with a project using [Cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html).

A minimum project to reproduce this problem: https://github.com/fastly/ecp-workspace-example

You can see the following error messages by `make build`.

```
(cd app && fastly compute build)
✓ Initializing...
✗ Building package using rust toolchain...

ERROR: copying wasm binary: cannot read source file: /Users/goro-fuji/ghq/github.com/fastly/ecp-workspace-example/app/target/wasm32-wasi/release/fastly-template-rust-default.wasm.

```

I'm using Cargo workspaces to test the `core` logic in my local machine, but it seems that `fastly compute build` cannot recognize the workspace project. 

Good news: `cargo test -p core` works! 🎉

gfx:
Resolve #171, as @pchickey suggested in https://github.com/fastly/cli/issues/171#issuecomment-731377807

Confirmed I can build and deploy https://github.com/fastly/ecp-workspace-example

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -54,12 +54,13 @@ func (m *CargoManifest) Read(filename string) error {
 // CargoMetadata models information about the workspace members and resolved
 // dependencies of the current package via `cargo metadata` command output.
 type CargoMetadata struct {
-	Package []CargoPackage `json:"packages"`
+	Package         []CargoPackage `json:"packages"`
+	TargetDirectory string         `json:"target_directory"`
 }
 
 // Read the contents of the Cargo.lock file from filename.
 func (m *CargoMetadata) Read() error {
-	cmd := exec.Command("cargo", "metadata", "--format-version", "1")
+	cmd := exec.Command("cargo", "metadata", "--quiet", "--format-version", "1")
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/compute/rust.go
```diff
@@ -54,12 +54,13 @@ func (m *CargoManifest) Read(filename string) error {
...
 type CargoMetadata struct {
-	Package []CargoPackage `json:"packages"`
+	Package         []CargoPackage `json:"packages"`
+	TargetDirectory string         `json:"target_directory"`
 }
 
 // Read the contents of the Cargo.lock file from filename.
 func (m *CargoMetadata) Read() error {
-	cmd := exec.Command("cargo", "metadata", "--format-version", "1")
+	cmd := exec.Command("cargo", "metadata", "--quiet", "--format-version", "1")
```
phamann, line 63: What is the reason for adding the `--quiet` flag here?

phamann (approved):
👍 LGTM - other than question regarding addition of quiet flag.

pkg/compute/rust.go
gfx: It suppresses some warnings, esp. for workspaces, like:

```
$ cargo metadata --format-version 1 > /dev/null
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/goro-fuji/ghq/github.com/gfx/ecp-hello/app/Cargo.toml
workspace: /Users/goro-fuji/ghq/github.com/gfx/ecp-ecp-hello/Cargo.toml
```

It doesn't affect the fastly cli but I feel it's better to suppress.

PR #177, opened 2020-12-15
Add support for Kinesis logging endpoint

kellymclaughlin:
This requires moving to version 2.1.0 of `go-fastly`. If there's other work that needs to be done before that can happen that's fine, but I wanted to go ahead and get the PR out there even if it can't be merged right away.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -30,8 +30,8 @@ github.com/dustinkirkland/golang-petname v0.0.0-20191129215211-8e5a1ed0cff0 h1:9
 github.com/dustinkirkland/golang-petname v0.0.0-20191129215211-8e5a1ed0cff0/go.mod h1:V+Qd57rJe8gd4eiGzZyg4h54VLHmYVVw54iMnlAMrF8=
 github.com/emirpasic/gods v1.12.0 h1:QAUIPSaCu4G+POclxeqb3F+WPpdKqFGlw36+yOzGlrg=
 github.com/emirpasic/gods v1.12.0/go.mod h1:YfzfFFoVP/catgzJb4IKIqXjX78Ha8FMSDh3ymbK86o=
-github.com/fastly/go-fastly/v2 v2.0.0 h1:dCPOEbTXy8ic5CGj6YibPNvRYG01y15H7kHTlC4d57k=
-github.com/fastly/go-fastly/v2 v2.0.0/go.mod h1:+gom+YR+9Q5I4biSk/ZjHQGWXxqpRxC3YDVYQcRpZwQ=
+github.com/fastly/go-fastly/v2 v2.1.0 h1:VQWCFiUZyCcknyxyhiQwU0MjLYzeK4g1mLo/US2r/P0=
+github.com/fastly/go-fastly/v2 v2.1.0/go.mod h1:+gom+YR+9Q5I4biSk/ZjHQGWXxqpRxC3YDVYQcRpZwQ=
 github.com/fastly/kingpin v2.1.12-0.20191105091915-95d230a53780+incompatible h1:FhrXlfhgGCS+uc6YwyiFUt04alnjpoX7vgDKJxS6Qbk=
 github.com/fastly/kingpin v2.1.12-0.20191105091915-95d230a53780+incompatible/go.mod h1:U8UynVoU1SQaqD2I4ZqgYd5lx3A1ipQYn4aSt2Y5h6c=
 github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -859,12 +859,11 @@ COMMANDS
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the S3 logging object
 
-  logging kinesis create --name=NAME --service-id=SERVICE-ID --version=VERSION --stream-name=STREAM-NAME --access-key=ACCESS-KEY --secret-key=SECRET-KEY --region=REGION [<flags>]
+  logging kinesis create --name=NAME --version=VERSION --stream-name=STREAM-NAME --access-key=ACCESS-KEY --secret-key=SECRET-KEY --region=REGION [<flags>]
     Create an Amazon Kinesis logging endpoint on a Fastly service version
 
     -n, --name=NAME                The name of the Kinesis logging object. Used
                                    as a primary key for API access
-    -s, --service-id=SERVICE-ID    Service ID
         --version=VERSION          Number of service version
         --stream-name=STREAM-NAME  The Amazon Kinesis stream to send logs to
         --access-key=ACCESS-KEY    The access key associated with the target
@@ -873,6 +872,7 @@ COMMANDS
                                    Amazon Kinesis stream
         --region=REGION            The AWS region where the Kinesis stream
                                    exists
+    -s, --service-id=SERVICE-ID    Service ID
         --format=FORMAT            Apache style log formatting
         --format-version=FORMAT-VERSION
                                    The version of the custom logging format used
@@ -901,12 +901,12 @@ COMMANDS
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the Kinesis logging object
 
-  logging kinesis update --service-id=SERVICE-ID --version=VERSION --name=NAME [<flags>]
+  logging kinesis update --version=VERSION --name=NAME [<flags>]
     Update a Kinesis logging endpoint on a Fastly service version
 
-    -s, --service-id=SERVICE-ID    Service ID
         --version=VERSION          Number of service version
     -n, --name=NAME                The name of the Kinesis logging object
+    -s, --service-id=SERVICE-ID    Service ID
         --new-name=NEW-NAME        New name of the Kinesis logging object
         --stream-name=STREAM-NAME  Your Kinesis stream name
         --access-key=ACCESS-KEY    Your Kinesis account access key
@@ -927,12 +927,12 @@ COMMANDS
                                    format_version default. Can be none or
                                    waf_debug
 
-  logging kinesis delete --service-id=SERVICE-ID --version=VERSION --name=NAME
+  logging kinesis delete --version=VERSION --name=NAME [<flags>]
     Delete a Kinesis logging endpoint on a Fastly service version
 
-    -s, --service-id=SERVICE-ID  Service ID
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the Kinesis logging object
+    -s, --service-id=SERVICE-ID  Service ID
 
   logging syslog create --name=NAME --version=VERSION --address=ADDRESS [<flags>]
     Create a Syslog logging endpoint on a Fastly service version
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -859,6 +859,81 @@ COMMANDS
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the S3 logging object
 
+  logging kinesis create --name=NAME --service-id=SERVICE-ID --version=VERSION --stream-name=STREAM-NAME --access-key=ACCESS-KEY --secret-key=SECRET-KEY --region=REGION [<flags>]
+    Create an Amazon Kinesis logging endpoint on a Fastly service version
+
+    -n, --name=NAME                The name of the Kinesis logging object. Used
+                                   as a primary key for API access
+    -s, --service-id=SERVICE-ID    Service ID
+        --version=VERSION          Number of service version
+        --stream-name=STREAM-NAME  The Amazon Kinesis stream to send logs to
+        --access-key=ACCESS-KEY    The access key associated with the target
+                                   Amazon Kinesis stream
+        --secret-key=SECRET-KEY    The secret key associated with the target
+                                   Amazon Kinesis stream
+        --region=REGION            The AWS region where the Kinesis stream
+                                   exists
+        --format=FORMAT            Apache style log formatting
+        --format-version=FORMAT-VERSION
+                                   The version of the custom logging format used
+                                   for the configured endpoint. Can be either 2
+                                   (default) or 1
+        --response-condition=RESPONSE-CONDITION
+                                   The name of an existing condition in the
+                                   configured endpoint, or leave blank to always
+                                   execute
+        --placement=PLACEMENT      Where in the generated VCL the logging call
+                                   should be placed, overriding any
+                                   format_version default. Can be none or
+                                   waf_debug
+
+  logging kinesis list --version=VERSION [<flags>]
+    List Kinesis endpoints on a Fastly service version
+
+    -s, --service-id=SERVICE-ID  Service ID
+        --version=VERSION        Number of service version
+
+  logging kinesis describe --version=VERSION --name=NAME [<flags>]
+    Show detailed information about a Kinesis logging endpoint on a Fastly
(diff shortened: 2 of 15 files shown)


pkg/logging/kinesis/describe.go
```diff
@@ -0,0 +1,59 @@
...
+	fmt.Fprintf(out, "Version: %d\n", kinesis.ServiceVersion)
+	fmt.Fprintf(out, "Name: %s\n", kinesis.Name)
+	fmt.Fprintf(out, "Stream name: %s\n", kinesis.StreamName)
+	fmt.Fprintf(out, "Region: %s\n", kinesis.Region)
+	fmt.Fprintf(out, "Access key: %s\n", kinesis.AccessKey)
+	fmt.Fprintf(out, "Secret key: %s\n", kinesis.SecretKey)
+	fmt.Fprintf(out, "Format: %s\n", kinesis.Format)
+	fmt.Fprintf(out, "Format version: %d\n", kinesis.FormatVersion)
+	fmt.Fprintf(out, "Response condition: %s\n", kinesis.ResponseCondition)
+	fmt.Fprintf(out, "Placement: %s\n", kinesis.Placement)
```
Integralist, line 56: I think we should create a new `/pkg/text/lines.go` to make it easier to render output for newline information (like logging endpoints) and then we'd also be consistent with other packages in the CLI that use the text package.

Something generic like this would suffice: https://play.golang.org/p/csgKDzLZPeu

Which means you could do something like: 

```go
import "github.com/fastly/cli/pkg/text"

text.PrintLines(out, text.Lines{
    "Service ID": kinesis.ServiceID,
    "Version": kinesis.ServiceVersion,
    "Name": kinesis.Name,
    ...
})
```

Integralist: @phamann what do you think about this...

If we combined this generic implementation (see my comment above) with `textio.NewPrefixWriter` then we could probably end up reducing quite a few of the existing `Print<T>` functions (e.g. `PrintBackend`, `PrintDictionaryItemKV` etc).

You could even take it a step further and add in the ability to pass in an optional callback function for those edge cases such as `PrintDictionaryItem` where they only print lines that have values (although to be fair that's probably logic you'd shift back up into the consuming package). But I'm just 'brain dumping' ideas at the moment.

pkg/logging/kinesis/root.go
```diff
@@ -0,0 +1,28 @@
...
+type RootCommand struct {
+	common.Base
+	// no flags
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent common.Registerer, globals *config.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("kinesis", "Manipulate Fastly service version Kinesis logging endpoints")
```
Integralist, line 21: I'm not a fan of the wording used in the CLI currently as it doesn't read like a correct sentence.

```suggestion
	c.CmdClause = parent.Command("kinesis", "Manipulate a Kinesis logging endpoint for a specific Fastly service version")
```

This could be something I fix more broadly in a separate PR though.

Oh, actually -- is this command even reachable (after looking at `Exec`)?

pkg/logging/kinesis/list.go
```diff
@@ -0,0 +1,75 @@
...
+		tw := text.NewTable(out)
+		tw.AddHeader("SERVICE", "VERSION", "NAME")
+		for _, kinesis := range kineses {
+			tw.AddLine(kinesis.ServiceID, kinesis.ServiceVersion, kinesis.Name)
+		}
+		tw.Print()
+		return nil
+	}
+
+	fmt.Fprintf(out, "Service ID: %s\n", c.Input.ServiceID)
```
Integralist, line 56: Similar to my earlier comment I think a new generic text implementation could be useful here as well.

Integralist (commented):
Only a few minor comments. Otherwise looking good 👍🏻

pkg/app/run_test.go
```diff
@@ -859,6 +859,81 @@ COMMANDS
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the S3 logging object
 
+  logging kinesis create --name=NAME --version=VERSION --stream-name=STREAM-NAME --access-key=ACCESS-KEY --secret-key=SECRET-KEY --region=REGION [<flags>]
+    Create an Amazon Kinesis logging endpoint on a Fastly service version
+
+    -n, --name=NAME                The name of the Kinesis logging object. Used
+                                   as a primary key for API access
+    -s, --service-id=SERVICE-ID    Service ID
+        --version=VERSION          Number of service version
```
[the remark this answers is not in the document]
kellymclaughlin: :+1: That makes sense to me. I'll defer to you on the timing of when to actually do it.

pkg/logging/kinesis/create.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/fastly/go-fastly/v2/fastly"
+)
+
+// CreateCommand calls the Fastly API to create Amazon Kinesis logging endpoints.
+type CreateCommand struct {
+	common.Base
+	manifest manifest.Data
+
+	// required
+	EndpointName string // Can't shaddow common.Base method Name().
```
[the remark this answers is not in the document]
kellymclaughlin: Nice catch! This is a copypasta typo that looks to be present in nearly all the other file under the logging subdirectory. I'll have to subsequently file a PR to fix all those.

pkg/logging/kinesis/create.go
```diff
@@ -0,0 +1,108 @@
...
+
+	c.CmdClause.Flag("name", "The name of the Kinesis logging object. Used as a primary key for API access").Short('n').Required().StringVar(&c.EndpointName)
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
+	c.CmdClause.Flag("version", "Number of service version").Required().IntVar(&c.Version)
+
+	c.CmdClause.Flag("stream-name", "The Amazon Kinesis stream to send logs to").Required().StringVar(&c.StreamName)
+	c.CmdClause.Flag("access-key", "The access key associated with the target Amazon Kinesis stream").Required().StringVar(&c.AccessKey)
+	c.CmdClause.Flag("secret-key", "The secret key associated with the target Amazon Kinesis stream").Required().StringVar(&c.SecretKey)
+
+	c.CmdClause.Flag("region", "The AWS region where the Kinesis stream exists").Required().StringVar(&c.Region)
```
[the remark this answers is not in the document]
kellymclaughlin: Yeah, it is required. The configuration validation will fail if it is not present. I see what you mean in the case of `go-fastly`. I followed the existing patterns of other logging endpoints when I did that work and they all seem to only have `ServiceId` and `ServiceVersion` as required. My guess is that the enforcement of required fields was deferred to the config validation, but maybe we want to revisit that.

pkg/logging/kinesis/describe.go
kellymclaughlin: :+1: to this idea from me. I'd say this is probably out of scope for this PR, but I think it would be a good change to implement broadly.

pkg/logging/kinesis/create.go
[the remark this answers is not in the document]
Integralist: OK cool I'll ignore this for now and expect a follow-up PR from you with all files corrected 👍🏻

pkg/logging/kinesis/create.go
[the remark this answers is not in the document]
Integralist: **Question**:
> The configuration validation will fail if it is not present.

I presume 'config validation' refers to the upstream API's own validation, would that be correct?

pkg/logging/kinesis/describe.go
phamann: Agree on both accounts:

1. Very useful abstraction and would reduce repetition across the codebase.
2. Out of scope of this PR.

pkg/logging/kinesis/root.go
phamann: > Oh, actually -- is this command even reachable (after looking at Exec)?

All command groups have unreachable roots in the tree. This is the root for this group of commands, the text is only used in `help` output when, for instance, `fastly logging --help` is ran.

pkg/logging/kinesis/create.go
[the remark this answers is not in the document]
Integralist: I see Go-Fastly currently doesn't define validation for the 'region' field, but if we're saying the API itself is expecting a region field otherwise it'll return an error, then we should ensure a validation check for the 'region' field is added into Go-Fastly (I have an existing PR open I can add this to).

But for the time being, this is fine to be marked as `.Required()` considering Go-Fastly doesn't _currently_ have logic to catch the missing field.

> Note: I can't actually test whether the API expects 'region' to be provided because it seems Kinesis logging isn't available on my personal Fastly account 😬 I did login as an admin to see if it's something I could enable as a feature flag but nothing obvious was coming up in my search.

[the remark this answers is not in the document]
Integralist: Also, just FYI (most of which will be obvious to you): my perspective on this is that we have three tools at play here:

```
CLI > Go-Fastly > API
```

We should be able to trust the API to always have the right validation in place, but (as you know) by having validation _outside_ of the API it means we can prevent slowing down a client trying to carry out operations that otherwise we know are just going to fail by the time they reach the API layer, so short-circuiting those requests is better for the client as far as the 'feedback loop' is concerned but also good for the API because it doesn't have to handle the request in the first place.

Considering Go-Fastly and CLI sit at effectively the same level (i.e. they exist _on_ the client) we need to decide which 'tool' is responsible for handling validation of obvious issues with a client request, and for me that should be Go-Fastly as it's the thing that's ultimately fronting the API.

So my hope is that with the work @doramatadora  has been doing on the new [OpenAPI specification](https://github.com/fastly/api-documentation/pull/433) it will mean we can move to a place where we more concretely know what is a required field and have Go-Fastly handle validation and thus reduce duplicating logic in the CLI (although admittedly at this point in time the OpenAPI spec [for Kinesis logging](https://github.com/fastly/api-documentation/pull/433/files#diff-56bc7bba0f7cc7c57afed506447d0e9d30eff59cc148624f3177437f225e0d3a) still doesn't indicate that 'region' is a required field nor does it show it as part of its example code so that would likely need updating).

pkg/logging/kinesis/describe.go
Integralist: 👍🏻 👌🏻

pkg/logging/kinesis/create.go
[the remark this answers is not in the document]
kellymclaughlin: > I presume 'config validation' refers to the upstream API's own validation, would that be correct?

That's correct.

> Note: I can't actually test whether the API expects 'region' to be provided because it seems Kinesis logging isn't available on my personal Fastly account grimacing I did login as an admin to see if it's something I could enable as a feature flag but nothing obvious was coming up in my search.

If you send me the service id in slack I can enable this for you.

I'll also follow-up on the docs side of things and open a change request if needed to have it indicate `region` as required.

pkg/logging/kinesis/create.go
```diff
@@ -0,0 +1,109 @@
...
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent common.Registerer, globals *config.Data) *CreateCommand {
+	var c CreateCommand
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("create", "Create an Amazon Kinesis logging endpoint on a Fastly service version").Alias("add")
+
+	// required
+	c.CmdClause.Flag("name", "The name of the Kinesis logging object. Used as a primary key for API access").Short('n').Required().StringVar(&c.EndpointName)
+	c.CmdClause.Flag("service-id", "Service ID").Required().Short('s').StringVar(&c.manifest.Flag.ServiceID)
```
kellymclaughlin, line 43: I noticed that `service-id` is not marked as required in many or maybe all of the logging endpoint code. I've set it for Kinesis in this PR, but I'll address the inconsistency in the rest of the logging endpoints in another PR.

pkg/logging/kinesis/create.go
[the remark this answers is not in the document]
kellymclaughlin: [Here](https://github.com/fastly/cli/pull/182) is the PR to fix this more broadly along with some other inconsistencies.

Integralist:
@kellymclaughlin just checking, was this PR ready for another review?

kellymclaughlin:
@Integralist Yes, this should be good for another look. Thanks!

go.sum
```diff
@@ -32,6 +32,8 @@ github.com/emirpasic/gods v1.12.0 h1:QAUIPSaCu4G+POclxeqb3F+WPpdKqFGlw36+yOzGlrg
 github.com/emirpasic/gods v1.12.0/go.mod h1:YfzfFFoVP/catgzJb4IKIqXjX78Ha8FMSDh3ymbK86o=
 github.com/fastly/go-fastly/v2 v2.0.0 h1:dCPOEbTXy8ic5CGj6YibPNvRYG01y15H7kHTlC4d57k=
 github.com/fastly/go-fastly/v2 v2.0.0/go.mod h1:+gom+YR+9Q5I4biSk/ZjHQGWXxqpRxC3YDVYQcRpZwQ=
```
Integralist, line 34: can we run a `go mod tidy` to see if we can get the v2.0.0 go-fastly dependency cleared from this file.

pkg/logging/kinesis/delete.go
```diff
@@ -0,0 +1,47 @@
...
+	Input    fastly.DeleteKinesisInput
+}
+
+// NewDeleteCommand returns a usable command registered under the parent.
+func NewDeleteCommand(parent common.Registerer, globals *config.Data) *DeleteCommand {
+	var c DeleteCommand
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("delete", "Delete a Kinesis logging endpoint on a Fastly service version").Alias("remove")
+	c.CmdClause.Flag("service-id", "Service ID").Required().Short('s').StringVar(&c.manifest.Flag.ServiceID)
```
Integralist, line 27: @kellymclaughlin so interestingly I'm not sure if service-id needs to be 'required' in the sense that later on in `Exec` we have...

```go
serviceID, source := c.manifest.ServiceID()
if source == manifest.SourceUndefined {
	return errors.ErrNoServiceID
}
```

If we inspect the `ServiceID()` method it's checking first the flag and _then_ the manifest file for a service id before erroring if both are missing...

https://github.com/fastly/cli/blob/c0f337321376bb029c1ffee8b8c1181289dff695/pkg/compute/manifest/manifest.go#L52-L63

kellymclaughlin: Oh interesting. Ok, I'll remove the `Required` and move the `service-id` option to the optional section then.

kellymclaughlin: Done

go.sum
kellymclaughlin: Done

Integralist (approved):
@kellymclaughlin this LGTM but before merging could you first rebase master.

Depending on when you see this notification you might need to wait until I've gotten someone to approve [this PR](https://github.com/fastly/cli/pull/185) (which will cut a new 'patch' CLI release `v0.21.2`). 

Once that's done and you've rebased, then your PR can be cut as a 'minor' release `v0.22.0` (along with your [Splunk PR](https://github.com/fastly/cli/pull/183#pullrequestreview-562765706)).

Thanks!

Issue #168, opened 2020-11-17
CLI shouldn't recommend Rust crate prerelease versions


acfoltzer:
**Version**

Please paste the output of `fastly version` here.

```
❯ fastly version
Fastly CLI version 0.18.1 (f548f88)
Built with go version go1.14.10 linux/amd64
```

**What happened**

`fastly compute build` checks that the version of `fastly-sys` used by the local project is compatible with the latest release from `crates.io`. Prerelease versions such as `0.6.0-beta1` mess with this check. I would propose two things:

1. If the local project is using a prerelease version, the CLI should not fail due to an out of date version. We might want to warn if there is a non-prerelease version available that is newer than the prerelease version, but in general I would prefer that we assume that folks know what they're doing if they're specifying a prerelease manually (it can only be used if you use an `=` constraint explicitly in `Cargo.toml`, like `fastly = "=0.6.0-beta1"`).

    Currently, if I specify `fastly = "=0.6.0-beta1"`, I get the following output, which is incorrect both in that it's recommending I upgrade to a version I'm already using, and recommending that I use a caret dependency which does not work with prerelease versions:

```
ERROR: fastly-sys crate not found.

To fix this error, edit Cargo.toml with:

         fastly = "^0.6.0-beta1"
```

2. If there is a prerelease version available on `crates.io` that is newer than the local project's version, it should never be recommended as an upgrade. For example, if a project doesn't have a `fastly-sys` dependency at all (e.g., if it has a `fastly = "=0.2.0-alpha4"` dependency), we currently recommend `"^0.6.0-beta1"` rather than the most recent non-prerelease version `^0.5.0`.

**Current workaround**

Running `fastly compute build --force` sidesteps the issue for now, but is probably too blunt of a tool to use whenever we want to deal with a prerelease crate.

PR #178, opened 2020-12-17
Handle prerelease versions of the Rust crate more gracefully

phamann:
Fixes: #168

### TL;DR
Updates the `rust.Verify()` verification logic of Rust packages to ensure we handle prerelease versions of the `fastly` crate more gracefully by:

- Exit early out of verification if a prerelease is found to ensure we don't display any confusing messages to "upgrade" an older version.
- Don't suggest prereleases as an optional upgrade if a newer one exists.

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -231,16 +231,23 @@ func (r Rust) Verify(out io.Writer) error {
 		return newCargoUpdateRemediationErr(err, latestFastly.String())
 	}
 
-	// If fastly-sys version doesn't meet our constraint of being within the
-	// minor range, error with dual remediation steps.
-	if ok := fastlySysConstraint.Check(fastlySysVersion); !ok {
-		return newCargoUpdateRemediationErr(fmt.Errorf("fastly crate not up-to-date"), latestFastly.String())
-	}
-
 	fastlyVersion, err := getCrateVersionFromMetadata(metadata, "fastly")
 	// If fastly crate not found, error with dual remediation steps.
 	if err != nil {
 		return newCargoUpdateRemediationErr(err, latestFastly.String())
+	}
+
+	// If fastly crate version is a prerelease, exit early. We assume that the
+	// user knows what they are doing and avoids any confusing messaging to
+	// "upgrade" to an older version.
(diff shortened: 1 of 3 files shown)


pkg/compute/rust.go
```diff
@@ -231,18 +231,25 @@ func (r Rust) Verify(out io.Writer) error {
 		return newCargoUpdateRemediationErr(err, latestFastly.String())
 	}
 
-	// If fastly-sys version doesn't meet our constraint of being within the
```
Integralist, line 234: Not part of your changeset but we should update the code comment further up in this file which is stale...

```
// 2) Check that the `1.43.0` toolchain is installed
```

to

```
// 2) Check that the desired toolchain version is installed
```

pkg/compute/rust.go
```diff
@@ -231,18 +231,25 @@ func (r Rust) Verify(out io.Writer) error {
...
+	// user knows what they are doing and avoids any confusing messaging to
+	// "upgrade" to an older version.
+	if fastlyVersion.Prerelease() != "" {
+		return nil
+	}
+
+	// If fastly-sys version doesn't meet our constraint of being within the
+	// minor range, error with dual remediation steps.
+	if ok := fastlySysConstraint.Check(fastlySysVersion); !ok {
+		return newCargoUpdateRemediationErr(fmt.Errorf("fastly crate not up-to-date"), latestFastly.String())
```
Integralist, line 250: Is `not up-to-date` an accurate description? (based on my interpretation of the code comment above, it sounds like we're _constraining_ to a minor version) 

So should it be something like:

```suggestion
		return newCargoUpdateRemediationErr(fmt.Errorf("fastly-sys crate invalid"), latestFastly.String())
```

...although that's a bit vague 😬

phamann: Happy to change, however this wasn't an addition in this PR, I just moved the block for control flow. Also the original message intentionally references `fastly` and not `fastly-sys` as `fastly-sys` is transitive and thus would be confusing to the user as they may not know it even exists, so we opted to tell them to upgrade their `fastly` version which in turn would upgrade `fastly-sys`. Hope that makes sense.

Integralist (approved):

PR #182, opened 2020-12-21
Consistency fixups across all logging endpoints

kellymclaughlin:
* Address some spelling and grammer issues in comments

### DIFF
--- a/pkg/logging/splunk/update.go
+++ b/pkg/logging/splunk/update.go
@@ -11,13 +11,13 @@ import (
 	"github.com/fastly/go-fastly/v2/fastly"
 )
 
-// UpdateCommand calls the Fastly API to update Splunk logging endpoints.
+// UpdateCommand calls the Fastly API to update a Splunk logging endpoint.
 type UpdateCommand struct {
 	common.Base
 	manifest manifest.Data
 
 	// required
-	EndpointName string // Can't shaddow common.Base method Name().
+	EndpointName string // Can't shadow common.Base method Name().
 	Version      int
 
 	// optional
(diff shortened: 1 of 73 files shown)


Integralist:
👋🏻 @kellymclaughlin as discussed in your other PR it looks like ServiceID shouldn't be a required flag because the internal implementation is first checking the flag, then for the value within a configuration file before erroring if no value exists in either of those places:

https://github.com/fastly/cli/blob/c0f337321376bb029c1ffee8b8c1181289dff695/pkg/compute/manifest/manifest.go#L52-L63

But your other tweaks might still be fine to be reviewed. I'll let you revert the ServiceID changes first before I review what's left.

Thanks!

kellymclaughlin:
@Integralist I made service-id optional again and so this should be ready for review now.

pkg/logging/splunk/update.go
```diff
@@ -17,7 +17,7 @@ type UpdateCommand struct {
 	manifest manifest.Data
 
 	// required
-	EndpointName string // Can't shaddow common.Base method Name().
```
Integralist, line 20: @kellymclaughlin looks like you missed out one instance:

```
// UpdateCommand calls the Fastly API to update Splunk logging endpoints.
```

Did you leave this in just as a way to ensure I reviewed every single file? 😉 well... I did check _every single file_ lol 😆

Integralist (commented):
👋🏻 @kellymclaughlin just one small addition please and then this is good to go 🙂

pkg/logging/splunk/update.go
kellymclaughlin: Darn it. How did I miss that? :facepalm:

kellymclaughlin: Fixed

Integralist (approved):
LGTM. Thanks Kelly 🎉

PR #184, opened 2021-01-06
Switch from third-party dependency to our own mirror

Integralist:
**Problem**: the third-party CLI parser dependency (Kingpin) is no longer maintained, nor are we using a stable branch.
**Solution**: use a mirror of the repo that is within the fastly organisation to mitigate potential risk of code deletion.
**Notes**: this also enables us the opportunity to integrate the various work-arounds within CLI into [fastly/kingpin](https://github.com/fastly/kingpin).


Integralist:
@phamann once I get your approval I'll cut a patch release so we're on `0.21.2`.

phamann (approved):
👍🏻 SHIP IT

PR #187, opened 2021-01-07
Dynamic Configuration

Integralist:
**Problem**: there are multiple hardcoded global configuration values that require a new CLI release to be updated.
**Solution**: pull configuration from an API endpoint.
**Notes**: 
The long term intention would be for the configuration to live under `api.fastly.com` but currently that means the response should be in JSON format, which the configuration is _not_ (it's toml, which would mean a substantial change to the code in this PR and would in general be considered a negative impact on the implementation). Subsequently we've placed the config under `developer.fastly.com` and defined the path as `/api/internal/cli-config` to signal that there is no backwards compatibility guarantees.

To save people from scanning the code in this PR, below is:

- Example toml configuration.
- Example VCL service to serve toml configuration.
- Example golang structures to marshal toml configuration.

---

Here is the example toml configuration:

```toml
[fastly]
api_endpoint = "https://api.fastly.com"

[cli]
remote_config = "https://developer.fastly.com/api/internal/cli-config"
ttl = "5m"
last_checked = "2021-01-21T16:07:20Z"

[[user]]
[[user.profile]]
name = "work"
email = "integralist@fastly.com"
token = "..."
[[user.profile]]
name = "personal"
email = "hello@integralist.co.uk"
token = "..."

[language]
  [language.rust]
  toolchain_version = "1.49.0"
  wasm_wasi_target = "wasm32-wasi"
  fastly_sys_constraint = ">= 0.3.7 <= 0.6.0"

[starter-kits]
[[starter-kits.assemblyscript]]
  name = "Default"
  path = "https://github.com/fastly/compute-starter-kit-assemblyscript-default"
  tag = "v0.2.0"
[[starter-kits.rust]]
  name = "Default"
  path = "https://github.com/fastly/compute-starter-kit-rust-default.git"
  branch = "0.6.0"
[[starter-kits.rust]]
  name = "Beacon"
  path = "https://github.com/fastly/compute-starter-kit-rust-beacon-termination.git"
[[starter-kits.rust]]
  name = "Static"
  path = "https://github.com/fastly/compute-starter-kit-rust-static-content.git"
```

Here is the simple api backend code (this will be replaced by an internal fastly implementation):
```vcl
sub vcl_recv {
  #FASTLY recv
  
  if (req.url.path ~ "^/cli/config$") {
    error 600 "Synthetic TOML";
  }
  error 404 "Not Found";
}

sub vcl_error {
  #FASTLY error
  
  if (obj.status == 404) {
    set obj.status = 404;
    set obj.response = "Not Found";
    set obj.http.Content-Type = "text/html; charset=UTF-8";
    synthetic {"
      	<!doctype html>
        <html lang="en">
          <head>
            <meta charset="utf-8">
            <title>Whoops!</title>
          </head>
          <body>
            <h1>FAIL</h1>
          </body>
        </html>
    "};
    
    return(deliver);
  }
  
  if (obj.status == 600) {
    set obj.status = 200;
    set obj.response = "OK";
    set obj.http.Content-Type = "application/toml; charset=utf8";
    
    # we omit sections (e.g. [user]) and keys (e.g. last_checked)
    # as we either shouldn't be setting those at the API layer
    # e.g. we don't know anything about a 'user'
    # or we want the CLI to handle generating the value
    # e.g. the last_checked value we determine within the CLI
    #
    synthetic {"[fastly]
api_endpoint = "https://api.fastly.com"

[cli]
remote_config = "http://developer.fastly.net/cli/config"
ttl = "5m"

[language]
  [language.rust]
  toolchain_version = "1.49.0"
  wasm_wasi_target = "wasm32-wasi"
  fastly_sys_constraint = ">= 0.3.7 <= 0.6.0"
  
[starter-kits]
[[starter-kits.assemblyscript]]
  name = "Default"
  path = "https://github.com/fastly/compute-starter-kit-assemblyscript-default"
  tag = "v0.2.0"
[[starter-kits.rust]]
  name = "Default"
  path = "https://github.com/fastly/compute-starter-kit-rust-default.git"
  branch = "0.6.0"
[[starter-kits.rust]]
  name = "Beacon"
  path = "https://github.com/fastly/compute-starter-kit-rust-beacon-termination.git"
[[starter-kits.rust]]
  name = "Static"
  path = "https://github.com/fastly/compute-starter-kit-rust-static-content.git""};
    
    return(deliver);
  }
}
```

Here is the go structure the toml is mapped to:

```go
// File represents our dynamic application toml configuration.
type File struct {
	Fastly      ConfigFastly              `toml:"fastly"`
	CLI         ConfigCLI                 `toml:"cli"`
	User        ConfigUser                `toml:"user"`
	Language    ConfigLanguage            `toml:"language"`
	StarterKits ConfigStarterKitLanguages `toml:"starter-kits"`
}

// ConfigFastly represents fastly specific configuration.
type ConfigFastly struct {
	APIEndpoint string `toml:"api_endpoint"`
}

// ConfigCLI represents CLI specific configuration.
type ConfigCLI struct {
	RemoteConfig string `toml:"remote_config"`
	TTL          string `toml:"ttl"`
	LastChecked  string `toml:"last_checked"`
}

// ConfigUser represents user specific configuration.
type ConfigUser struct {
	Token string `toml:"token"`
	Email string `toml:"email"`
}

// ConfigLanguage represents C@E language specific configuration.
type ConfigLanguage struct {
	Rust ConfigRust `toml:"rust"`
}

// ConfigLanguage represents Rust C@E language specific configuration.
type ConfigRust struct {
	// ToolchainVersion is the `rustup` toolchain string for the compiler that we
	// support
	ToolchainVersion string `toml:"toolchain_version"`

	// WasmWasiTarget is the Rust compilation target for Wasi capable Wasm.
	WasmWasiTarget string `toml:"wasm_wasi_target"`

	// FastlySysConstraint is a free-form semver constraint for the internal Rust
	// ABI version that should be supported.
	FastlySysConstraint string `toml:"fastly_sys_constraint"`
}

// ConfigStarterKitLanguages represents language specific starter kits.
type ConfigStarterKitLanguages struct {
	AssemblyScript []ConfigStarterKit `toml:"assemblyscript"`
	Rust           []ConfigStarterKit `toml:"rust"`
}

// ConfigStarterKit represents starter kit specific configuration.
type ConfigStarterKit struct {
	Name     string `toml:"name"`
	Path     string `toml:"path"`
	Tag      string `toml:"tag"`
	Branch   string `toml:"branch"`
}
```

### Legacy Configuration Format

If you have a configuration file using the legacy format (see below), then we will automatically migrate your token and email.

```toml
token = "..."
email = "integralist@fastly.com"
endpoint = "https://api.fastly.com"
last_version_check = "2021-02-11T11:48:52Z"
```

We'll extract the `token` and `email` fields and place them in the correct place within the new format:

```
...

[user]
  token = "..."
  email = "integralist@fastly.com"

...
```


pkg/compute/assemblyscript.go
```diff
@@ -52,7 +53,7 @@ func (a AssemblyScript) Verify(out io.Writer) error {
 		return fmt.Errorf("getting package.json path: %w", err)
 	}
 
-	if !common.FileExists(fpath) {
+	if !filesystem.FileExists(fpath) {
```
Integralist, line 56: I wanted to use an existing function `MakeDirectoryIfNotExists` but I couldn't because of a cyclic import issue, so I needed to move those functions into a separate module (as `common` appeared to be containing too much miscellaneous behaviour).

pkg/config/data.go
```diff
@@ -118,15 +123,64 @@ type File struct {
 	Email            string `toml:"email"`
 	Endpoint         string `toml:"endpoint"`
 	LastVersionCheck string `toml:"last_version_check"`
+
+	// Allow the API endpoint itself to be dynamically switched if we wanted it to
+	APIEndpoint string `toml:"api_endpoint"`
```
Integralist, line 128: Not sure if this is useful, but could be interesting if at some point we wanted to dynamically change the backend serving up the configuration data.

pkg/config/data.go
```diff
@@ -118,15 +123,64 @@ type File struct {
...
+	}
+	defer resp.Body.Close()
+
+	_, err = toml.DecodeReader(resp.Body, f)
+	if err != nil {
+		return err
+	}
+
+	// Create the destination directory for the config file
+	basePath := strings.Split(FilePath, "config.toml")[0]
```
Integralist, line 159: Not sure if this will be OK for non-unix systems 🤷🏻

pkg/config/data.go
phamann: Isn't this what the `file.Endpoint` field on line 124 above is already for?

pkg/config/data.go
phamann: I believe you want [`filepath.Base()`](https://golang.org/pkg/path/filepath/#Base) which should be safe cross-platform

pkg/config/data.go
Integralist: oh really!? lol ok then 👍🏻 🙂

pkg/config/data.go
Integralist: Ah nice!

Integralist: @phamann Actually I don't think that is what we want as it gives back `config.toml` instead of the path leading to that section.

Integralist: Ah there we go `filepath.Dir`

cmd/fastly/main.go
```diff
@@ -2,27 +2,27 @@ package main
...
 
 	"github.com/fastly/cli/pkg/app"
+	"github.com/fastly/cli/pkg/check"
 	"github.com/fastly/cli/pkg/common"
 	"github.com/fastly/cli/pkg/config"
 	"github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/cli/pkg/update"
 )
 
+const configEndpoint = "http://integralist-cli-dynamic-config.com.global.prod.fastly.net/"
```
Integralist, line 19: This would obviously be an `api.fastly.com` endpoint.

cmd/fastly/main.go
```diff
@@ -39,11 +39,103 @@ func main() {
 		out            io.Writer = common.NewSyncWriter(os.Stdout)
 	)
 
+	// Extract a subset of configuration options from the local application directory.
+	var file config.ConfigFile
+	err := file.Read(config.FilePath)
+	if err != nil {
+		text.Output(out, `
+			We were unable to locate a local configuration file required to use the CLI.
```
Integralist, line 47: As noted in my design doc... 

**Should we be informing users of the existence of this file?** 
Currently I am displaying messages to the user to inform them when the local configuration file is missing and whether it’s stale, and whether we’ve been able to update it successfully. This feels overly noisy and not necessarily of value to a user. It might be better to hide this output behind the `-v`, `-verbose` flag (or `-d`, `-debug` †)

> † as noted by https://clig.dev/#arguments-and-flags `-v` can be confused with `--version`.

Integralist:
👋🏻 @acfoltzer here is the CLI 'dynamic config' PR. If you review the [last commit](https://github.com/fastly/cli/pull/187/commits/2b284171650e3f22302eb0780527a5920c75b652) you'll see the changes I've implemented to use a value pulled from the remote configuration object (specifically the file `pkg/compute/rust.go`).

pkg/compute/rust.go
```diff
@@ -218,10 +218,14 @@ func (r Rust) Verify(out io.Writer) error {
 	if err != nil {
 		return fmt.Errorf("error fetching latest crate version: %w", err)
 	}
-	latestFastlySys, err := getLatestCrateVersion(r.client, "fastly-sys")
+
+	latestFastlySys, err := semver.NewVersion(r.config.File.Language.Rust.FastlySys)
```
acfoltzer, line 222: I believe this should be `semver.NewConstraint` in order to parse the kinds of ranges we'll want to support. For example, we want to make sure that the fastly-sys in a package is `>= 0.3.0 && < 0.5.0`, not that it equals a precise version.

Integralist: Aha, ok gotcha. I'll fix that up 👍🏻

Integralist:
## TODO

- `semver.NewConstraint` instead of `semver.NewVersion`.
- don't include hardcoded failover, just error.
- implement new starter-kit toml syntax.

pkg/compute/rust.go
Integralist: Actually I just realize that we do create a constraint on a line just following this section. But it looks like we need an explicit min and max version for fastly-sys rather than constraining by a single value to the minor version.

pkg/compute/rust.go
```diff
@@ -213,14 +218,8 @@ func (r Rust) Verify(out io.Writer) error {
...
 	}
-	latestFastlySys, err := getLatestCrateVersion(r.client, "fastly-sys")
-	if err != nil {
-		return fmt.Errorf("error fetching latest crate version: %w", err)
-	}
 
-	// Create a semver constraint to be within the latest minor range or above.
-	// TODO(phamann): Update this to major when fastly-sys hits 1.x.x.
-	fastlySysConstraint, err := semver.NewConstraint(fmt.Sprintf("~%d.%d.0", latestFastlySys.Major(), latestFastlySys.Minor()))
+	fastlySysConstraint, err := semver.NewConstraint(fmt.Sprintf(">= %s <= %s", r.config.File.Language.Rust.FastlySysMin, r.config.File.Language.Rust.FastlySysMax))
```
acfoltzer, line 222: Can we just have this be a free-form constraint? If we ever have to yank a version of the crate we'd want this to be a disjoint range, not just a min-max.

Integralist: No problem. I've updated this 👍🏻

acfoltzer: Thank you! :tada:

pkg/compute/compute_integration_test.go
```diff
@@ -399,92 +571,166 @@ func TestBuildRust(t *testing.T) {
...
+			[[package]]
+			name = "fastly"
+			version = "0.3.2"`,
 			client: versionClient{
-				fastlyVersions:    []string{"0.4.0"},
-				fastlySysVersions: []string{"0.0.0"}, // included to stop REST API failing to not find crate
+				fastlyVersions: []string{"0.4.0"},
 			},
-			wantError:            "fastly-sys crate not found",
+			wantError:            "fastly-sys crate not found", // because fastly <0.4.0 has no fastly-sys
```
acfoltzer, line 647: `fastly 0.3.3` is where `fastly-sys` was introduced, but the earlier `0.3.x` do not have it.

acfoltzer (commented):
Approved from the Rust semantics perspective. Thank you so much!

I am not a Go programmer so I will leave it to someone else to give the genuine :heavy_check_mark:

kailan (changes requested):
This looks awesome! Only a couple of comments on the transition from the old format to the new:

- When users have a config in the old format, the relevant values should be migrated over to the new format. When I ran this branch, I had to re-run `fastly configure` to set my token again.
- When users have a config in the old format, this logic currently seems to skip downloading the new config and instead irreversibly overwrites the config with empty data. (https://github.com/fastly/cli/pull/187/files#diff-3bb21ba08d70ab5924cf7e23ea51e7e68f73f0b164a0b23f1ec980b70a323104R55)

cmd/fastly/main.go
```diff
@@ -39,11 +38,119 @@ func main() {
...
+	// latest version and write that back to disk. To ensure the CLI program
+	// doesn't complete before the write has finished, we block via a channel.
+	waitForWrite := make(chan bool)
+	wait := false
+
+	// Validate if configuration is older than its TTL
+	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) {
+		if verboseOutput {
+			text.Warning(out, `
+Your local application configuration is out-of-date.
```
kailan, line 81: This could be misleading – "local application" makes it sound like it could be an issue with their C@E project. I would maybe explain that this is the "CLI configuration".

Also rather than out of date... "the Fastly CLI configuration has recently changed, we'll refresh your local copy"? Sounds less like a problem.

cmd/fastly/main.go
```diff
@@ -39,11 +38,119 @@ func main() {
...
+We'll refresh this for you in the background and it'll be used next time.
```
triblondon, line 82: 'local application' is likely to be mis-read as the customer's package, not the fastly CLI itself.

```suggestion
Compatibility and versioning information for the Fastly CLI is being updated in the background.  The updated data will be used next time you execute a fastly command.
```

pkg/config/data.go
```diff
@@ -34,6 +38,17 @@ const (
...
+	// RemoteEndpoint represents the API endpoint where we'll pull the dynamic
+	// configuration file from.
+	//
+	// NOTE: once the configuration is stored locally, it will allow for
+	// overriding this default endpoint.
+	RemoteEndpoint = "https://developer.fastly.com/api/internal/cli-config"
+
+	// UpdateSuccessful represents the message shown to a user when their
+	// application configuration has been updated successfully.
+	UpdateSuccessful = "Successfully wrote updated application configuration file to disk."
```
triblondon, line 51:
```suggestion
	UpdateSuccessful = "Successfully updated platform compatibility and versioning information."
```

cmd/fastly/main.go
triblondon: Snap!  I had the same comment - though we can't say for sure that it's changed, only that the current version is out of date.  I made a suggestion in my review.

pkg/config/data.go
```diff
@@ -111,29 +126,115 @@ var FilePath = func() string {
...
+	Token string `toml:"token"`
+	Email string `toml:"email"`
+}
+
+// ConfigLanguage represents C@E language specific configuration.
+type ConfigLanguage struct {
+	Rust ConfigRust `toml:"rust"`
+}
+
+// ConfigLanguage represents Rust C@E language specific configuration.
```
phamann, line 161:
```suggestion
// ConfigRust represents Rust C@E language specific configuration.
```

pkg/config/data.go
```diff
@@ -111,29 +126,115 @@ var FilePath = func() string {
...
+type ConfigRust struct {
+	// ToolchainVersion is the `rustup` toolchain string for the compiler that we
+	// support
+	ToolchainVersion string `toml:"toolchain_version"`
+
+	// WasmWasiTarget is the Rust compilation target for Wasi capable Wasm.
+	WasmWasiTarget string `toml:"wasm_wasi_target"`
```
phamann, line 168: Does this need to be Dynamic configuration? I don't ever see this changing. cc @acfoltzer

pkg/config/data.go
```diff
@@ -111,29 +126,115 @@ var FilePath = func() string {
...
+	Name   string `toml:"name"`
+	Path   string `toml:"path"`
+	Tag    string `toml:"tag"`
+	Branch string `toml:"branch"`
+}
+
+// Load gets the configuration file from the CLI API endpoint and encodes it
+// from memory into config.File.
+func (f *File) Load(configEndpoint string, httpClient api.HTTPClient) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
phamann, line 192: 3 seconds seems very high? Also can we move this timeout to a named constant at the top of the file?

cmd/fastly/main.go
```diff
@@ -39,11 +38,118 @@ func main() {
 		out            io.Writer = common.NewSyncWriter(os.Stdout)
 	)
 
+	// We have to manually handle the inclusion of the verbose flag here because
+	// Kingpin doesn't evaluate the provided arguments until app.Run which
+	// happens later in the file and yet we need to know if we should be printing
+	// output related to the application configuration file in this file.
+	var verboseOutput bool
+	for _, seg := range args {
+		if seg == "-v" || seg == "-verbose" || seg == "--verbose" {
```
phamann, line 47: I don't think `-verbose` is valid?

```suggestion
		if seg == "-v" || seg == "--verbose" {
```

cmd/fastly/main.go
```diff
@@ -39,11 +38,118 @@ func main() {
...
+	// latest version and write that back to disk. To ensure the CLI program
+	// doesn't complete before the write has finished, we block via a channel.
+	waitForWrite := make(chan bool)
+	wait := false
+
+	// Validate if configuration is older than its TTL
+	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) {
+		if verboseOutput {
+			text.Warning(out, `
+Compatibility and versioning information for the Fastly CLI is being updated in the background.  The updated data will be used next time you execute a fastly command.
```
phamann, line 81: General quesiton which applies to a couple of output messages in this PR: whats the reasoning for _"Compatibility and versioning information"_ versus simply _"Configuration"_?

phamann (commented):
Looking good, so happy to see how this has evolved! Couple of minor nits and questions.

cmd/fastly/main.go
kailan: Discussion [here](https://github.com/fastly/cli/pull/187#discussion_r574444593) and [here](https://github.com/fastly/cli/pull/187#discussion_r574486737).

phamann: That discussion is about the use of 'local application' - which I agree with. However, @triblondon's suggested edit of "compatibility and versioning information" seems strange to me and might not always be true in the future, where a generic "configuration" will.

triblondon: I don't love what I suggested here.  It was to solve the 'local application' problem, but 'configuration' is too broad imho.  I think we need to give the user a sense of what this data is for.

pkg/config/data.go
acfoltzer: It changed in the past from `wasm32-unknown-wasi` to `wasm32-wasi`, and while I doubt it'll change again we might as well be prepared

cmd/fastly/main.go
Integralist: I'm OK with this for now.

pkg/compute/compute_integration_test.go
Integralist, line 647:
```suggestion
			wantError:            "fastly-sys crate not found", // fastly 0.3.3 is where fastly-sys was introduced
```

pkg/config/data.go
Integralist: @phamann considering we expect the response to be a lot quicker, is there any real danger in setting it to three seconds? In the case of a freak scenario where the response is slow I'd rather give the client a better chance at acquiring the file rather than setting it too low.

Integralist:
@phamann @kailan I've addressed both your comments. 

If you could please take a look at the last 6 commits, re-review and approve if you're happy. Thanks!

kailan (approved):
Looks great! This is going to be so useful.

cmd/fastly/main.go
```diff
@@ -39,11 +38,125 @@ func main() {
...
+
+	// Extract a subset of configuration options from the local application directory.
+	var file config.File
+	err := file.Read(configFilePath)
+	if err != nil {
+		if verboseOutput {
+			if err == config.ErrLegacyConfig {
+				text.Output(out, `
+					We found your local configuration file (required to use the CLI) was using a legacy format.
+					We will upgrade that file for you now.
```
phamann, line 60: Just noting that I think this is the first time we've used a third person tone of voice "we" in the CLI output. And thus inconsistent with other messages. If thats our preferred tone going forward we should consider starting a document somewhere (internally or in this repo) with best practices for writing output messaging for the product. I personally found the use of "we" here strange 😉

cmd/fastly/main.go
```diff
@@ -39,11 +38,125 @@ func main() {
...
+					We will create that file for you now.
+				`)
+			}
+			text.Break(out)
+		}
+
+		err := file.Load(config.RemoteEndpoint, httpClient)
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1) // TODO: offer a clearer remediation step
```
phamann, line 74: Noting an outstanding `todo`, should we add a proper remediation step to this PR?

cmd/fastly/main.go
phamann: Still find this weird 😬

cmd/fastly/main.go
```diff
@@ -39,11 +38,125 @@ func main() {
...
+
+		wait = true
+		go func() {
+			// NOTE: we no longer use the hardcoded config.RemoteEndpoint constant.
+			// Instead we rely on the values inside of the application
+			// configuration file to determine where to load the config from.
+			err := file.Load(file.CLI.RemoteConfig, httpClient)
+			if err != nil {
+				fmt.Println(err)
+				os.Exit(1) // TODO: offer a clearer remediation step
```
phamann, line 100: Another outstanding TODO?

phamann (approved):
LGTM - A couple of outstanding TODO items and some nits about use of language but none are blocking 😄

cmd/fastly/main.go
Integralist: So something like this instead?

```
Found local configuration file...
File is being upgraded.
```

Integralist:
Tested this again locally...

```bash
go run cmd/fastly/main.go version -v
```

Using a legacy configuration file:

```toml
# /Users/integralist/Library/Application Support/fastly/config.toml

token = "abcdefg"
email = "integralist@fastly.com"
endpoint = "https://api.fastly.com"
last_version_check = "2021-02-11T11:48:52Z"
```

This caused the following output:

```
Found your local configuration file (required to use the CLI) was using a legacy format.
File is being upgraded now.

Fastly API token provided via config file
Fastly CLI version v0.0.0-unknown (unknown)
Built with go version unknown
```

This updated the legacy toml configuration to:

```toml
[cli]
  last_checked = "2021-03-15T16:57:50Z"
  remote_config = "https://developer.fastly.com/api/internal/cli-config"
  ttl = "5m"

[fastly]
  api_endpoint = "https://api.fastly.com"

[language]

  [language.rust]
    fastly_sys_constraint = ">= 0.3.7 <= 0.6.0"
    toolchain_version = "1.49.0"
    wasm_wasi_target = "wasm32-wasi"

[legacy]
  email = "integralist@fastly.com"
  token = "abcdefg"

[starter-kits]

  [[starter-kits.assemblyscript]]
    branch = ""
    name = "Default"
    path = "https://github.com/fastly/compute-starter-kit-assemblyscript-default"
    tag = "v0.2.0"

  [[starter-kits.rust]]
    branch = "0.6.0"
    name = "Default"
    path = "https://github.com/fastly/compute-starter-kit-rust-default.git"
    tag = ""

  [[starter-kits.rust]]
    branch = ""
    name = "Beacon"
    path = "https://github.com/fastly/compute-starter-kit-rust-beacon-termination.git"
    tag = ""

  [[starter-kits.rust]]
    branch = ""
    name = "Static"
    path = "https://github.com/fastly/compute-starter-kit-rust-static-content.git"
    tag = ""

[user]
  email = "integralist@fastly.com"
  token = "abcdefg"
```

PR #192, opened 2021-01-19
Support for Compute@Edge Log Tailing

jaredmorrow:
## What does it do?

Adds command: fastly logs tail --service-id=<sid>

## Output

Because this doesn't run w/o the above mentioned go-fastly, the following is what a log tail session looks like with three requests that just print out a simple message.

```
fastly/cli [ ./fastly logs tail -s <redacted>

INFO: Managed logging enabled on service <redacted>
stdout | 3fbf507b | Hello from the root path!
stdout | 3fbf507b | 2nd println in this request
stdout | 3fbf507b | 3rd println in this request
stdout | 3fbf507b | 4th println in this request
stdout | 3fbf507b | 5th println in this request
stdout | 3fbf507b | 6th println in this request
stdout | c5035791 | Hello from the root path!
stdout | c5035791 | 2nd println in this request
stdout | c5035791 | 3rd println in this request
stdout | c5035791 | 4th println in this request
stdout | c5035791 | 5th println in this request
stdout | c5035791 | 6th println in this request
stdout | 9154fde8 | Hello from the root path!
stdout | 9154fde8 | 2nd println in this request
stdout | 9154fde8 | 3rd println in this request
stdout | 9154fde8 | 4th println in this request
stdout | 9154fde8 | 5th println in this request
stdout | 9154fde8 | 6th println in this request
```

### DIFF
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -114,6 +114,7 @@ COMMANDS
   dictionary       Manipulate Fastly edge dictionaries
   dictionaryitem   Manipulate Fastly edge dictionary items
   logging          Manipulate Fastly service version logging endpoints
+  logs             Compute@Edge Log Tailing
   stats            View statistics (historical and realtime) for a Fastly
                    service
 `) + "\n\n"
@@ -3020,6 +3021,21 @@ COMMANDS
         --version=VERSION        Number of service version
     -n, --name=NAME              The name of the OpenStack logging object
     -s, --service-id=SERVICE-ID  Service ID
+
+  logs tail [<flags>]
+    Tail Compute@Edge logs
+
+    -s, --service-id=SERVICE-ID  Service ID
+        --from=FROM              From time, in unix seconds
+        --to=TO                  To time, in unix seconds
+        --sort-buffer=1s         Sort buffer is how long to buffer logs,
+                                 attempting to sort them before printing,
+                                 defaults to 1s (second)
+        --search-padding=2s      Search padding is how much of a window on
+                                 either side of From and To to use for
+                                 searching, defaults to 2s (seconds)
+        --stream=STREAM          Stream specifies which of 'stdout' or 'stderr'
+                                 to output, defaults to undefined (all streams)
 
   stats regions
     List stats regions
--- a/pkg/logs/tail.go
+++ b/pkg/logs/tail.go
@@ -1,0 +1,604 @@
+package logs
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"os/signal"
+	"sort"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/tomnomnom/linkheader"
+
+	"github.com/fastly/cli/pkg/common"
+	"github.com/fastly/cli/pkg/compute/manifest"
+	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/text"
+
+	"github.com/fastly/go-fastly/v3/fastly"
+)
+
+type (
+	TailCommand struct {
+		common.Base
+		manifest manifest.Data
+		Input    fastly.CreateManagedLoggingInput
+
+		cfg cfg
+
+		dieCh   chan struct{} // channel to end output/printing
+		batchCh chan Batch    // send batches to output loop
+		doneCh  chan struct{} // channel to signal we've reached the end of the run
+
+		hClient *http.Client // TODO: this will go away when GET is in go-fastly
+		token   string       // TODO: this will go away when GET is in go-fastly
+	}
+
+	// cfg holds the configuration parameters passed in through
+	// command line arguments.
+	cfg struct {
+		// path is the full path to fetch
+		path string
+
+		// from is how far in the past to start showing logs.
+		from int64
+
+		// to is when to get logs until.
+		to int64
+
+		// sortBuffer is how long to buffer logs from when the cli
+		// receives them to when the cli prints them. It will sort
+		// by RequestID for that buffer period.
+		sortBuffer time.Duration
+		// searchPadding is how much of a window on either side of
+		// from and to to use for searching for the beginning or
+		// through the end timestamps.
+		searchPadding time.Duration
+		// stream specifies which of stdout or stderr or both the
+		// customer wants to consume.
+		// Undefined == both stderr and stdout.
+		stream string
+	}
+
+	// Log defines the message envelope that compute@edge (C@E) wraps the
+	// user messages in.
+	Log struct {
+		// SequenceNum is the message sequence number used to reorder
+		// messages.
+		SequenceNum int `json:"sequence_number"`
+		// RequestTime is the time in microseconds when the request
+		// was received.
+		RequestStart int64 `json:"request_start_us"`
+		// Stream is the C@E stream, either stdout or stderr.
+		Stream string `json:"stream"`
+		// RequestID is a UUID representing individual requests to the
+		// particular WASM service.
+		RequestID string `json:"id"`
+		// Message is the actual message body the user wants printed.
+		Message string `json:"message"`
+	}
+
+	// Batch encompasses a batch ID and the logs for this batch.
+	Batch struct {
+		ID   string `json:"batch_id"`
+		Logs []Log  `json:"logs"`
+	}
+)
+
+// NewTailCommand returns a usable command registered under the parent.
+func NewTailCommand(parent common.Registerer, globals *config.Data) *TailCommand {
+	var c TailCommand
+	c.Globals = globals
+	c.manifest.File.Read(manifest.Filename)
+	c.CmdClause = parent.Command("tail", "Tail Compute@Edge logs")
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
+	c.CmdClause.Flag("from", "From time, in unix seconds").Int64Var(&c.cfg.from)
+	c.CmdClause.Flag("to", "To time, in unix seconds").Int64Var(&c.cfg.to)
+	c.CmdClause.Flag("sort-buffer",
+		"Sort buffer is how long to buffer logs, attempting to sort them before printing, defaults to 1s (second)").Default("1s").DurationVar(&c.cfg.sortBuffer)
+	c.CmdClause.Flag("search-padding",
+		"Search padding is how much of a window on either side of From and To to use for searching, defaults to 2s (seconds)").Default("2s").DurationVar(&c.cfg.searchPadding)
+	c.CmdClause.Flag("stream", "Stream specifies which of 'stdout' or 'stderr' to output, defaults to undefined (all streams)").StringVar(&c.cfg.stream)
+
+	return &c
+}
+
+// Exec invokes the application logic for the command.
+func (c *TailCommand) Exec(in io.Reader, out io.Writer) error {
+	serviceID, source := c.manifest.ServiceID()
+	if source == manifest.SourceUndefined {
+		return errors.ErrNoServiceID
+	}
+	c.Input.ServiceID = serviceID
+
+	c.Input.Kind = fastly.ManagedLoggingInstanceOutput
+	c.cfg.path = fmt.Sprintf("%s/service/%s/log_stream/managed/instance_output", config.DefaultEndpoint, c.Input.ServiceID)
+
+	c.dieCh = make(chan struct{})
+	c.batchCh = make(chan Batch)
+	c.doneCh = make(chan struct{})
+
+	c.hClient = http.DefaultClient
+	c.token, _ = c.Globals.Token()
+
+	// Adjust the from/to times if they are
+	// defined. We adjust the times based on searchPadding.
+	c.adjustTimes()
+
+	// Enable managed logging if not already enabled.
+	if err := c.enableManagedLogging(out); err != nil {
+		return err
+	}
+
+	sigs := make(chan os.Signal, 2)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+
+	// Start the output loop.
+	go c.outputLoop(out)
+
+	// Start tailing the logs.
+	go c.tail(out)
+
+	<-sigs
+	close(c.dieCh)
+
+	return nil
+}
+
+//
+// Client
+//
+
+// Tail starts the virtual tail process. Tail fetches data from the eventbuffer
+// API. It hands off the requested logs to the outputloop for the actual
+// printing.
+func (c *TailCommand) tail(out io.Writer) {
+	// Start this with --from and --to if set.
+	curWindow := c.cfg.from
+	toWindow := c.cfg.to
+
+	// Start the loop with an initial address to query.
+	path := makeNewPath(out, c.cfg.path, curWindow, "")
+
+	// lastBatchID keeps the last successfully read Batch.ID in case we need
+	// re-request on failure.
+	var lastBatchID string
+
+	for {
+		// Check to see if we already passed the "to" requirement.
+		if toWindow != 0 && curWindow > toWindow {
+			text.Info(out, "Reached window: %v which is newer than the requested 'to': %v", curWindow, toWindow)
+			// We are done, but we still want printing to finish.
+			close(c.doneCh)
+			break
+		}
+
+		req, err := http.NewRequest("GET", path, nil)
+		req.Header.Add("Fastly-Key", c.token)
+		if err != nil {
+			text.Error(out, "unable to create new request: %v", err)
+			os.Exit(1)
+		}
+
+		resp, err := c.doReq(req)
+		if err != nil {
+			text.Error(out, "unable to request: %v", err)
+			os.Exit(1)
+		}
+
+		// Check that our request was successful. If the server is
+		// having trouble, retry after waiting for some time.
+		if resp.StatusCode != http.StatusOK {
+			// In an effort to clean up the output, do not print on
+			// 503's.
+			if resp.StatusCode != http.StatusServiceUnavailable {
+				text.Warning(out, "non-200 resp %d", resp.StatusCode)
+			}
+			io.Copy(ioutil.Discard, resp.Body)
+			// Try the response again after a 1 second wait.
+			if resp.StatusCode/100 == 5 && resp.StatusCode != 501 ||
+				resp.StatusCode == 429 {
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			// Failing at this point is unrecoverable,
+			return
+		}
+
+		// Read and parse response, send batches to the output loop.
+		scanner := bufio.NewScanner(resp.Body)
+		// Use a 10MB buffer for the bufio scanner, as we don't know
+		// how big some of the responses will be.
+		const tmb = 10 << 20
+		buf := make([]byte, tmb)
+		scanner.Buffer(buf, tmb)
+
+		for scanner.Scan() {
+			// Scan one line at a time, and get only one batch
+			// at a time.
+			b := scanner.Bytes()
+			batch, err := parseResponseData(b)
+			if err != nil {
+				// We can't parse the response, attempt to
+				// re-request from the last window & batch.
+				text.Warning(out, "unable to parse resp body: %v", err)
+				path = makeNewPath(out, path, curWindow, lastBatchID)
+				continue
+			}
+
+			// If we got a batch back, there will be an ID.
+			if batch.ID != "" {
+				// Record last batchID in case
+				// anything fails along the way, we
+				// can re-request.
+				lastBatchID = batch.ID
+				// Send batch down batchCh to the output loop.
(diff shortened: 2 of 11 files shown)


Integralist:
@jaredmorrow thanks for opening this PR. Seems to be a couple of failing tests due to a missing method expected on an interface.

jaredmorrow:
@Integralist go-fastly v3 required :)

jaredmorrow:
With go-fastly v3, it is now passing tests.

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+		// having trouble, retry after waiting for some time.
+		if resp.StatusCode != http.StatusOK {
+			// In an effort to clean up the output, do not print on
+			// 503's.
+			if resp.StatusCode != http.StatusServiceUnavailable {
+				text.Warning(out, "non-200 resp %d", resp.StatusCode)
+			}
+			io.Copy(ioutil.Discard, resp.Body)
+			// Try the response again after a 1 second wait.
+			if resp.StatusCode/100 == 5 && resp.StatusCode != 501 ||
```
Integralist, line 203: Do we want to implement a form of exponential backoff? (not necessary for this PR but just a thought)

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+				time.Sleep(1 * time.Second)
+				continue
+			}
+			// Failing at this point is unrecoverable,
+			return
+		}
+
+		// Read and parse response, send batches to the output loop.
+		scanner := bufio.NewScanner(resp.Body)
+		// Use a 10MB buffer for the bufio scanner, as we don't know
```
Integralist, line 214: Should this be configurable some day in the future, or is 10mb totally sufficient?

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+
+		if err := scanner.Err(); err != nil {
+			// ErrUnexpectedEOFs need to be retried, but they
+			// produce a lot of noise for the user, so don't log.
+			if err != io.ErrUnexpectedEOF {
+				text.Warning(out, "error scanning response body: %v", err)
+			}
+			// Something happened in the scanner, re-request the
+			// current batchID.
+			path, err = makeNewPath(path, curWindow, lastBatchID)
```
Integralist, line 259: This call and the subsequent error failure handling logic appears in a few places. Might be overkill but could we shift to a separate function?

Integralist (commented):
@jaredmorrow I'm not all the way done yet. Stopped at the `tail_test.go` file, and I'll continue on from there will my review shortly (just gotta step away for a bit).

pkg/logs/tail.go
jaredmorrow: 10mb is probably overkill as it is.  Okay to leave.

pkg/logs/tail.go
jaredmorrow: At some point I had a bunch of TODO's saying that it was annoying we did the error handling over and over.  I think I might just make `makeNewPath` never return an error.

pkg/logs/tail.go
jaredmorrow: Possibly, but the window size is only 10 seconds, so it might be overkill atm.

pkg/app/run_test.go
```diff
@@ -3021,6 +3022,20 @@ COMMANDS
     -n, --name=NAME              The name of the OpenStack logging object
     -s, --service-id=SERVICE-ID  Service ID
 
+  logs tail --service-id=SERVICE-ID [<flags>]
+    Tail Compute@Edge logs
+
+    --service-id=SERVICE-ID  Service ID
+    --from=FROM              From time, in unix seconds
+    --to=TO                  To time, in unix seconds
+    --sort-buffer=1s         Sort buffer is how long to buffer logs before
```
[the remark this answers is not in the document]
jaredmorrow: I went with sort-buffer firstly as you mentioned it was intended to sort out-of-order incoming messages.  Secondly, it was described in detail in our specification and docs as sort-buffer.

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+
+	// Log defines the message envelope that compute@edge (C@E) wraps the
+	// user messages in.
+	Log struct {
+		// SequenceNum is the message sequence number used to reorder
+		// messages.
+		SequenceNum int `json:"sequence_number"`
+		// RequestTime is the time in microseconds when the request
+		// was received.
+		RequestStart int64 `json:"request_start_us"`
```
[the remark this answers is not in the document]
jaredmorrow: Microseconds.  `ms` would be millisecond

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+
+	if c.cfg.to != 0 {
+		// Adjust to based on search padding, we want look forward more.
+		c.cfg.to = c.cfg.to + int64(c.cfg.searchPadding.Seconds())
+	}
+}
+
+// enableManagedLogging enables managed logging in our API.
+func (c *TailCommand) enableManagedLogging(out io.Writer) error {
+	text.Output(out, "Enabling managed logging on your service (this might take a few seconds)...")
```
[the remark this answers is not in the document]
jaredmorrow: Do we quietly just do it behind the scenes then?  Unfortunately there is no way to check if it is enabled w/o doing the call.  I also don't like doing it on every run.  I'll remove this text.

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+	// defined. We adjust the times based on searchPadding.
+	c.adjustTimes()
+
+	// Enable managed logging if not already enabled.
+	if err := c.enableManagedLogging(out); err != nil {
+		return err
+	}
+
+	sigs := make(chan os.Signal, 2)
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
```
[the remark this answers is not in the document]
jaredmorrow: It is weird.  I can never remember, so I just use the same line over and over in every application ;)

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+		case batch := <-c.batchCh: // Got new batch.
+			// Range through batch logs, for each
+			// RequestID we create a timer based on the
+			// highest SequenceNum we got in this batch
+			// for that RequestID.  If a timer already
+			// exists for the RequestID, we append the new
+			// time.Now() and high SequenceNum.  At most
+			// there should be one timer per RequestID.
+			for reqid, ls := range splitByReqID(batch.Logs) {
+				req := reqid
```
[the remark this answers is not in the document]
jaredmorrow: This is actually a VERY nasty bug that took me hours to figure out.  `req` is used below in `Afterfunc` around line 389 and if you don't do this assignment every timer will get the same request ID.

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+		{
+			in:  cfg{searchPadding: dur},
+			exp: cfg{searchPadding: dur},
+		},
+	} {
+		c := TailCommand{cfg: test.in}
+		c.adjustTimes()
+		if equal := reflect.DeepEqual(test.exp, c.cfg); !equal {
+			t.Errorf("#%d: validateTimes mismatch: got: %#+v  want: %#+v", i, c.cfg, test.exp)
```
Integralist, line 37: Presumed this was an old name?
```suggestion
			t.Errorf("#%d: adjustTimes mismatch: got: %#+v  want: %#+v", i, c.cfg, test.exp)
```

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+					func(i, j int) bool {
+						return reqLogs.logs[i].SequenceNum < reqLogs.logs[j].SequenceNum
+					})
+
+				// Check to see if we already have a timer
+				// running or if the current high sequence is
+				// higher than the one with the timer.
+				// The timer will always be running on the head
+				// of the slice.
+				recv := reqLogs.rcvs
```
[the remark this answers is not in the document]
jaredmorrow: will fix

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+		in   []Log
+		want map[string][]Log
+	}{
+		{in: full, want: expfull},
+		{in: single, want: expsingle},
+		{in: []Log{}, want: make(map[string][]Log)},
+	} {
+		got := splitByReqID(test.in)
+		if diff := cmp.Diff(test.want, got); diff != "" {
+			t.Errorf("#%d: parseResponseData mismatch (-want +got):\n%s", i, diff)
```
Integralist, line 95: I believe this should be what's referenced as that's the function actually called.
```suggestion
			t.Errorf("#%d: splitByReqID mismatch (-want +got):\n%s", i, diff)
```

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+
+				// Record highest SequenceNum in this new batch
+				// for this RequestID
+				rHighSeq := highSequence(ls)
+
+				// Whether we have the RequestID or not, we
+				// append and sort the logs slice.
+				reqLogs := logmap[req]
+				reqLogs.logs = append(reqLogs.logs, ls...)
+				// Sort after appending.
```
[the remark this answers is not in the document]
jaredmorrow: Fixed

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+			return
+		case batch := <-c.batchCh: // Got new batch.
+			// Range through batch logs, for each
+			// RequestID we create a timer based on the
+			// highest SequenceNum we got in this batch
+			// for that RequestID.  If a timer already
+			// exists for the RequestID, we append the new
+			// time.Now() and high SequenceNum.  At most
+			// there should be one timer per RequestID.
+			for reqid, ls := range splitByReqID(batch.Logs) {
```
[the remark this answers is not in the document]
jaredmorrow: Fixed

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+}
+
+// TestParseResponseData tests parseResponseData() and logsfromBatch()
```
Integralist, line 100:
```suggestion
// TestParseResponseData tests parseResponseData()
```

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+// Exec invokes the application logic for the command.
+func (c *TailCommand) Exec(in io.Reader, out io.Writer) error {
+	c.Input.Kind = fastly.ManagedLoggingInstanceOutput
+	c.cfg.path = fmt.Sprintf("%s/service/%s/log_stream/managed/instance_output", config.DefaultEndpoint, c.Input.ServiceID)
+
+	c.dieCh = make(chan struct{})
+	c.batchCh = make(chan Batch)
+	c.doneCh = make(chan struct{})
+
+	c.hClient = http.DefaultClient
```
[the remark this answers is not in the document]
jaredmorrow: I don't know if we want to timeout in case they actually know what they want and are requesting a window ahead of time and would like to wait on it.  Either way, this client will be going away when go-fastly has the GET logic in it.

pkg/logs/tail.go
```diff
@@ -0,0 +1,610 @@
...
+		Logs []Log  `json:"logs"`
+	}
+)
+
+// NewTailCommand returns a usable command registered under the parent.
+func NewTailCommand(parent common.Registerer, globals *config.Data) *TailCommand {
+	var c TailCommand
+	c.Globals = globals
+	c.CmdClause = parent.Command("tail", "Tail Compute@Edge logs")
+	c.CmdClause.Flag("service-id", "Service ID").Required().StringVar(&c.Input.ServiceID)
```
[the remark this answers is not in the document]
jaredmorrow: Ah, I wasn't aware this existed, I'll use it.  Thanks.

pkg/logs/tail_test.go
Integralist: Side note, this docstring doesn't really tell us much about what the test is validating.

For example, I'd expect something more along the lines of:

```go
// TestParseResponseData validates we're correctly decoding a batch JSON log
// response into a logs.Batch type.
```

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	} {
+		out := filterStream(test.stream, test.logs)
+		if len(out) != test.explen {
+			t.Errorf("#%d: exp: %d != got: %d", i, test.explen, len(out))
+		}
+	}
+}
+
+// TestGetLinks tests both getLinks() and getTimeFromLink()
```
Integralist, line 189: As mentioned above, it would be nice if the docstrings could be a bit more _descriptive_ of what they're trying to achieve. Otherwise I don't see much _value_ in the current description. I guess if I was to rephrase what I'm asking it would be: help me to understand the purpose of the test without me having to read the code.

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	pTime, err := getTimeFromLink(prev)
+	if err != nil {
+		t.Fatalf("unexpected error parsing prev link: %s", err)
+	}
+	exp := int64(1601412620)
+	if pTime != exp {
+		t.Errorf("prev time exp: %v != got: %v", exp, pTime)
+	}
```
Integralist, line 213: I guess for completeness we should also validate a `nTime` (i.e. the time is extracted from the 'next' link).

Integralist (commented):
Here's the second half of my review (although I see you've pushed some changes so maybe some of this has become irrelevant).

pkg/logs/tail.go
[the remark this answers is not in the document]
jaredmorrow: Fixed.  This is much more clean, thanks.

pkg/logs/tail.go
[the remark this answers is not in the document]
Integralist: oooh ouch. good catch!

pkg/logs/tail.go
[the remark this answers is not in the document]
Integralist: gotcha 👍🏻

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+	// defined. We adjust the times based on searchPadding.
+	c.adjustTimes()
+
+	// Enable managed logging if not already enabled.
+	if err := c.enableManagedLogging(out); err != nil {
+		return err
+	}
+
+	sigs := make(chan os.Signal, 2)
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
```
Integralist, line 144: OK so this: https://go-review.googlesource.com/c/go/+/187739 << suggests that SIGTERM and SIGINT are both already ok to use then when compiled for Windows. It looks like they get internally re-mapped to the posix equivalent in the signal handler. So we should be OK to swap to syscall for a bit more consistency:
```suggestion
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
```

But I'm also happy for this to be left as is if you prefer not to discover some strange edge case.

Integralist (commented):
One other final comment, not essential.

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// enableManagedLogging enables managed logging in our API.
+func (c *TailCommand) enableManagedLogging(out io.Writer) error {
+	_, err := c.Globals.Client.CreateManagedLogging(&c.Input)
+	if err != nil && err != fastly.ErrManagedLoggingEnabled {
+		return err
+	}
+
+	text.Success(out, "Managed logging enabled on service %s", c.Input.ServiceID)
```
Integralist, line 301: Do we also want to remove this message (for the same reason as removing the other message), and if so this is something else that should be removed/reflected in the blog post.

jaredmorrow: I don't think we should.  This will go out and make the call every time, sometimes with a large delay (I've noticed).  Without the message, it makes for an odd silence in the CLI output.  We can make a -q (--quiet) flag down the road if people want a perfectly clean output they can pipe to something else.

pkg/logs/tail_test.go
```diff
@@ -0,0 +1,323 @@
...
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+const responseFile = "testdata/response.json"
+
+func TestValidateTimes(t *testing.T) {
```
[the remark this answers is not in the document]
jaredmorrow: Will fix

pkg/logs/tail_test.go
jaredmorrow: Yes, and will also change the test name.

pkg/logs/tail_test.go
jaredmorrow: Yes, and will add a comment to the test function.

pkg/logs/tail_test.go
jaredmorrow: Fixed

pkg/logs/tail_test.go
jaredmorrow: Fixed

pkg/logs/tail.go
jaredmorrow: Fixed

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+	Batch struct {
+		ID   string `json:"batch_id"`
+		Logs []Log  `json:"logs"`
+	}
+)
+
+// NewTailCommand returns a usable command registered under the parent.
+func NewTailCommand(parent common.Registerer, globals *config.Data) *TailCommand {
+	var c TailCommand
+	c.Globals = globals
```
Integralist, line 102: We're missing:

```go
```

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// enableManagedLogging enables managed logging in our API.
+func (c *TailCommand) enableManagedLogging(out io.Writer) error {
+	_, err := c.Globals.Client.CreateManagedLogging(&c.Input)
+	if err != nil && err != fastly.ErrManagedLoggingEnabled {
+		return err
+	}
+
+	text.Success(out, "Managed logging enabled on service %s", c.Input.ServiceID)
```
Integralist, line 301: Patrick suggested we should only use `text.Success` as the very end output for a command, so we should look to replace it with `text.Info` instead...

```suggestion
	text.Info(out, "Managed logging enabled on service %s", c.Input.ServiceID)
```

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// printLogs is a simple printer for Log slices, only printing requested
+// streams.
+func (c *TailCommand) printLogs(out io.Writer, logs []Log) {
+	if len(logs) > 0 {
+		filtered := filterStream(c.cfg.stream, logs)
+
+		for _, l := range filtered {
+			text.Output(out, l.String())
```
Integralist, line 448: The `text.Output` function internally 'wraps' the output where as this sort of data should flow as wide on the screen as necessary otherwise users will find that piping the data to another command will break.

It was suggested by Patrick that there isn't another function within the `text` package we could use that didn't also wrap the text output so we should look to replace `text.Output` with a standard `fmt.Print<T>` variation.

pkg/logs/tail.go
```diff
@@ -0,0 +1,603 @@
...
+	// RequestTime comes as unix time in microseconds. Convert to
+	// nanoseconds, then parse with stdlib.
+	nano := l.RequestStart * 1000
+	return time.Unix(0, nano)
+}
+
+// String is used to print a log for the tail output.
+func (l *Log) String() string {
+	// Trim the RequestID for nicer output, it might be a long UUID.
+	return fmt.Sprintf("| %6s | %8.8s | %s |",
```
Integralist, line 485: We should lose the left most pipe:

```suggestion
	return fmt.Sprintf("%6s | %8.8s | %s |",
```

pkg/logs/tail.go
Integralist: There was a long gap between seeing this message and anything happening. We should use the 'spinner' code in the CLI to indicate to the user that something is happening.

Either A.) we're waiting for buffered logs, or B.) we're waiting for logs in general to come through.

pkg/logs/tail.go
jaredmorrow: Fixed

pkg/logs/tail.go
jaredmorrow: Changed to INFO.  Discussed with @freeformz as it would adversely affect piping, we are not going to do a spinner.

pkg/logs/tail.go
jaredmorrow: Removed leading and trailing separator.  That will make using `cut` easier if need be

pkg/logs/tail.go
freeformz: I dm'd @phamann and he agreed to drop it because IME things like spinners interfere with certain pipes users may want to construct to filter log tails. If we get feedback that people don't know what is happening while waiting we can address it in a future update.

pkg/logs/tail.go
jaredmorrow: changed to fprintln

Integralist (approved):
LGTM. I'll merge once the CI tests have passed.

PR #200, opened 2021-02-03
Update GitHub Action to not use commit hash

Integralist:
**Problem**:  GitHub are deprecating the ability to reference an Action via a commit hash.

### DIFF
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Run GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
-        version: latest
+        version: v2.4.1
         args: release --rm-dist --release-notes=RELEASE_CHANGELOG.md
       env:
         GOVERSION: ${{ env.GOVERSION }}
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -26,9 +26,9 @@ jobs:
       env:
         CHANGELOG_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Run GoReleaser
-      uses: goreleaser/goreleaser-action@108bc4d
+      uses: goreleaser/goreleaser-action@v2
       with:
-        version: 'v0.135.0'
+        version: latest
         args: release --rm-dist --release-notes=RELEASE_CHANGELOG.md
       env:
         GOVERSION: ${{ env.GOVERSION }}
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -32,4 +32,4 @@ jobs:
         args: release --rm-dist --release-notes=RELEASE_CHANGELOG.md
       env:
         GOVERSION: ${{ env.GOVERSION }}
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tag_release.yml
```diff
@@ -26,9 +26,9 @@ jobs:
...
         CHANGELOG_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Run GoReleaser
-      uses: goreleaser/goreleaser-action@108bc4d
+      uses: goreleaser/goreleaser-action@v2
       with:
-        version: 'v0.135.0'
+        version: latest
         args: release --rm-dist --release-notes=RELEASE_CHANGELOG.md
       env:
         GOVERSION: ${{ env.GOVERSION }}
```
Integralist, line 34: @phamann @kailan @doramatadora I'm not sure if any of you know if the below secret needs to be updated from `RELEASE_GITHUB_TOKEN` to just `GITHUB_TOKEN`?

I read through https://docs.github.com/en/actions/reference/authentication-in-a-workflow but it doesn't mention a `RELEASE_GITHUB_TOKEN` which makes me wonder if that is a legacy reference from back when we were using the older version of goreleaser.

Integralist: Looking at the terraform provider which uses v2 of goreleaser and that's using `GITHUB_TOKEN` https://github.com/fastly/terraform-provider-fastly/blob/master/.github/workflows/release.yml#L40

.github/workflows/tag_release.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       env:
         CHANGELOG_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Run GoReleaser
-      uses: goreleaser/goreleaser-action@108bc4d
+      uses: goreleaser/goreleaser-action@v2
       with:
-        version: 'v0.135.0'
+        version: latest
```
phamann, line 31: There was a preference from team-sec from our original security review of this project that we pin to specfic versions instead of using `latest`.

phamann (changes requested):
LGTM - other than the goreleaser pin.

.github/workflows/tag_release.yml
Integralist: @phamann I'm not sure where this 'version' came from. Looking at the goreleaser repo it didn't have this version as either a tag or a branch. There was a [`0.1.0`](https://github.com/goreleaser/goreleaser-action/releases/tag/v0.1.0) but not `0.135.0` 🤔

Integralist: Although now looking at the code. I'm not use what `with` `version` actually means?

Integralist: Ah ok so reading https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswith this would suggest that `version` is being passed into the goreleaser, so there appears to be _two_ distinct 'versions' at play: 

1. the version we pass in (e.g. `latest` or `v0.135.0`)
2. the release version of goreleaser (e.g. `@v2`)

Integralist: Yup so the version `v0.135.0` _should_ have aligned with some kind of tag I think based on the goreleaser docs:

https://github.com/goreleaser/goreleaser-action#inputs

> Can be a fixed version like v0.117.0 or a max satisfying semver one like ~> 0.132. In this case this will return v0.132.1.

But I'm still not sure where we got that original version from as it doesn't show up anywhere 🤔

Integralist:
@phamann I've fixed to the latest release of goreleaser now. Ready for another review whenever you have a moment.

phamann (approved):
LGTM

PR #201, opened 2021-02-03
Logs Tail: Give the user better feedback when --from flag errors

jaredmorrow:
This fixes two things:
- When an unknown status code is returned we previously return from
the function without and exit. This would just leave us in and endless
loop doing nothing for the user.
- When a user specifies a --from time that is too far in the past or
future, we should let them know that so it doesn't just mysteriously
give them a 404 warning and exit.

Previously it would respond like the following and just hang:
```
WARNING: non-200 resp 404
```

With this change it looks errors and exits:
```
ERROR: specified from time 1622362928 not found, either too far in the past or future
```

### DIFF
--- a/pkg/logs/tail.go
+++ b/pkg/logs/tail.go
@@ -201,6 +201,13 @@ func (c *TailCommand) tail(out io.Writer) {
 		// Check that our request was successful. If the server is
 		// having trouble, retry after waiting for some time.
 		if resp.StatusCode != http.StatusOK {
+			// If the response was a 404, the from time was
+			// not valid, give them an error stating this and exit.
+			if resp.StatusCode == http.StatusNotFound &&
+				c.cfg.from != 0 {
+				text.Error(out, "specified from time %d not found, either too far in the past or future", c.cfg.from)
+				os.Exit(1)
+			}
 			// In an effort to clean up the output, do not print on
 			// 503's.
 			if resp.StatusCode != http.StatusServiceUnavailable {
@@ -213,8 +220,9 @@ func (c *TailCommand) tail(out io.Writer) {
 				time.Sleep(1 * time.Second)
 				continue
 			}
-			// Failing at this point is unrecoverable,
-			return
+			// Failing at this point is unrecoverable.
+			text.Error(out, "unrecoverable error, response code: %d", resp.StatusCode)
+			os.Exit(1)
 		}
 
 		// Read and parse response, send batches to the output loop.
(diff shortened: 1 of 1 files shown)


pkg/logs/tail.go
```diff
@@ -201,6 +201,13 @@ func (c *TailCommand) tail(out io.Writer) {
 		// Check that our request was successful. If the server is
 		// having trouble, retry after waiting for some time.
 		if resp.StatusCode != http.StatusOK {
+			// If the response was a 404, the from time was
+			// not valid, give them an error stating this and exit.
+			if resp.StatusCode == http.StatusNotFound &&
+				c.cfg.from != 0 {
+				text.Error(out, "specified from time %d not found, either too far in the past or future", c.cfg.from)
```
Integralist, line 208: I personally prefer calling out words that _mean_ something in the tech sense, so `from` referring to a value provided by the user. Without some kind of indicator around the word (in this case I've used single quotes) I feel like it's too easy to miss what's being suggested...

```suggestion
				text.Error(out, "specified 'from' time %d not found, either too far in the past or future", c.cfg.from)
```

Integralist (commented):
One minor suggested change.

I would ideally like to see a test around this change.

Integralist (approved):

pkg/logs/tail.go
```diff
@@ -201,6 +201,13 @@ func (c *TailCommand) tail(out io.Writer) {
 		// Check that our request was successful. If the server is
 		// having trouble, retry after waiting for some time.
 		if resp.StatusCode != http.StatusOK {
+			// If the response was a 404, the from time was
+			// not valid, give them an error stating this and exit.
+			if resp.StatusCode == http.StatusNotFound &&
+				c.cfg.from != 0 {
```
ejthurgo, line 207: Sorry if I'm missing context, how does `c.cfg.from != 0` mean the from date is invalid?
I fear this would catch and mislabel any 404 error?

Integralist: @jaredmorrow are you able to provide any extra context? I understand the point @ejthurgo is making but I was under the impression this logic was based on a specific behaviour inherent to your API (e.g. in the sense that the API would only return a 404 when an incorrect flag value was provided). 

If that's not the case and the API is expected to return a 404 in other scenarios, then yes this could end up be an accidental match where the user provides a valid flag value but something else in their request has caused the API to return a 404.

jaredmorrow: The combination of the 404 & the from field being set is the key here.  If from is set and it is out of range, the 404 returned would be due to the range being too far out from Now.  There are currently no other 404s sent by the server, but just in case I added the `from != 0` to be specific about the case.

ejthurgo: Surprising the server doesn't give any specific detail on why it failed? Relying on the fact it **currently** only sends a 404 for this circumstance seems awfully risky.
But thanks for the context.

PR #203, opened 2021-02-08
add port prompt for compute init

Integralist:
When specifying a backend as part of the interactive prompt input for the `fastly compute init` command the default behaviour is to create the backend without TLS (i.e. HTTPS/port 443).

This is the prompt currently displayed: 

```
Backend (originless, hostname or IP address): [originless] mock-s3.edgecompute.app
```

When the Backend is created it's actually assigned port 80 (i.e. HTTP) which the mock backend isn't accessible via.

We should provide a prompt for specifying the port so the user has control over this value.

### DIFF
--- a/pkg/compute/compute_integration_test.go
+++ b/pkg/compute/compute_integration_test.go
@@ -51,7 +51,7 @@ func TestInit(t *testing.T) {
 		},
 		{
 			name:       "unkown repository",
-			args:       []string{"compute", "init", "--from", "https://example.com/template"},
+			args:       []string{"compute", "init", "--from", "https://example.com/template", "--backend-port", "80"},
 			configFile: config.File{Token: "123"},
 			api: mock.API{
 				GetTokenSelfFn:  tokenOK,
@@ -67,7 +67,7 @@ func TestInit(t *testing.T) {
 		},
 		{
 			name:       "create service error",
-			args:       []string{"compute", "init"},
+			args:       []string{"compute", "init", "--backend-port", "80"},
 			configFile: config.File{Token: "123"},
 			api: mock.API{
 				GetTokenSelfFn:  tokenOK,
@@ -78,7 +78,7 @@ func TestInit(t *testing.T) {
 		},
 		{
 			name:       "create domain error",
-			args:       []string{"compute", "init"},
+			args:       []string{"compute", "init", "--backend-port", "80"},
 			configFile: config.File{Token: "123"},
 			api: mock.API{
 				GetTokenSelfFn:  tokenOK,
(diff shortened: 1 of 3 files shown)


pkg/compute/init.go
```diff
@@ -96,8 +97,9 @@ func NewInitCommand(parent common.Registerer, client api.HTTPClient, globals *co
 	c.CmdClause.Flag("branch", "Git branch name to clone from package template repository").Hidden().StringVar(&c.branch)
 	c.CmdClause.Flag("tag", "Git tag name to clone from package template repository").Hidden().StringVar(&c.tag)
 	c.CmdClause.Flag("path", "Destination to write the new package, defaulting to the current directory").Short('p').StringVar(&c.path)
-	c.CmdClause.Flag("domain", "The name of the domain associated to the package").StringVar(&c.path)
-	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.path)
+	c.CmdClause.Flag("domain", "The name of the domain associated to the package").StringVar(&c.domain)
+	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.backend)
```
Integralist, line 101: I noticed these flags were assigning values to the wrong variable 🤦🏻 so have fixed in this PR.

Integralist, line 101: I noticed these flags were assigning values to the wrong variable 🤦🏻 so have fixed in this PR.

pkg/compute/compute_integration_test.go
```diff
@@ -51,7 +51,7 @@ func TestInit(t *testing.T) {
 		},
 		{
 			name:       "unkown repository",
-			args:       []string{"compute", "init", "--from", "https://example.com/template"},
+			args:       []string{"compute", "init", "--from", "https://example.com/template", "--backend-port", "80"},
```
Integralist, line 54: @phamann I'm having to pass this flag in everywhere to prevent the integration tests from failing. Which is a smell.

The fundamental issue is that there is no 'prompt' in the integration test and so if I don't pass the flag, then the code will error when it tries to parse an empty string into an integer (because the code logic expects a user to have entered some value when prompted).

I'd ideally like to mock the behaviour somehow so that I don't need to pass the flag everywhere, but I'm not sure of the best approach to achieve that 🤔

Integralist: One thing I could do, without mocking (and would mean I could remove the `--backend-port` flag from every test setup), is if we get an error when trying to parse the given port number from the user, to just default to port 80 rather than return the error (we would then have to display a message to the user so they were aware that there was an issue).

In reality, the only time we'd get an error parsing is if someone purposely provides an invalid value (like "xyz") or if they make a mistake like prefixing the port number with a colon ":443" otherwise I don't expect users to see this sort of error.

doramatadora (approved):
These changes look fine to me :)

PR #204, opened 2021-02-11
Override error message in compute init

kailan:
As reported by @ezambrano-fastly, users without the C@E feature flag enabled currently see a relatively cryptic message: "Unable to create service: valid values for 'type' are: 'vcl'"

### DIFF
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -361,7 +361,7 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 					Remediation: fmt.Sprintf("see more at https://fastly.dev/learning/compute/#create-a-new-fastly-account-and-invite-your-collaborators"),
 				}
 			}
-		  return fmt.Errorf("error creating service: %w", err)
+			return fmt.Errorf("error creating service: %w", err)
 		}
 		version = 1
 		undoStack.Push(func() error {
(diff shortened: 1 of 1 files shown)


pkg/compute/init.go
```diff
@@ -355,6 +355,12 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			Comment: description,
 		})
 		if err != nil {
+			if strings.Contains(err.Error(), "Valid values for 'type' are: 'vcl'") {
```
Integralist, line 358: Using `strings.Contains` for checking an error always make me a little nervous due to its brittle nature. Sadly I'm not sure if there's a way around this. I just wanted to vocalise that this might be a problem that resurfaces if the backend API ever changes the text content.

pkg/compute/init.go
```diff
@@ -355,6 +355,12 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+				return errors.RemediationError{
+					Inner:       fmt.Errorf("error creating service: you do not have the Compute@Edge feature flag enabled on your Fastly account"),
+					Remediation: "see more at https://fastly.dev/learning/compute/#create-a-new-fastly-account-and-invite-your-collaborators",
```
Integralist, line 361: Looking at other remediation messages they all look to start capitalised...

```suggestion
					Remediation: "See more at https://fastly.dev/learning/compute/#create-a-new-fastly-account-and-invite-your-collaborators",
```

Integralist (approved):
One requested change, but approved still.

pkg/compute/init.go
kailan: Yep, this is not great – we would want an error code to match on in an ideal world. For now this will have to do though, unless anyone has any suggestions.

PR #210, opened 2021-03-10
Implement manifest_version into the fastly.toml

Integralist:
**Problem**: the fastly.toml has no versioning semantics. 
**Solution**: when the CLI reads fastly.toml ensure it has a `manifest_version` of `1`.
**Notes**: we don't overwrite the original fastly.toml with the correct manifest_version type (the correctly converted type is used in-memory only).


kailan:
There are some manifests floating around with the following formats:
```toml
manifest_version = 1
```
```toml
manifest_version = "0.1.0"
```

Just want to be sure that both of these will parse and be reformatted without errors.

triblondon (approved):
Chatted to Mark and learned some Go!

PR #211, opened 2021-03-11
Replace TOML parser

Integralist:
The current toml parser is no longer maintained and we would also like to take advantage of more recent spec features. 

This will also help unblock implementing `manifest_version` into the fastly.toml file, as the new toml parser provides functionality that enables us to accommodate multiple data types (which we need to support for legacy reasons).

### DIFF
--- a/pkg/config/data.go
+++ b/pkg/config/data.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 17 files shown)


cratelyn (approved):
Woohoo!

pkg/config/data.go
```diff
@@ -121,8 +121,17 @@ type File struct {
...
 // Read the File and populate its fields from the filename on disk.
-func (f *File) Read(filename string) error {
-	_, err := toml.DecodeFile(filename, f)
+func (f *File) Read(fpath string) error {
+	// G304 (CWE-22): Potential file inclusion via variable.
+	// gosec flagged this:
+	// Disabling as we need to load the config.toml from the user's file system.
+	// This file is decoded into a predefined struct, any unrecognised fields are dropped.
+	/* #nosec */
+	bs, err := os.ReadFile(fpath)
```
phamann, line 130: Should this be `ioutil.ReadFile()` to be consistent with all the other changes int he PR?

phamann (approved):
👍🏻

doramatadora (approved):
This looks good to me! 🎉

PR #212, opened 2021-03-11
Replace deprecated ioutil functions with go 1.16.

Integralist:
Go 1.16 has deprecated the use of the `ioutil` package and moved the functions into the existing `io` and `os` packages.


triblondon (approved):

triblondon:
Looks like a positive change!

PR #214, opened 2021-03-16
The manifest_version should default to 1 if missing.

Integralist:
**Problem**: Originally if the manifest_version field was missing from the fastly.toml manifest file, then we would return an error. 
**Solution**: We should instead set a default value of `1`.

**Notes**: to issue an appropriate “warning” message, it required the manifest `File` object to have access to a consistent output stream. This meant everywhere we call `c.manifest.File.Read(manifest.Filename)` (which is in every command) I’ve had to precede it with a call to `c.manifest.File.SetOutput(c.Globals.Output)`.

### DIFF
--- a/pkg/compute/manifest/manifest.go
+++ b/pkg/compute/manifest/manifest.go
@@ -207,7 +207,7 @@ func (f *File) Read(fpath string) error {
 	f.exists = true
 
 	if f.ManifestVersion.int == 0 {
-		return errors.ErrMissingManifestVersion
+		f.ManifestVersion.int = 1
 	}
 
 	return nil
--- a/pkg/compute/manifest/manifest.go
+++ b/pkg/compute/manifest/manifest.go
@@ -1,11 +1,13 @@
 package manifest
 
 import (
+	"io"
 	"os"
 	"strconv"
 	"strings"
 
 	"github.com/fastly/cli/pkg/errors"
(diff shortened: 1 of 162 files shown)


kailan (approved):
Fantastic! Glad this got caught.

pkg/compute/manifest/manifest.go
```diff
@@ -207,7 +207,7 @@ func (f *File) Read(fpath string) error {
 	f.exists = true
 
 	if f.ManifestVersion.int == 0 {
-		return errors.ErrMissingManifestVersion
+		f.ManifestVersion.int = 1
```
phamann, line 210: This should be
```suggestion
		f.ManifestVersion.int = ManifestLatestVersion
```

phamann (changes requested):
Minor nit:

pkg/compute/manifest/manifest.go
Integralist: Ooh good shout!

kailan: I would disagree actually, if we were to release a v2 of the manifest then it should definitely have `manifest_version` in all new files. So this should always default to 1 if not defined.

phamann (approved):
:shipit:

Integralist:
Kailan, Patrick and myself have discussed the implementation internally and settled on a solution (which is now implemented).

PR #219, opened 2021-03-17
Refactor UserAgent.

Integralist:
**Problem**: we weren't setting the User-Agent HTTP request header when making requests to the dynamic config endpoint.
**Notes**: in order to support this change I needed to make a bunch of refactoring. Otherwise without the refactoring we would have had to duplicate the UserAgent logic due to 'cyclic import' issues.

Also, we already had a package named `version` which is used for a subsequent CLI subcommand. So I had to find an alternative name to encapsulate a bunch of version related types. So I settled on 'revision'. It might not be the best name so I'm open to suggestions.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,9 +11,9 @@ builds:
   - <<: &build_defaults
       main: ./cmd/fastly
       ldflags:
-        - -s -w -X "github.com/fastly/cli/pkg/version.AppVersion={{ .Version }}"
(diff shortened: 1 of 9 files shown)


kailan (commented):
User agent is reported as expected!

Screenshot 2021-03-17 at 12 09 58

kailan (approved):
Looks great! I like that the go-fastly user agent is appended so we can see the full dependency chain.

PR #220, opened 2021-03-17
Update the fastly.toml manifest if missing manifest_version.

Integralist:
**Problem**: the error message (see below) was being shown every time `fastly compute build` was executed. This happened because although the logic set a correct manifest_version value, that only happened in-memory and so the same error would happen each time the CLI was used (unless the user manually updated the file).

```
The fastly.toml was missing a `manifest_version` field. A default schema version of `1` will be used.
```

Additionally to this I discovered that the current implementation wouldn't write the manifest_version to the file correctly. It would be written as a separate block `[manifest_version]` rather than as a top-level field. So I've refactored the logic to fix that bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ lint:
 .PHONY: test
 test:
 	go test -race $(TESTARGS)
+
 .PHONY: build
 build:
 	go build ./cmd/fastly
--- a/pkg/compute/manifest/manifest.go
+++ b/pkg/compute/manifest/manifest.go
@@ -111,9 +111,7 @@ func (d *Data) Authors() ([]string, Source) {
 //
 // In the near future release the `Version` field will be removed and so there
 // will be less ambiguity around what this type refers to.
-type Version struct {
-	int
-}
+type Version int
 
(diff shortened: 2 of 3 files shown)


kailan (approved):
Got a weird dynamic config related error but I don't think it's anything to do with these changes.

```
kailan@KB-Fastly not-real-cli-config % ../../fastly/cli/fastly compute build

WARNING: The fastly.toml was missing a `manifest_version` field. A default schema version of `1` will be used.

✓ Initializing...
- Checking if Rust 1.49.0 is installed...
ERROR: Get "https://developer.fastly.com/api/internal/cli-config": context deadline exceeded.

This error may be caused by transient network issues. Please verify your network connection and try again.

kailan@KB-Fastly not-real-cli-config % ../../fastly/cli/fastly compute build
✓ Initializing...
✓ Building package using rust toolchain...
✓ Creating package archive...

SUCCESS: Built rust package cli-config (pkg/cli-config.tar.gz)
```

The manifest_version functionality works as intended. Test case looks good. Ship it 🚀

PR #221, opened 2021-03-17
Don't run "cargo update" before building rust app.

Integralist:
**Problem**: users have had issues with the fact that "cargo update" is called at the start of the `compute build` subcommand.
**Solution**: remove the `cargo update` now we have dynamic config as it is determining the fastly-sys constraints.
**Notes**: background context on the changes this PR reverts can be found here: https://github.com/fastly/cli/pull/179


Integralist:
Hi @acfoltzer can we get your review/approval on this PR as it directly affects your team. Thanks!

phamann (approved):
👍🏻 Approving from Go code perspective.

acfoltzer (approved):
Great to unwind this workaround, thank you!

PR #223, opened 2021-03-18
Loading remote config.toml should fail gracefully.

Integralist:
**Problem**: When a timeout occurred between the client and the service that serves the dynamic config used by the CLI, the logic would print the error and then shutdown the process (the error would also appear while the user's primary command was still executing, which of course would be very confusing).

**Solution**: Firstly, increase the timeout. This won't affect the user much because the config fetch happens asynchronously in the background and we already use the stale config for the purpose of the user's current command execution. Secondly, we shouldn't shutdown the process because of an error. We should tweak the error so the user understands their config is still stale but we'll attempt a re-fetch the next time around. Thirdly, we should defer the printing of the error until _after_ their command has finished executing.

**Notes**: 
Below is an example of one of the errors seen _before_ the changes implemented as part of this PR. Notice how we were displaying an error related to the config update right in the middle of the user's primary command (which would mislead a user into thinking the error was related to the previous output line "Checking if Rust 1.49.0 is installed").

```
- Checking if Rust 1.49.0 is installed...
ERROR: Get "https://developer.fastly.com/api/internal/cli-config": context deadline exceeded.
```

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
@@ -83,6 +84,8 @@ func main() {
 	waitForWrite := make(chan bool)
 	wait := false
 
+	var errLoadConfig error
+
 	// Validate if configuration is older than its TTL
 	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) {
 		if verboseOutput {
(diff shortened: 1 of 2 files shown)


thommahoney (approved):

cmd/fastly/main.go
```diff
@@ -98,11 +101,10 @@ Compatibility and versioning information for the Fastly CLI is being updated in
 			// configuration file to determine where to load the config from.
 			err := file.Load(file.CLI.RemoteConfig, httpClient)
 			if err != nil {
-				errors.RemediationError{
-					Inner:       err,
-					Remediation: errors.NetworkRemediation,
-				}.Print(os.Stderr)
-				os.Exit(1)
+				errLoadConfig = errors.RemediationError{
```
Integralist, line 104: @triblondon from a content perspective does this read OK to you?

triblondon (approved):
LGTM

PR #228, opened 2021-03-22
Set GOVERSION for goreleaser.

Integralist:
**Problem**: noticed the golang version used to build the released binary was not being included in the `fastly version` output.
**Notes**: see [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) for details.


phamann (approved):
👍🏻 Good catch!

doramatadora (approved):

PR #229, opened 2021-03-22
Extra error handling around loading remote configuration data.

Integralist:
**Problem**: we received a couple of reports that configuration data was missing and causing subsequently CLI flows to fail.
**Solution**: add extra error handling checks and retry specific operations.
**Notes**: Patrick and I ran through the code and were unable to pin-point an obvious source of the error and so we've made some 'best effort' modifications to hopefully help mitigate the problem.

### DIFF
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -11,7 +11,9 @@ import (
 func Stale(lastVersionCheck string, dur string) bool {
 	d, err := time.ParseDuration("-" + dur)
 	if err != nil {
-		return false
+		// if there' is no duration provided, then we should presume the loading of
+		// remote configuration failed and that we should retry that operation.
+		return true
 	}
 
 	if t, _ := time.Parse(time.RFC3339, lastVersionCheck); !t.Before(time.Now().Add(d)) {
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -11,7 +11,7 @@ import (
 func Stale(lastVersionCheck string, dur string) bool {
 	d, err := time.ParseDuration("-" + dur)
 	if err != nil {
-		// if there' is no duration provided, then we should presume the loading of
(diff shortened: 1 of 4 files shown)


pkg/check/check.go
```diff
@@ -11,7 +11,9 @@ import (
 func Stale(lastVersionCheck string, dur string) bool {
 	d, err := time.ParseDuration("-" + dur)
 	if err != nil {
-		return false
+		// if there is no duration provided, then we should presume the loading of
+		// remote configuration failed and that we should retry that operation.
+		return true
 	}
 
 	if t, _ := time.Parse(time.RFC3339, lastVersionCheck); !t.Before(time.Now().Add(d)) {
```
phamann, line 19: It also occurs to me that we're not checking the error here either. A scenario could be that we get this far but have a blank lastVersionCheck?

pkg/config/data.go
```diff
@@ -235,6 +235,10 @@ func (f *File) Load(configEndpoint string, httpClient api.HTTPClient) error {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("remote config: expected '200 OK', received '%s'", resp.Status)
```
phamann, line 239:
```suggestion
		return fmt.Errorf("fetching remote configuration: expected '200 OK', received '%s'", resp.Status)
```

phamann (approved):
Some small nits and a question, but still approving to unblock.

pkg/check/check.go
Integralist: We now won't get here because we have the additional check for:

```go
if file.CLI.LastChecked == "" {
```

Before we check stale.

triblondon (approved):

PR #233, opened 2021-03-24
Avoid setting fields that cause API to fail when given zero value.

Integralist:
**Problem**: The CLI was getting the backend and blanket assigning the response field values into a struct to be used to make an update to the backend. This was a problem because the zero value for some fields were being set in the update struct, which causes the API itself to fail because the zero values are invalid for the field.

### DIFF
--- a/pkg/backend/update.go
+++ b/pkg/backend/update.go
@@ -88,31 +88,55 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 
 	// Copy existing values from GET to UpdateBackendInput strcuture
 	input := &fastly.UpdateBackendInput{
-		ServiceID:           b.ServiceID,
-		ServiceVersion:      b.ServiceVersion,
-		Name:                b.Name,
-		NewName:             fastly.String(b.Name),
-		Address:             fastly.String(b.Address),
-		Port:                fastly.Uint(b.Port),
-		OverrideHost:        fastly.String(b.OverrideHost),
-		ConnectTimeout:      fastly.Uint(b.ConnectTimeout),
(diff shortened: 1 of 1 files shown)


Integralist:
@kailan @phamann have updated to include all fields now.

kailan (approved):

pkg/backend/update.go
```diff
@@ -86,37 +86,12 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
...
-		MinTLSVersion:       fastly.String(b.MinTLSVersion),
-		MaxTLSVersion:       fastly.String(b.MaxTLSVersion),
-		SSLCiphers:          b.SSLCiphers,
-	}
-
-	// Set values to existing ones to prevent accidental overwrite if empty.
+		ServiceID:      b.ServiceID,
+		ServiceVersion: b.ServiceVersion,
+		Name:           b.Name,
+	}
```
Integralist, line 93: @kailan just realised that my previous changes weren't necessary. All I had to do in fact was delete this chunk of code where it sets values into the update struct for otherwise optional fields.

The reason this works is because in go-fastly we have all the optional fields in an update struct as 'pointers' which means if the fields are never set here in the CLI where it makes the go-fastly call, then those fields won't get passed through to the upstream API.

The only time the fields get set is later on in this file when we use `WasSet` to indicate an _actual_ change in the field was provided by the user.

Integralist: I'm going to open a separate PR to ensure this pattern (which is what we use in terraform but never went back and implemented in the CLI) is applied throughout the CLI (I'm not going to do it here in this PR).

PR #234, opened 2021-03-24
Default to port 443 if UseSSL set.

Integralist:
**Problem**: if `--use-ssl` flag is provided but no port is set, then we should default to port 443 and also warn the user about this behaviour if the `--verbose` flag is provided.

**Notes**: you can't set a port number as part of the `address` field (e.g. `127.0.0.1:443`) as the API considers that invalid...

```bash
$ curl -s -H "Accept:application/vnd.api+json" -H "Fastly-Key: $FASTLY_API_KEY" -X PUT -d "address=127.0.0.1:443" https://api.fastly.com/service/4nE2aoohVrsds0OZGpSIP8/version/1/backend/httpbin | jq

{              
  "errors": [   
    {                     
      "detail": "Address 'address' is not a valid IPv4, IPv6 or hostname",
      "title": "Bad request"
    }                                                                                                                                                                                                                                         
  ]            
}   
```

### DIFF
--- a/pkg/backend/backend_test.go
+++ b/pkg/backend/backend_test.go
@@ -41,7 +41,7 @@ func TestBackendCreate(t *testing.T) {
 		{
 			args:       []string{"backend", "create", "--service-id", "123", "--version", "1", "--address", "127.0.0.1", "--name", "www.test.com", "--use-ssl", "--verbose"},
 			api:        mock.API{CreateBackendFn: createBackendOK},
-			wantOutput: "SSL to the backend was set but no port was specified, so default port :443 will be used",
+			wantOutput: "Use-ssl was set but no port was specified, so default port 443 will be used",
 		},
 	} {
 		t.Run(strings.Join(testcase.args, " "), func(t *testing.T) {
--- a/pkg/backend/backend_test.go
+++ b/pkg/backend/backend_test.go
@@ -38,6 +38,11 @@ func TestBackendCreate(t *testing.T) {
 			api:        mock.API{CreateBackendFn: createBackendOK},
 			wantOutput: "Created backend www.test.com (service 123 version 1)",
 		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/backend/create.go
```diff
@@ -67,6 +67,13 @@ func (c *CreateCommand) Exec(in io.Reader, out io.Writer) error {
 	c.Input.UseSSL = fastly.Compatibool(c.UseSSL)
 	c.Input.SSLCheckCert = fastly.Compatibool(c.SSLCheckCert)
 
+	if c.UseSSL && c.Input.Port == 0 {
+		if c.Globals.Flag.Verbose {
+			text.Warning(out, "SSL to the backend was set but no port was specified, so default port :443 will be used")
```
phamann, line 72: "SSL to the backend" sounds strange to me personally and I think its better to use the explicit name of the flag when you're saying it was "set". Ports don't have `:`.

```suggestion
			text.Warning(out, "Use-ssl was set but no port was specified, so default port 443 will be used")
```

phamann (approved):
Looks good to me other than minor nit around terminology in the verbose output. It would have been good to also have a test to actually assert 443 was set if no port was supplied, however that won't be possible until we refactor these command tests to mirror the patterns used in the logging endpoints.

pkg/backend/backend_test.go
```diff
@@ -38,6 +38,11 @@ func TestBackendCreate(t *testing.T) {
 			api:        mock.API{CreateBackendFn: createBackendOK},
 			wantOutput: "Created backend www.test.com (service 123 version 1)",
 		},
+		{
+			args:       []string{"backend", "create", "--service-id", "123", "--version", "1", "--address", "127.0.0.1", "--name", "www.test.com", "--use-ssl", "--verbose"},
+			api:        mock.API{CreateBackendFn: createBackendOK},
+			wantOutput: "SSL to the backend was set but no port was specified, so default port :443 will be used",
```
Integralist, line 44:
```suggestion
			wantOutput: "Use-ssl was set but no port was specified, so default port 443 will be used",
```

PR #235, opened 2021-03-24
Ensure all UPDATE operations don't set optional fields.

Integralist:
**Problem**: we get data and then set it back into an 'update' struct but this can cause errors to occur as the go-fastly client uses pointers now for its values and so there's no need to set 'optional' fields for UPDATE operations.

**Notes**: I've been manually validating the code for these changes work and I've discovered bugs. 

## Broken commands

Updating a domain doesn't work in master, but now works in this PR branch...

```bash
$ go run cmd/fastly/main.go domain update --token=$FASTLY_API_KEY --service-id=123 --version=1 --name=foo.com --new-name=bar.com

ERROR: missing required field 'Name'.

exit status 1
```

But in this PR branch it now works:

```bash
$ go run cmd/fastly/main.go domain update --token=$FASTLY_API_KEY --service-id=123 --version=1 --name=foo.com --new-name=bar.com
                                                                                                                       
SUCCESS: Updated domain bar.com (service 123 version 1)
```

Something else I noticed was that the domain update logic was calling GET first for no real benefit as we don't use any values from the GET API response to feed into the UPDATE API call. So I removed the redundant GET call.

## Inconsistent output

I've also made the output of some commands consistent. For example, when updating a dictionary item the output was:

```
$ go run cmd/fastly/main.go dictionaryitem update --token=$FASTLY_API_KEY --service-id=123 --dictionary-id=456 --key=bar --value=beepboop

Service ID: 123
Dictionary ID: 456
Item Key: bar
Item Value: beepboop
```

But now it's:

```
$ go run cmd/fastly/main.go dictionaryitem update --token=$FASTLY_API_KEY --service-id=123 --dictionary-id=456 --key=bar --value=baz

SUCCESS: Updated dictionary item (service 123)

Dictionary ID: 456
Item Key: bar
Item Value: baz
```

Which keeps it more inline with other commands that use the `SUCCESS` format with liberal spacing.

## Extra Comments

I don't like the fact the user has to provide a `--placement=none` flag when creating a logging endpoint for a WASM service. That feels like an implementation detail that shouldn't be their concern. Patrick noticed that the API does implement a check for this but there's some areas where it's missing, so it shouldn't be fixed in our clients.


pkg/domain/update.go
```diff
@@ -56,9 +56,7 @@ func (c *UpdateCommand) Exec(in io.Reader, out io.Writer) error {
 
 	c.updateInput.ServiceID = d.ServiceID
 	c.updateInput.ServiceVersion = d.ServiceVersion
-
-	// Set original value, and then afterwards check if we can use the flag value.
-	c.updateInput.Comment = &d.Comment
+	c.updateInput.Name = d.Name
```
Integralist, line 59: This missing field meant the CLI was broken in `master` for updating a domain.

Integralist: Also there's a difference between (for example) domains and dictionaries in that with the domain file we get the current domain and then update the domain, where as with dictionaries we don't bother getting the dictionary we just update it. This is relevant because it means there's an extra API step for domains where as the dictionary assigns the `--new-name` flag value directly into the update struct.

Integralist: It might be worth revisiting the domain implementation to not bother 'getting' the current domain because we no longer are reassigning any pre-existing values.

phamann (approved):
👍🏻 LGTM - as long as the tests are passing and you've done some manual testing I'm happy.

Integralist:
@phamann yup, I've manually tested...

- domain update
- dictionary item update
- s3 log endpoint creation
- updating a service

So I've not manually checked every single logging endpoint nor every command, but enough to be sure that the general approach is working.

PR #236, opened 2021-03-25
Make GitHub Versioner configurable.

Integralist:
**Problem**: We want to start integrating a new binary for local testing, which will be published as a public GitHub release, but the current Versioner type has hardcoded references to the CLI binary.

**Notes**: Because in future there will be two instances of a Versioner type in the code, I decided it best to rename all references to the variable `v` or `versioner` to be `cliVersioner` so it's clearer what versioner instance is being referred to.

The primary change is in `pkg/update/versioner.go`

### DIFF
--- a/pkg/update/versioner.go
+++ b/pkg/update/versioner.go
@@ -25,21 +25,25 @@ type Versioner interface {
 // GitHub is a versioner that uses GitHub releases.
 type GitHub struct {
 	client *github.Client
+	org    string
+	repo   string
 }
 
 // NewGitHub returns a usable GitHub versioner utilizing the provided token.
-func NewGitHub(ctx context.Context) *GitHub {
+func NewGitHub(ctx context.Context, org string, repo string) *GitHub {
 	var (
 		githubClient = github.NewClient(nil)
 	)
 	return &GitHub{
 		client: githubClient,
+		org:    org,
+		repo:   repo,
 	}
 }
 
 // LatestVersion implements the Versioner interface.
 func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
-	release, _, err := g.client.Repositories.GetLatestRelease(ctx, "fastly", "cli")
+	release, _, err := g.client.Repositories.GetLatestRelease(ctx, g.org, g.repo)
 	if err != nil {
 		return semver.Version{}, err
 	}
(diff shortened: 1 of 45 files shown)


pkg/update/versioner.go
```diff
@@ -135,7 +139,7 @@ func (g GitHub) getReleaseID(ctx context.Context, version semver.Version) (id in
 }
 
 func (g GitHub) getAssetID(assets []github.ReleaseAsset, version semver.Version) (id int64, err error) {
-	target := fmt.Sprintf("fastly_v%s_%s-%s.tar.gz", version, runtime.GOOS, runtime.GOARCH)
+	target := fmt.Sprintf("%s_v%s_%s-%s.tar.gz", g.org, version, runtime.GOOS, runtime.GOARCH)
```
triblondon, line 142: `fastly` here is possibly not a github org

phamann: Agreed, `fastly` in this context is the name of the binary program not the org. This is also applicable to many references above in the `LatestVersion()` method above, thus similar comments have been elided.

Furthermore, the filename pattern `fastly_v0.26.2_linux-amd64.tar.gz` is tightly coupled to the CLI release implementation and thus very opinionated and might not be the pattern that Viceroy or other future binaries use. Easy option is we ensure Viceroy uses this pattern or we will have to give some thought and refactoring as to how we generate the asset filename.

phamann (changes requested):
Per the inline discussion, I think we need to put more thought into how the filenames are constructed to be forwards compatible. 

Personal preference/opinion, but I also found the renaming of a local variable `s/versioner/cliVersioner` unnecessary and made the review harder than it needed to be.

pkg/update/versioner.go
Integralist: @cratelyn 👋🏻 would you be able to offer any insights here ^^ re: filename pattern. Thanks.

cratelyn: I think that this filename pattern sounds okay to me. Using semantic versioning would be fine, and putting a target/arch at the end is definitely worthwhile. No objections here!

Integralist:
@phamann @triblondon looks like Viceroy team are happy with the file format used here and so are we unblocked on this PR now?

phamann (approved):
👍🏻 LGTM

PR #238, opened 2021-03-27
pkg/logs: fix typo in error message

peterbourgon:
I noticed this while doing some testing of my own.

Also, I noticed that many of the error paths in pkg/logs.TailCommand.tail  `os.Exit(1)`. Unless I'm missing a subtlety here, they shouldn't be doing that — the tail method should be returning an error, which should  be handled properly by the caller, etc. May I make another PR which fixes this?

### DIFF
--- a/pkg/logs/tail.go
+++ b/pkg/logs/tail.go
@@ -17,15 +17,13 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/tomnomnom/linkheader"
-
 	"github.com/fastly/cli/pkg/common"
 	"github.com/fastly/cli/pkg/compute/manifest"
 	"github.com/fastly/cli/pkg/config"
 	"github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/cli/pkg/text"
-
 	"github.com/fastly/go-fastly/v3/fastly"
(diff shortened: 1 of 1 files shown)


pkg/logs/tail.go
```diff
@@ -186,15 +184,15 @@ func (c *TailCommand) tail(out io.Writer) {
 		}
 
 		req, err := http.NewRequest("GET", path, nil)
-		req.Header.Add("Fastly-Key", c.token)
 		if err != nil {
 			text.Error(out, "unable to create new request: %v", err)
 			os.Exit(1)
 		}
+		req.Header.Add("Fastly-Key", c.token)
```
peterbourgon, line 191: I also fixed this little bug.

Integralist:
👋🏻 @peterbourgon 

re: 

> I noticed that many of the error paths in pkg/logs.TailCommand.tail os.Exit(1). Unless I'm missing a subtlety here, they shouldn't be doing that — the tail method should be returning an error, which should be handled properly by the caller, etc. 

That would be great. This was very likely my mistake when reviewing @jaredmorrow's original Log Tail PR.

Integralist (approved):

PR #239, opened 2021-03-29
Retain RUSTFLAGS values from the environment

fgsch:
Append the debugging flags to any existing values so cargo can see both.

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -305,7 +305,12 @@ func (r Rust) Build(out io.Writer, verbose bool) error {
 	}
 	// Append debuginfo RUSTFLAGS to command environment to ensure DWARF debug
 	// information (such as, source mappings) are compiled into the binary.
-	os.Setenv("RUSTFLAGS", os.Getenv("RUSTFLAGS")+` -C debuginfo=2`)
+	rustflags := "-C debuginfo=2"
+	if val, ok := os.LookupEnv("RUSTFLAGS"); ok {
+		os.Setenv("RUSTFLAGS", fmt.Sprintf("%s %s", val, rustflags))
+	} else {
+		os.Setenv("RUSTFLAGS", rustflags)
+	}
 
 	// Execute the `cargo build` commands with the Wasm WASI target, release
 	// flags and env vars.
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -303,13 +303,13 @@ func (r Rust) Build(out io.Writer, verbose bool) error {
 	if verbose {
 		args = append(args, "--verbose")
 	}
-	// Add debuginfo RUSTFLAGS to command environment to ensure DWARF debug
+	// Append debuginfo RUSTFLAGS to command environment to ensure DWARF debug
 	// information (such as, source mappings) are compiled into the binary.
-	env := append(os.Environ(), `RUSTFLAGS=-C debuginfo=2`)
+	os.Setenv("RUSTFLAGS", os.Getenv("RUSTFLAGS")+` -C debuginfo=2`)
 
 	// Execute the `cargo build` commands with the Wasm WASI target, release
 	// flags and env vars.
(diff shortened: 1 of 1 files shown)


pkg/compute/rust.go
```diff
@@ -303,13 +303,13 @@ func (r Rust) Build(out io.Writer, verbose bool) error {
 	if verbose {
 		args = append(args, "--verbose")
 	}
-	// Add debuginfo RUSTFLAGS to command environment to ensure DWARF debug
+	// Append debuginfo RUSTFLAGS to command environment to ensure DWARF debug
 	// information (such as, source mappings) are compiled into the binary.
-	env := append(os.Environ(), `RUSTFLAGS=-C debuginfo=2`)
+	os.Setenv("RUSTFLAGS", os.Getenv("RUSTFLAGS")+` -C debuginfo=2`)
```
phamann, line 308: Although overly verbose, it would be better to check if the value exists using [`os.LookupEnv()`](https://golang.org/pkg/os/#LookupEnv) first to avoid the empty case which results in an empty space being prepended. Something like:

```suggestion
      rustflags := "-C debuginfo=2"
      if val, ok := os.LookupEnv("RUSTFLAGS"); ok {
          os.Setenv("RUSTFLAGS", val + rustflags)
      } else {
          os.Setenv("RUSTFLAGS", rustflags)
      }
```

Integralist, line 308: What happens when `RUSTFLAGS` isn't set and so we end up setting `RUSTFLAGS= -C debuginfo=2` with a space after the `=` character?

fgsch: Yeah, I thought about this but decided not to do it since it's too verbose for just a space that is a nop.

fgsch: Nothing.

fgsch: Let me know if this is your preference and I can change it but as I mentioned below, leading spaces (and trailing for the matter) are ignored.

Integralist: If it's not a breaking change then it's fine as it is but my concern was mentally mapping this to doing something like the following in a shell...

```bash
$ TEST_WITH_SPACE= -C debuginfo=2

-bash: -C: command not found
```

or whether it ends up more like the following where the space is removed by the shell:

```bash
$ TEST_WITH_SPACE=" -C debuginfo=2"                                                                                                                                                                                                                                              
$ echo $TEST_WITH_SPACE

-C debuginfo=2
```

fgsch: For completion:
```
$ export RUSTFLAGS="                         --cfg mock                  "
$ echo $RUSTFLAGS
                         --cfg mock
$ cargo build -v
   Compiling foo v0.1.0 (/private/tmp/foo)
     Running `rustc --crate-name foo --edition=2018 src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C metadata=57b6cdda4d28c9c9 --out-dir /private/tmp/foo/target/debug/deps -C incremental=/private/tmp/foo/target/debug/incremental -L dependency=/private/tmp/foo/target/debug/deps --cfg mock`
    Finished dev [unoptimized + debuginfo] target(s) in 0.99s

```

Integralist: I'm OK with this then, as it would seem the extra code @phamann suggested doesn't avoid anything that's not already handled by the shell. 

@phamann let me know if you have strong feelings about this.

fgsch: FWIW, this is not the shell but cargo.

Integralist (approved):

pkg/compute/rust.go
Integralist: @fgsch I've discussed with Patrick offline and we've agreed that we should probably go with the more verbose implementation. If you wouldn't mind addressing that and I'll get this PR approved/merged. Thanks.

fgsch: Updated. Let me know if I should squash them.

PR #240, opened 2021-03-29
Allow local binary to be renamed.

Integralist:
**Problem**: as we integrate a new binary for C@E local testing we don't want to confuse users by downloading the binary using its internal name.
**Solution**: support renaming the binary once it has been extracted from the release. Something like...

```go
org := "fastly"
repo := "viceroy"
binary := "viceroy"

viceroyVersioner = update.NewGitHub(context.Background(), org, repo, binary)
viceroyVersioner.RenameLocalBinary("fastly-localtesting")
```

### DIFF
--- a/pkg/update/versioner.go
+++ b/pkg/update/versioner.go
@@ -25,9 +25,10 @@ type Versioner interface {
 // GitHub is a versioner that uses GitHub releases.
 type GitHub struct {
 	client *github.Client
-	binary string
 	org    string
 	repo   string
+	binary string // name of compiled binary
+	local  string // name to use for binary once extracted
 }
 
 // NewGitHub returns a usable GitHub versioner utilizing the provided token.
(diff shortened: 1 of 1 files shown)


kailan:
My thoughts on this as discussed with @Integralist:

- If Viceroy is being pulled from GitHub releases, as this PR suggests, then it seems like the binary being named "viceroy" would not be an issue. A quick search for "fastly viceroy" will answer the question of what it is.
- If Viceroy is staying as a private repo, I would suggest renaming the binary on the update server rather than downloading it and then renaming locally.

Out of interest, where will viceroy be downloaded to? I assumed the config dir (`~/Library/Application Support/fastly` on macOS), in which case I would not worry about file naming. If they are confident enough to poke around in system dirs and delete a random binary, it should be obvious what has happened when `fastly compute serve` stops working.

cratelyn (approved):
This looks good to me! While I think I do agree with @kailan's point above, the ability to rename a local binary ourselves seems like a worthwhile addition.

Integralist:
Thanks @cratelyn 

@kailan I think under the circumstances I'm OK to merge this PR then. Just wanted to give you one last opportunity to holla if you were strongly opposed.

Issue #232, opened 2021-03-23
Add new subcommand to build and deploy a C@E service


fgsch:
Add a `publish` subcommand to `fastly compute` to build and deploy a service. 
This was mentioned recently in two separate conversations.

After this change, `fastly compute publish` always builds. 
If you use `fastly compute deploy` and have local changes, they will be ignored (existing behaviour).

Integralist:
@fgsch I'm presuming the flags that are defined for build/deploy will not be taken into consideration when running `fastly compute publish`, otherwise those flags would have to be aggregated into the `publish` subcommand for it to be able to pass the data down into the `build`/`deploy` functions.

fgsch:
My suggestion is to keep the flags for `deploy`, they seem to be useful for publishing as well, and perhaps also include `--include-source`. The rest of the `build` flags won't be missed, I believe.

fgsch:
My argument against merging both flags is that if you really need to use most (all?) of them, you are better off running build and deploy separately.  

Also it's easier to add flags when needed than removing flags nobody uses, so I'm sticking to my comments in https://github.com/fastly/cli/issues/232#issuecomment-811031932.

Integralist:
Agreed it's easier to add flags than remove those no-one needs but the fact is we don't necessarily know what people are going to want to use and as we're already suggesting implementing half of the available flags (which means we'd only be leaving off three of them) I just wonder if it's worth leaving the other 3 flags off. I'll have a think about it at any rate.

fgsch:
Done in b96f5152a3d08e6b75fe1b1a8eeffed0c5121cf7.

PR #242, opened 2021-03-31
Add `compute publish` subcommand.

Integralist:
**Problem**: Having to run two separate commands when the common behaviour is to need both can become tedious. 
**Solution**: aggregate the `compute build` and `compute deploy` behaviours into a single `compute publish`.

Fixes issue #232

### DIFF
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -303,6 +303,10 @@ COMMANDS
         --version=VERSION        Number of version to activate
     -p, --path=PATH              Path to package
 
+  compute publish
+    Runs "build" then "deploy" using fastly.toml manifest configuration
+
+
   compute update --service-id=SERVICE-ID --version=VERSION --path=PATH
     Update a package on a Fastly Compute@Edge service version
 
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -303,9 +303,16 @@ COMMANDS
         --version=VERSION        Number of version to activate
     -p, --path=PATH              Path to package
 
-  compute publish
-    Runs "build" then "deploy" using fastly.toml manifest configuration
+  compute publish [<flags>]
+    Composite of "build", then "deploy"
 
+    -s, --service-id=SERVICE-ID  Service ID
+        --version=VERSION        Number of version to activate
+    -p, --path=PATH              Path to package
+        --name=NAME              Package name
+        --language=LANGUAGE      Language type
+        --include-source         Include source code in built package
+        --force                  Skip verification steps and force build
 
   compute update --service-id=SERVICE-ID --version=VERSION --path=PATH
     Update a package on a Fastly Compute@Edge service version
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -304,7 +304,7 @@ COMMANDS
     -p, --path=PATH              Path to package
 
   compute publish [<flags>]
-    Composite of "build", then "deploy"
+    Build a Compute@Edge package and deploy it to a Fastly service
 
     -s, --service-id=SERVICE-ID  Service ID
         --version=VERSION        Number of version to activate
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -304,7 +304,7 @@ COMMANDS
     -p, --path=PATH              Path to package
 
   compute publish [<flags>]
-    Build a Compute@Edge package and deploy it to a Fastly service
+    Build and deploy a Compute@Edge package to a Fastly service
 
         --name=NAME              Package name
         --language=LANGUAGE      Language type
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -306,13 +306,13 @@ COMMANDS
   compute publish [<flags>]
     Build a Compute@Edge package and deploy it to a Fastly service
 
-    -s, --service-id=SERVICE-ID  Service ID
-        --version=VERSION        Number of version to activate
-    -p, --path=PATH              Path to package
         --name=NAME              Package name
         --language=LANGUAGE      Language type
         --include-source         Include source code in built package
         --force                  Skip verification steps and force build
+    -s, --service-id=SERVICE-ID  Service ID
+        --version=VERSION        Number of version to activate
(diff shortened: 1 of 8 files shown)


pkg/app/run_test.go
```diff
@@ -303,6 +303,10 @@ COMMANDS
         --version=VERSION        Number of version to activate
     -p, --path=PATH              Path to package
 
+  compute publish
+    Runs "build" then "deploy" using fastly.toml manifest configuration
+
+
```
Integralist, line 309: @fgsch @phamann this extra line break is annoying but it looks to be a consequence of not having any flags (TBD! see convo so far in [issue here](https://github.com/fastly/cli/issues/232#issuecomment-810946616)).

If we actually add flags, then this extra line break will go away.

Integralist:
@fgsch it would be great if you could pull down this branch and do some manual testing to be sure it works as you expect/need it to.

triblondon:
I'm very much in favour of a one-step command for this, because I agree there is rarely any use case for one without the other.

But I'm not sure if this should be `publish` or whether it should be an additional flag to `deploy`, such as `--build`.  It feels right to me that we help the user out by saying hey look I know you are saying to want to deploy this but you changed it and haven't rebuilt it since so we're just gonna take care of that for you.

triblondon:
(also I think it's _just_ possible that @phamann has thoughts)

fgsch:
@triblondon I think a warning makes sense for users of `deploy` when there are changes, but the whole idea of `publish`, at least to me, is that you don't care about those details.

pkg/compute/publish.go
```diff
@@ -0,0 +1,98 @@
...
+
+// NewPublishCommand returns a usable command registered under the parent.
+func NewPublishCommand(parent common.Registerer, globals *config.Data, build *BuildCommand, deploy *DeployCommand) *PublishCommand {
+	var c PublishCommand
+	c.Globals = globals
+	c.manifest.File.SetOutput(c.Globals.Output)
+	c.manifest.File.Read(manifest.Filename)
+	c.build = build
+	c.deploy = deploy
+	c.CmdClause = parent.Command("publish", "Composite of \"build\", then \"deploy\"")
```
triblondon, line 39: 'Composite' seems a bit of a complicated way to describe it, and requires the user to go look up two other things.  Maybe "Build a Compute@Edge package and deploy it to a Fastly service" ?

triblondon (approved):

pkg/compute/publish.go
```diff
@@ -0,0 +1,98 @@
...
+	// Deploy flags
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
+	c.CmdClause.Flag("version", "Number of version to activate").Action(c.version.Set).IntVar(&c.version.Value)
+	c.CmdClause.Flag("path", "Path to package").Short('p').Action(c.path.Set).StringVar(&c.path.Value)
+
+	// Build flags
+	c.CmdClause.Flag("name", "Package name").Action(c.name.Set).StringVar(&c.name.Value)
+	c.CmdClause.Flag("language", "Language type").Action(c.lang.Set).StringVar(&c.lang.Value)
+	c.CmdClause.Flag("include-source", "Include source code in built package").Action(c.includeSrc.Set).BoolVar(&c.includeSrc.Value)
+	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
```
phamann, line 50: By redefining the flags for each sub-command we risk them quickly diverging if a new flag is added to the real command. Couple of thoughts:

- Could we access the sub-commands flags programmatically via the `CmdClause` structure?  My memory seems to think this isn't possible as the flaggroup and argGroup fields are private, but worth a proper investigation. 
- If not, can we add a comment in the parent commands to ensure future developers update this list
- If not, can we write some kind of test which will fail if the commands diverge?

pkg/compute/publish.go
phamann: 👍🏻 to @triblondon's suggestion.

pkg/compute/publish.go
phamann: The ordering of these is also weird to me, as we build before we deploy.

pkg/compute/publish.go
Integralist: This is why I was happy GitHub chose you to review 🙂 because my description really sucked.

Integralist:
Thanks @triblondon for the approval. I'm going to wait for either @phamann or @fgsch to code review (and also for fgsch to do some manual testing to be sure the end user experience is satisfactory).

pkg/compute/publish.go
```diff
@@ -0,0 +1,98 @@
...
+	if c.includeSrc.WasSet {
+		c.build.IncludeSrc = c.includeSrc.Value
+	}
+	if c.force.WasSet {
+		c.build.Force = c.force.Value
+	}
+
+	err = c.build.Exec(in, out)
+	if err != nil {
+		return fmt.Errorf("error building package: %w", err)
```
phamann, line 79: I don't think the message here is necessary and the same with the deploy one on line 94, as the sub-commands already have their own relevant messaging.

phamann (commented):
LGTM other than comments around flags.

pkg/compute/publish.go
Integralist: Good point about the divergence issue.

A quick print of the `BuildCommand.CmdClause` (for example) shows the fields are indeed private:

```go
&{
  cmdMixin:{
    actionMixin:{actions:[] preActions:[]} 
    flagGroup:0xc000275b00 
    argGroup:0xc000210d80 
    cmdGroup:0xc000275bc0
  } 
  app:0xc00026f9e0 
  name:build 
  aliases:[] 
  help:Build a Compute@Edge package locally 
  isDefault:false 
  validator:<nil> 
  hidden:false 
  completionAlts:[] 
  optionalSubcommands:false
}
```

So I think I'll investigate the failing test scenario first, and if that's not reasonable to achieve then I'll comment in the code on the coupling.

Integralist: Added a test.

phamann (approved):
LGTM

PR #247, opened 2021-04-12
Warn on empty directory

Integralist:
**Problem**:  Not cleaning up hidden `.cargo` files has proven to cause issues for users who re-use directories.

### DIFF
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -90,6 +90,16 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	text.Output(out, "Press ^C at any time to quit.")
 	text.Break(out)
 
+	files, err := os.ReadDir(".")
+	if err != nil {
+		return err
+	}
+
+	if len(files) > 0 {
+		text.Warning(out, "The current directory isn't empty. This might prevent the CLI from building your project if there are conflicting language configuration files.")
+		text.Break(out)
+	}
+
 	var progress text.Progress
 	if c.Globals.Verbose() {
 		progress = text.NewVerboseProgress(out)
--- a/pkg/compute/init.go
+++ b/pkg/compute/init.go
@@ -90,14 +90,16 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	text.Output(out, "Press ^C at any time to quit.")
 	text.Break(out)
 
-	files, err := os.ReadDir(".")
+	cont, err := verifyDirectory(out, in)
 	if err != nil {
 		return err
 	}
 
-	if len(files) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/compute/init.go
```diff
@@ -90,6 +90,16 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	text.Output(out, "Press ^C at any time to quit.")
 	text.Break(out)
 
+	files, err := os.ReadDir(".")
+	if err != nil {
+		return err
+	}
+
+	if len(files) > 0 {
+		text.Warning(out, "The current directory isn't empty. This might prevent the CLI from building your project if there are conflicting language configuration files.")
```
cratelyn, line 99:
```suggestion
		text.Warning(out, "The current directory isn't empty. Are you sure you want to initialize a Compute@Edge project in $PWD?")
```

Maybe something like this? `$PWD` would probably require calling `os.Getwd` i think?

cratelyn (commented):
Thank you for adding this extra guardrail! I have a suggestion for the prompt, but that's a lightly-held position. This looks great to me :sparkles:

pkg/compute/init.go
Integralist: OK making changes locally and just wanted to share the prompt feedback...

**You don't want to continue...** (i.e. we short-circuit the execution)
Screenshot 2021-04-12 at 15 58 10

**You don't type what we're expecting...** (i.e. we short-circuit the execution)
Screenshot 2021-04-12 at 15 58 17

**You want to continue...** (i.e. we continue with the execution flow)
Screenshot 2021-04-12 at 15 58 25

cratelyn: I love it!! :heart_eyes_cat:

pkg/compute/init.go
```diff
@@ -491,6 +505,45 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+			return false, nil
+		}
+
+		if contl == "y" || contl == "yes" {
+			return true, nil
+		}
+
+		// NOTE: be defensive and default to short-circuiting the execution flow if
+		// the input is unrecognised.
+		return false, nil
```
Integralist, line 541: @phamann maybe we should not default to short-circuiting because in order to support that flow we have to introduce a `--force` flag so people can automate skipping the verification step and that is a breaking change for the `compute init` command 😬

Integralist: Discussed with Andrew and he feels this isn't a breaking change.

triblondon: I am very relaxed about this because if it does trip up any customers, they have multiple remediation options and arguably were just flagging an error state.

triblondon (approved):

PR #248, opened 2021-04-12
Check the rustup version

### LINKED ISSUE #123: Self-update fails on Arch Linux
**Version**
Applies to all versions

Please paste the output of `fastly version` here.

```
Fastly CLI version 0.10.0 (306abd1)
Built with go version go1.14.3 linux/amd64
```

**System Info**

```
$ uname -a
Linux hostname 5.6.15-arch1-1 #1 SMP PREEMPT Wed, 27 May 2020 23:42:26 +0000 x86_64 GNU/Linux
```

**What happened**
1. `wget https://github.com/fastly/cli/releases/download/v0.10.0/fastly_v0.10.0_linux-amd64.tar.gz`
2. `tar -xzf fastly_v0.10.0_linux-amd64.tar.gz -C /usr/local/bin`
3. Update using `fastly update`
```
$ sudo /usr/local/bin/fastly update
- Initializing...Current version: 0.10.0
Latest version: 0.12.0
✓ Initializing...
✓ Fetching latest release...
✗ Replacing binary...

ERROR: error moving latest binary in place: error creating destination file: open /usr/local/bin/fastly: text file busy.

```

### LINKED ISSUE #137: fastly compute build fails with "EOF"
**Version**
`v0.15.0`

Please paste the output of `fastly version` here.

```bash
Fastly CLI version 0.15.0 (05a2fe4)
Built with go version go1.14.4 linux/amd64
```

**What happened**

Please describe the command you ran, what you expected to happen, and what happened instead.

1. `fastly compute init --path fun`
2. Take all of the defaults
3. `fastly compute build`

```bash
✓ Initializing...
✗ Verifying local rust toolchain...

ERROR: error reading cargo metadata: EOF.

If you believe this error is the result of a bug, please file an issue: https://github.com/fastly/cli/issues/new?labels=bug&template=bug_repor
t.md
```

---

I expected it to pull down necessary dependencies.

fgsch:
Turns out, the toolchain section is only available in rustup 1.23.0 and higher.

Fixes #137.


pkg/compute/rust.go
```diff
@@ -123,15 +123,47 @@ func (r Rust) Verify(out io.Writer) error {
...
+	// 2) Check that the desired rustup version is installed
+	fmt.Fprintf(out, "Checking if rustup 1.23.0 or higher is installed...\n")
+
+	cmd := exec.Command("rustup", "--version")
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error executing rustup: %w", err)
+	}
+
+	reader := bufio.NewReader(strings.NewReader(string(stdoutStderr)))
```
Integralist, line 135:
```suggestion
	reader := bufio.NewReader(bytes.NewReader(stdoutStderr))
```

Integralist (changes requested):
The code looks fine to me but I think the constraint `">= 1.23.0"` should be part of the dynamic configuration (e.g. inside the `language.rust` section of https://developer.fastly.com/api/internal/cli-config).

This means updating https://github.com/fastly/cli/blob/master/pkg/config/data.go#L192-L204 to reflect a new `rustup_constraint` key set to `>= 1.23.0`, while also updating the internal dynamic configuration repo to publish an updated config with this value added.

> **NOTE**: the reason I suggest including the constraint syntax `>=` and not just the expected version `1.23.0` is because it provides more control to the team responsible for handling the rust dependencies (for example, the team had previously requested for `fastly_sys_constraint` to be set to ">= 0.3.3 < 0.5.0" rather than just a specific version). I appreciate that was to solve a different kind of coupling issue but I think I still the prefer the flexibility of this being inside the dynamic config if a change was necessary in future.

cc @acfoltzer and @cratelyn

fgsch:
> The code looks fine to me but I think the constraint ">= 1.23.0" should be part of the dynamic configuration (e.g. inside the language.rust section of https://developer.fastly.com/api/internal/cli-config).

I was not aware of this but I agree. I will update the PR. Thank you.

acfoltzer:
Looks good from the rustup semantics POV, thanks for the cc @Integralist!

Integralist (approved):
LGTM. Let's get the dynamic config PR merged first and check the config endpoint is serving the updated configuration.

PR #251, opened 2021-04-13
Add support for file_max_bytes configuration for azure logging endpoint

kellymclaughlin:
Support file_max_bytes configuration option for azure blob storage logging endpoint. Follows from https://github.com/fastly/go-fastly/pull/255.


Integralist (approved):
LGTM but can you resolve the go mod/sum conflicts and then I can get this merged.

PR #252, opened 2021-04-13
Implement `compute serve` subcommand.

Integralist:
Integrate a local testing environment for Compute@Edge.

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -15,16 +15,16 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/go/bin
-        key: ${{ runner.os }}-go-bin-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-gobin-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-go-bin-
+          ${{ runner.os }}-gobin-
     - name: Restore cache
       uses: actions/cache@v1
       with:
         path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-gomod-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-go-mod-
+          ${{ runner.os }}-gomod-
     - name: Install dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: make dependencies
@@ -67,9 +67,9 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/go/pkg/mod
-        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-gomod-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-go-mod-
+          ${{ runner.os }}-gomod-
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -15,16 +15,16 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/go/bin
-        key: ${{ runner.os }}-gobin-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-go-bin-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-gobin-
+          ${{ runner.os }}-go-bin-
     - name: Restore cache
       uses: actions/cache@v1
       with:
         path: ~/go/pkg/mod
-        key: ${{ runner.os }}-gomod-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-gomod-
+          ${{ runner.os }}-go-mod-
     - name: Install dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: make dependencies
@@ -67,9 +67,9 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/go/pkg/mod
-        key: ${{ runner.os }}-gomod-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-gomod-
+          ${{ runner.os }}-go-mod-
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -34,12 +34,13 @@ func main() {
 	// the "real" versions that pull e.g. actual commandline arguments, the
 	// user's real environment, etc.
 	var (
-		args                    = os.Args[1:]
-		clientFactory           = app.FastlyAPIClient
-		httpClient              = http.DefaultClient
-		in            io.Reader = os.Stdin
-		out           io.Writer = sync.NewWriter(os.Stdout)
-		versionerCLI            = update.NewGitHub(context.Background(), "fastly", "cli", "fastly")
+		args                       = os.Args[1:]
+		clientFactory              = app.FastlyAPIClient
+		httpClient                 = http.DefaultClient
+		in               io.Reader = os.Stdin
+		out              io.Writer = sync.NewWriter(os.Stdout)
+		versionerCLI               = update.NewGitHub(context.Background(), "fastly", "cli", "fastly")
+		versionerViceroy           = update.NewGitHub(context.Background(), "fastly", "viceroy", "viceroy")
 	)
 
 	// We have to manually handle the inclusion of the verbose flag here because
@@ -125,15 +126,18 @@ Compatibility and versioning information for the Fastly CLI is being updated in 
 
 	// Main is basically just a shim to call Run, so we do that here.
 	opts := app.RunOpts{
-		APIClient:    clientFactory,
-		Args:         args,
-		ConfigFile:   file,
-		ConfigPath:   config.FilePath,
-		Env:          env,
-		HTTPClient:   httpClient,
-		Stdin:        in,
-		Stdout:       out,
-		VersionerCLI: versionerCLI,
+		APIClient:  clientFactory,
+		Args:       args,
+		ConfigFile: file,
+		ConfigPath: config.FilePath,
+		Env:        env,
+		HTTPClient: httpClient,
+		Stdin:      in,
+		Stdout:     out,
+		Versioners: app.Versioners{
+			CLI:     versionerCLI,
+			Viceroy: versionerViceroy,
+		},
 	}
 	if err := app.Run(opts); err != nil {
 		fsterrors.Deduce(err).Print(os.Stderr)
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	_ "embed"
 
-	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -34,13 +33,21 @@ func main() {
 	// the "real" versions that pull e.g. actual commandline arguments, the
 	// user's real environment, etc.
 	var (
-		args                       = os.Args[1:]
-		clientFactory              = app.FastlyAPIClient
-		httpClient                 = http.DefaultClient
-		in               io.Reader = os.Stdin
-		out              io.Writer = sync.NewWriter(os.Stdout)
-		versionerCLI               = update.NewGitHub(context.Background(), "fastly", "cli", "fastly")
-		versionerViceroy           = update.NewGitHub(context.Background(), "fastly", "viceroy", "viceroy")
+		args                    = os.Args[1:]
+		clientFactory           = app.FastlyAPIClient
+		httpClient              = http.DefaultClient
+		in            io.Reader = os.Stdin
+		out           io.Writer = sync.NewWriter(os.Stdout)
+		versionerCLI            = update.NewGitHub(update.GitHubOpts{
+			Org:    "fastly",
+			Repo:   "cli",
+			Binary: "fastly",
+		})
+		versionerViceroy = update.NewGitHub(update.GitHubOpts{
+			Org:    "fastly",
+			Repo:   "viceroy",
+			Binary: "viceroy",
+		})
 	)
 
 	// We have to manually handle the inclusion of the verbose flag here because
--- a/pkg/compute/manifest/manifest.go
+++ b/pkg/compute/manifest/manifest.go
@@ -191,19 +191,19 @@ func (v *Version) UnmarshalText(text []byte) error {
 // File represents all of the configuration parameters in the fastly.toml
 // manifest file schema.
 type File struct {
-	ManifestVersion Version  `toml:"manifest_version"`
-	Name            string   `toml:"name"`
-	Description     string   `toml:"description"`
-	Authors         []string `toml:"authors"`
-	Language        string   `toml:"language"`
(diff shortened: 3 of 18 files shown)


Integralist:
@phamann this is still very much WIP but just letting you know this exists.

pkg/compute/serve.go
```diff
@@ -0,0 +1,278 @@
...
+
+	stdoutStderr, err := cmd.CombinedOutput()
+
+	if err != nil {
+		// We presume an error executing `viceroy --version` means it isn't installed.
+		//
+		// NOTE: we can't use exec.LookPath("viceroy") because PATH is unreliable
+		// across OS platforms but also we actually install viceroy in the same
+		// location as the application configuration, which means it wouldn't be
+		// found looking up by the PATH env var.
```
Integralist, line 133: @phamann The following thoughts all feel like I'm worrying about stuff that's too low-level, hence the implementation I've used doesn't try to concern itself. But that said, there are two things I'm wondering about this section:

1. I don't know if my comment "PATH is unreliable across OS platforms" is _**100%**_ accurate.

2. re: "it wouldn't be found looking up by the PATH env var" should we consider adding the config directory to the PATH (again, I'm not sure how cross platform `os.Setenv()` is**)?

> ** looking at the [`syscall.Setenv`](https://golang.org/src/syscall/env_unix.go?s=1956:1992#L86) implementation it has an OS check for Plan9 but no other OS which might not necessarily mean an issue for Windows.

pkg/compute/manifest/manifest.go
```diff
@@ -189,11 +189,20 @@ type File struct {
 	Authors         []string `toml:"authors"`
 	Language        string   `toml:"language"`
 	ServiceID       string   `toml:"service_id"`
+	Testing         Testing  `toml:"testing"`
```
Integralist: 5. We update the manifest `File` data structure to reflect there is now a new `[testing]` section (as per the [`fastly.toml`](https://developer.fastly.com/reference/fastly-toml/) specification).

pkg/compute/manifest/manifest.go
```diff
@@ -183,17 +183,26 @@ func (v *Version) UnmarshalText(text []byte) error {
...
-	Authors         []string `toml:"authors"`
-	Language        string   `toml:"language"`
-	ServiceID       string   `toml:"service_id"`
+	ManifestVersion Version     `toml:"manifest_version"`
+	Name            string      `toml:"name"`
+	Description     string      `toml:"description"`
+	Authors         []string    `toml:"authors"`
+	Language        string      `toml:"language"`
+	ServiceID       string      `toml:"service_id"`
+	LocalServer     LocalServer `toml:"local_server"`
```
Integralist, line 192: 5. We update the manifest `File` data structure to reflect there is now a new `[local_server]` section (as per the [`fastly.toml`](https://developer.fastly.com/reference/fastly-toml/) specification).

> **NOTE**: This doesn't reflect the current state of the world (the section is defined as `[testing]` in the design document). @triblondon and I have been pushing to rename 'testing', and after a discussion we both agreed on 'local_server' but we still need to get agreement from the wider development team.

Integralist:
@phamann I've now marked this PR as "ready for review" although strictly speaking it's still blocked by external factors (i.e. a public repo to pull viceroy from). I'd be happy to jump on a Zoom call to run through the relevant changes (as there's lots of files that are updated, which was caused by a necessity to update the function signature for `app.Run()`, and so I can help focus in on the _actual_ changes that need reviewing).

/cc @triblondon

pkg/compute/manifest/manifest.go
Integralist: /cc @acfoltzer @cratelyn

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+	c.CmdClause = parent.Command("serve", "Build and run a Compute@Edge package locally")
+
+	c.manifest.File.SetOutput(c.Globals.Output)
+	c.manifest.File.Read(manifest.Filename)
+
+	// Build flags
+	c.CmdClause.Flag("name", "Package name").Action(c.name.Set).StringVar(&c.name.Value)
+	c.CmdClause.Flag("language", "Language type").Action(c.lang.Set).StringVar(&c.lang.Value)
+	c.CmdClause.Flag("include-source", "Include source code in built package").Action(c.includeSrc.Set).BoolVar(&c.includeSrc.Value)
+	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
```
phamann, line 59: Do we also want to include a flag to skip building?

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// Viceroy flags
+	c.CmdClause.Flag("env", "The environment to use when selecting backend definitions (e.g. stage)").Action(c.env.Set).StringVar(&c.env.Value)
```
phamann, line 62: While currently the only object we support is backends we will eventually allow to configure more, so consider re-wording this description. Something like:
```suggestion
	c.CmdClause.Flag("env", "The environment configuration to use (e.g. stage)").Action(c.env.Set).StringVar(&c.env.Value)
```

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+	viceroyError := errors.RemediationError{
+		Inner:       fmt.Errorf("viceroy version not found"),
+		Remediation: errors.BugRemediation,
+	}
+
+	// version output has the expected format: `viceroy 0.1.0`
+	segs := strings.Split(version, " ")
+
+	if len(segs) < 2 {
+		return viceroyError
```
phamann, line 209:
```suggestion
		progress.Fail()
		return viceroyError
```

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	installedViceroyVersion = segs[1]
+
+	if installedViceroyVersion == "" {
+		return viceroyError
```
phamann, line 215:
```suggestion
		progress.Fail()
		return viceroyError
```

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+		syscall.SIGTERM,
+	}
+
+	signal.Notify(sig, signals...)
+
+	if env != "" {
+		env = "." + env
+	}
+	manifest := fmt.Sprintf("fastly%s.toml", env)
+	args := []string{"bin/main.wasm", "-C", manifest}
```
phamann, line 268: Not an issue now, but something for us to consider, we're making a lot of assumptions here that `fastly compute serve` was run in the root directory of a project. Is this ok? Do we want users to provide a path to teh manifest and binary via flags so it can be ran elsewhere, this may be useful in CI scenarios.

pkg/compute/serve.go
```diff
@@ -0,0 +1,287 @@
...
+
+	go func(sig chan os.Signal, cmd *common.StreamingExec) {
+		<-sig
+		signal.Stop(sig)
+
+		err := cmd.Signal(os.Kill)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}(sig, cmd)
```
phamann, line 280: This seems like good default behaviour and an oversight in the original implementation of `StreamingExec`, maybe we should consider moving this logic inside `streamingexec.Exec()` so all instances of it in the CLI benefit?

phamann (changes requested):
Looking good. Couple of comments.

pkg/compute/serve.go
Integralist: @phamann we could do, but isn't the build step (with build caching) quick enough to mean it's not a concern?

Integralist:
The macOS CI check has been failing since renaming from `master` to `main`. Tried changing the cache key entry in the workflow file (see last commit). @phamann do you have any bright ideas on why that might now be failing (no error details other than "This check was cancelled") 🤷🏻 

Otherwise I've addressed your feedback.

pkg/common/exec.go
```diff
@@ -60,3 +91,14 @@ func (s StreamingExec) Exec() error {
 
 	return nil
 }
+
+// Signal enables spawned subprocess to accept given signal.
+func (s StreamingExec) Signal(signal os.Signal) error {
```
Integralist, line 96: @Integralist 

**TODO**: this doesn't need to be publicly exported anymore.

pkg/compute/manifest/manifest.go
Integralist: UPDATE: the change from `testing` to `local_server` is now agreed.

cmd/fastly/main.go
```diff
@@ -30,13 +30,14 @@ func main() {
...
-		httpClient               = http.DefaultClient
-		cliVersioner             = update.NewGitHub(context.Background(), "fastly", "cli", "fastly")
-		in             io.Reader = os.Stdin
-		out            io.Writer = common.NewSyncWriter(os.Stdout)
+		args                       = os.Args[1:]
+		configFilePath             = config.FilePath // write-only for `fastly configure`
+		clientFactory              = app.FastlyAPIClient
+		httpClient                 = http.DefaultClient
+		cliVersioner               = update.NewGitHub(context.Background(), "fastly", "cli", "fastly")
+		viceroyVersioner           = update.NewGitHub(context.Background(), "fastly", "viceroy", "viceroy")
```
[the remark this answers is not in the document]
doramatadora: I really appreciated all these comments when reviewing, thank you 🙌🏼

doramatadora (approved):

.github/workflows/pr_test.yml
```diff
@@ -15,16 +15,16 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/go/bin
-        key: ${{ runner.os }}-go-bin-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-gobin-${{ hashFiles('**/go.sum') }}
```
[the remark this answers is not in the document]
Integralist: I think I can revert these 👍🏻

Integralist:
We might need to watch out for errors going to stdout rather than stderr https://github.com/fastly/Viceroy/pull/171#discussion_r656566379

pkg/compute/serve.go
phamann: I am keen to still explore adding a flag to skip building. While I agree that build+serve is the most common pattern users will want locally, however, I think in automated scenario's (i.e. CI builds) you'll most likely separate build, test, and serve (integration tests) stages and double building will slow down builds. Its also useful if you share binaries with colleagues or integrate into other scripts etc.

pkg/compute/serve.go
```diff
@@ -0,0 +1,309 @@
...
+	if env != "" {
+		env = "." + env
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	manifest := filepath.Join(wd, fmt.Sprintf("fastly%s.toml", env))
```
phamann, line 286: Am I interpreting this correctly that a `fastly compute serve --env test` will require a file named `fastlytest.toml` noting the lack of any separator between fastly and the stage? I think it would be a nicer ux if we add one:

```suggestion
	manifest := filepath.Join(wd, fmt.Sprintf("fastly-%s.toml", env))
```

pkg/compute/serve.go
```diff
@@ -0,0 +1,309 @@
...
+	c.manifest.File.SetOutput(c.Globals.Output)
+	c.manifest.File.Read(manifest.Filename)
+
+	// Build flags
+	c.CmdClause.Flag("name", "Package name").Action(c.name.Set).StringVar(&c.name.Value)
+	c.CmdClause.Flag("language", "Language type").Action(c.lang.Set).StringVar(&c.lang.Value)
+	c.CmdClause.Flag("include-source", "Include source code in built package").Action(c.includeSrc.Set).BoolVar(&c.includeSrc.Value)
+	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
+
+	// Viceroy flags
```
phamann, line 60: To complement my request for a skip building flag above, I think it would be nice if we also allowed users to provide the binary file to run as an optional flag `--file` or `--bin`. This could default to `bin/main.wasm`.

```suggestion
	// Viceroy flags
        c.CmdClause.Flag("file", "The Wasm file to run").Action(c.file.Set).StringVar(&c.file.Value)	
```

pkg/compute/serve.go
```diff
@@ -0,0 +1,309 @@
...
+		env = "." + env
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	manifest := filepath.Join(wd, fmt.Sprintf("fastly%s.toml", env))
+	args := []string{"bin/main.wasm", "-C", manifest}
```
phamann, line 287: Per my comment above in this file, I don't like how we have hard coded `bin/main.wasm` here and would prefer if this was configurable.

pkg/update/check.go
```diff
@@ -24,6 +25,9 @@ func Check(ctx context.Context, currentVersion string, cliVersioner Versioner) (
 		return current, latest, false, fmt.Errorf("error fetching latest version: %w", err)
 	}
 
+	asset := fmt.Sprintf("%s_v%s_%s-%s.tar.gz", cliVersioner.Binary(), latest, runtime.GOOS, runtime.GOARCH)
```
phamann, line 28: This template string is used in a couple of places across the code base, maybe consider moving it to a package level constant.

phamann (changes requested):
Looking good, a couple of requests around exposing more flags for a nicer UX.

pkg/compute/serve.go
Integralist: @phamann there's a `.` inbetween so `fastly.test.toml` look a couple lines up from that one I think.

pkg/compute/serve.go
Integralist: OK that makes sense. I'll take a look into it.

pkg/compute/serve.go
Integralist: OK. I'll take a look.

pkg/compute/serve.go
Integralist: 👍🏻

pkg/update/check.go
Integralist: 👍🏻

pkg/compute/serve.go
Integralist: Lines 277-279

phamann: Apologies, my fault, thats perfect 👌🏻

pkg/compute/serve.go
```diff
@@ -0,0 +1,313 @@
...
+	c.viceroyVersioner = viceroyVersioner
+
+	c.Globals = globals
+	c.CmdClause = parent.Command("serve", "Build and run a Compute@Edge package locally")
+
+	c.manifest.File.SetOutput(c.Globals.Output)
+	c.manifest.File.Read(manifest.Filename)
+
+	c.CmdClause.Flag("env", "The environment configuration to use (e.g. stage)").Action(c.env.Set).StringVar(&c.env.Value)
+	c.CmdClause.Flag("file", "The Wasm file to run").StringVar(&c.file)
```
phamann, line 54: Am interested in knowing why you didn't set the default via the flag itself. I personally prefer this method, as the default value then gets documented rather than being magic. 

```suggestion
	c.CmdClause.Flag("file", "The Wasm file to run").Default("bin/main.wasm").StringVar(&c.file)
```

phamann (approved):
👍🏻  Looks good to me, other than minor non-blocking comment.

pkg/compute/serve.go
Integralist: I didn't know that was a thing with Kingpin. So I'll sort that 👍🏻

dkegel-fastly:
An --addr option would let me use this instead of directly running Viceroy.
I can't push a branch, so here's a git patch that implements one:
[0001-serve-accept-optional-addr-flag-and-pass-along-to-Vi.patch.txt](https://github.com/fastly/cli/files/6812622/0001-serve-accept-optional-addr-flag-and-pass-along-to-Vi.patch.txt)

kailan (approved):
All seems to work as intended 👍 🚀

PR #253, opened 2021-04-13
Support usage of IAM role in S3 and Kinesis logging endpoints

kellymclaughlin:
Add the ability to specify an IAM role for the S3 and Kinesis logging endpoints. Follows from https://github.com/fastly/go-fastly/pull/269.


pkg/logging/s3/create.go
```diff
@@ -86,8 +93,37 @@ func (c *CreateCommand) createInput() (*fastly.CreateS3Input, error) {
 	input.ServiceVersion = c.Version
 	input.Name = c.EndpointName
 	input.BucketName = c.BucketName
-	input.AccessKey = c.AccessKey
-	input.SecretKey = c.SecretKey
+
+	// The following block checks for invalid permutations of the ways in
```
kellymclaughlin, line 97: I was looking for something akin to the [ArgGroup](https://github.com/clap-rs/clap/blob/v2.33.1/examples/14_groups.rs) feature in the `clap` library for rust, but I didn't see anything similar present in `kingpin` so I just implemented the checks manually here. If this ever became a more common pattern we could try to add it into the kingpin library or just our fork of it since it's unmaintained because it'd be a lot nicer to have it handled inside the library versus a bunch of ad hoc enforcement in random places in our code.

Integralist: Agreed, if this becomes more common then we can think about implementing it within the Fastly Kingpin fork.

Integralist (approved):

PR #257, opened 2021-04-15
Check when binary was updated.

Integralist:
**Problem**: When we have an updated [configuration](https://developer.fastly.com/api/internal/cli-config) the CLI code is updated so it can parse the new configuration schema, but if the CLI is run (this initial run will cause the config data to be cached for 5mins) before the binary is updated to a new version (which includes code that expects new config fields) _and_ if the **new** CLI version is then subsequently run before the currently cached config TTL has expired, then a user will find the new CLI code will try to read a struct field that doesn't have a value because the underlying configuration data isn't updated.

See also: https://github.com/fastly/cli/issues/258

**Solution**: The customer waits 5 minutes (or however long the TTL is set to in the config) and then tries the CLI command again and the configuration will be considered stale and subsequently updated before the subcommand is executed.

**Better Solution**: Implement a check alongside the 'stale config check' that validates if the CLI binary was recently modified (i.e. updated) and if so we'll force the fetching of the configuration.

**Notes**: this will require an update to the [configuration endpoint](https://developer.fastly.com/api/internal/cli-config) to include a new `version`.


pkg/check/check.go
```diff
@@ -11,13 +12,44 @@ import (
...
+	if err != nil {
+		return true
+	}
+
+	fi, err := os.Stat(bin)
+	if err != nil {
+		return true
+	}
+
+	if t := fi.ModTime(); !t.Before(time.Now().Add(d)) {
```
fgsch, line 50: What about `t.After(time.Now().Add(d))` instead and return `true`?

Integralist:
The implementation currently doesn't work. Discussed with Kailan and am addressing a different way.

Integralist:
@phamann have reworked the implementation and @kailan has validated it locally too.

cmd/fastly/main.go
```diff
@@ -78,6 +79,17 @@ func main() {
...
+	if file.CLI.Version != revision.SemVer(revision.AppVersion) {
+		err := file.Load(config.RemoteEndpoint, httpClient)
+		if err != nil {
+			errors.RemediationError{
+				Inner:       err,
+				Remediation: errors.NetworkRemediation,
+			}.Print(os.Stderr)
+			os.Exit(1)
+		}
+	}
```
fgsch, line 91: This is almost identical to the code above. 
Perhaps move the code above out and add `err != nil` to this condition?

pkg/update/root.go
```diff
@@ -75,6 +79,13 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 		}
 	}
 
+	c.Globals.File.CLI.Version = latest.String()
+
+	// Write the file data to disk.
+	if err := c.Globals.File.Write(c.configFilePath); err != nil {
+		return fmt.Errorf("error saving config file: %w", err)
+	}
+
```
fgsch, line 88: IIUC by the time we reach this code we've already fetched the latest config. 
Is this code to avoid refetching it after the update?

cmd/fastly/main.go
Integralist: I wasn't sure I understood what you were suggesting, so I'm going to just clarify the two distinct flows that are defined, and which I don't believe can be combined (if that's what you were asking)...

**The first scenario** we're addressing begins on L55, where it attempts to read configuration from the file system and if there is no existing local config, then it will fetch the config from the remote endpoint.

**The second scenario is**: if there is an existing config, then the error block on L56 won't be executed. So on L82 we check whether the CLI has been updated recently, and if it has, then there is the _possibility_ that the existing config data we have is outdated (e.g. maybe the data is missing a new field which the updated CLI code is going to soon run and expect to exist). So in that case we'll refetch the config from the remote endpoint. 

> **NOTE**: The second load of remote configuration (scenario 2) won't be executed if the first load of remote configuration (scenario 1) did happen because when we fetch the remote configuration we update the `file.CLI.Version` and so the conditional on L82 won't pass.

fgsch: My suggestion was to write the code combining the two file.Load() together:

```
	if err != nil {
		if verboseOutput {
			if err == config.ErrLegacyConfig {
				text.Output(out, `
					Found your local configuration file (required to use the CLI) was using a legacy format.
					File is being upgraded now.
				`)
			} else {
				text.Output(out, `
					Unable to locate a local configuration file (required to use the CLI).
					File is being created now.
				`)
			}
			text.Break(out)
		}
	}

	if err != nil || file.CLI.Version != revision.SemVer(revision.AppVersion) {
		err := file.Load(config.RemoteEndpoint, httpClient)
		if err != nil {
			errors.RemediationError{
				Inner:       err,
				Remediation: errors.NetworkRemediation,
			}.Print(os.Stderr)
			os.Exit(1)
		}
	}
```

pkg/update/root.go
Integralist: Correct! The scenario here is this:

- User runs `fastly update` and the config is (potentially) fetched from the remote endpoint. 
  - The local config is now cached for 5 minutes.
  - We write the updated version back to the local configuration file.
- User runs a new command, config won't be re-fetched because it's TTL hasn't expired.
  - But now in `main.go` we have a check for `file.CLI.Version` to see if it matches what's baked into the binary (i.e. `revision.AppVersion`), so we see the config has the same version as the running binary and so we avoid re-fetching the remote configuration.

cmd/fastly/main.go
Integralist: Ah I see, OK cool 👍🏻

pkg/update/root.go
fgsch: Thank you.

Personally I'd prefer to remove this code and fetch it again for better encapsulation but ymmv.

fgsch (approved):
Lightly tested but LGTM.

kailan (approved):
Code reviewed and tested locally – works as expected on configurations from 0.26.x. No more requests than needed are made to the config service. Ship it 🚀

PR #262, opened 2021-04-20
GitHub Packages support via goreleaser.

Integralist:
**Problem**: We we would like to better support Dockerised versions of the CLI.
**Solution**: Implement goreleaser API for generating docker images.

**Useful references**:

- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
- https://goreleaser.com/customization/docker/
- https://goreleaser.com/customization/docker_manifest/
- https://docs.github.com/en/packages/guides/configuring-docker-for-use-with-github-packages
- https://docs.github.com/en/packages/guides/about-github-container-registry#visibility-and-access-permissions-for-container-images
- https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list
- https://docs.docker.com/engine/reference/commandline/manifest/
- https://www.docker.com/blog/multi-platform-docker-builds/


Dockerfile-assemblyscript
```diff
@@ -0,0 +1,16 @@
+FROM node:latest
+
+WORKDIR /tmp
+
+# NOTE: I have presumed we'd pass these env vars via the Docker -e flag.
```
Integralist, line 5: Actually I'm not sure what I was thinking here, because it doesn't look like goreleaser let's me configure any form of docker related flags and so goreleaser is just handling things internally as far as building a docker image is concerned.

.goreleaser.yml
```diff
@@ -93,3 +93,47 @@ snapshot:
...
+  dockerfile: Dockerfile-rust
+- <<: *docker_opts
+  <<: *amd64
+  dockerfile: Dockerfile-rust
+- <<: *docker_opts
+  <<: *arm64
+  dockerfile: Dockerfile-rust
+  build_flag_templates:
+    - "--platform=linux/arm/v8"
+docker_manifests:
```
Integralist, line 134: Do we **_need_** to use `docker manifest`?

We have to use `dockers` any way to build the images, and it seems like it's possible to define multiple dockerfile files to be used when building these images, so I'm not sure what the value is in generating a manifest as well? 

Especially considering the trade-off is that 'manifests' is an experimental feature that, in some cases (depending on the version of Docker CLI that is running), needs to be explicitly enabled.

Hmm, thinking about it, I guess using this manifest feature would be useful from a 'workflow' perspective. 

**So for example**: I would imagine without a manifest the workflow, for people who want to use the dockerised version of the CLI, would be to visit the GitHub release page where (again I don't actually know how these docker images will be shown) the user will find the relevant image link to the `ghcr.io` endpoint. 

If they had the manifest, then an automated CI flow they could just use the Docker CLI to request a manifest for the CLI and then parse out the image version they're interested in.

andersjanmyr:
LGTM Thanks!

Dockerfile-assemblyscript
```diff
@@ -0,0 +1,20 @@
+FROM node:15.14.0
+LABEL maintainer="Developer Relations Team <support@fastly.com>"
```
phamann, line 2: We tend to prefer using the open-source software distribution list for all public communication, i.e. 
```suggestion
LABEL maintainer="Fastly OSS <oss@fastly.com>"
```

Dockerfile-assemblyscript
```diff
@@ -0,0 +1,20 @@
...
+
+WORKDIR /tmp
+RUN npm install --global npm@latest
+RUN npm install --save-dev assemblyscript
```
phamann, line 6: Why have you done this inside a `/tmp` directory? The `npm install --save-dev assemblyscript` here will likely fail as there isn't an initialised npm package (i.e. a package.json file) to `--save-dev` persist too? A couple of options:
- as the users source code will likely have a valid package.json with `assemblyscript` as a dependency, you could skip it all together
- alternatively you should `npm install --global assemblyscript`, however, you then have the issue of the global version of `asc` being different to that which is defined in the users local package and depending on how their $PATH is configured the global one may be chosen first and lead to confusion.

Dockerfile-assemblyscript
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update && apt-get install -y curl jq
+RUN export FASTLY_CLI_VERSION=$(curl --silent https://api.github.com/repos/fastly/cli/releases/latest | jq -r .tag_name | cut -d 'v' -f 2) \
+  GOARCH=$(dpkg --print-architecture) \
+  && curl -vL "https://github.com/fastly/cli/releases/download/v${FASTLY_CLI_VERSION}/fastly_v${FASTLY_CLI_VERSION}_linux-$GOARCH.tar.gz" -o fastly.tar.gz
+RUN tar -xzf fastly.tar.gz
+RUN mv fastly /usr/bin
```
phamann, line 13: Instead of doing this inside a `tmp` directory you could extract the binary directly to `/usr/bin`
```suggestion
RUN tar -xzf fastly.tar.gz --directory /usr/bin
```

Dockerfile-assemblyscript
```diff
@@ -0,0 +1,20 @@
...
+LABEL maintainer="Developer Relations Team <support@fastly.com>"
+
+WORKDIR /tmp
+RUN npm install --global npm@latest
+RUN npm install --save-dev assemblyscript
+
+RUN apt-get update && apt-get install -y curl jq
+RUN export FASTLY_CLI_VERSION=$(curl --silent https://api.github.com/repos/fastly/cli/releases/latest | jq -r .tag_name | cut -d 'v' -f 2) \
+  GOARCH=$(dpkg --print-architecture) \
+  && curl -vL "https://github.com/fastly/cli/releases/download/v${FASTLY_CLI_VERSION}/fastly_v${FASTLY_CLI_VERSION}_linux-$GOARCH.tar.gz" -o fastly.tar.gz
```
phamann, line 11: It would be nice to see some integrity verification here via the shasum, we export the `fastly_vx.x.x_SHA256SUMS` file alongside every release for this purpose.

Dockerfile-assemblyscript
phamann: This also applies to the Rust image below.

Dockerfile-assemblyscript
phamann: This also applies to the Rsut image below.

phamann (changes requested):
Looking good. A couple of comments around best-practice.

Dockerfile-assemblyscript
Integralist: @phamann 

re: 

> Why have you done this inside a /tmp directory? The npm install --save-dev assemblyscript here will likely fail as there isn't an initialised npm package (i.e. a package.json file) to --save-dev persist too? 

This doesn't appear to be the case:

```
root@792eb4caa11c:/tmp# pwd
/tmp
root@792eb4caa11c:/tmp# ls -la
total 5260
drwxrwxrwt 1 root root    4096 Apr 21 14:24 .
drwxr-xr-x 1 root root    4096 Apr 23 14:27 ..
-rw-r--r-- 1 root root 5354637 Apr 21 14:24 fastly.tar.gz
drwxr-xr-x 6 root root    4096 Apr 21 14:24 node_modules
-rw-r--r-- 1 root root    2391 Apr 21 14:24 package-lock.json
-rw-r--r-- 1 root root      64 Apr 21 14:24 package.json
drwxr-xr-x 3 root root    4096 Apr 10 07:39 v8-compile-cache-0
root@792eb4caa11c:/tmp# cat package.json
{
  "devDependencies": {
    "assemblyscript": "^0.18.27"
  }
}
```

I found a `package.json` was generated automatically for me when one wasn't present. 

With that in mind, I realise now that this is redundant because as you also commented...

> as the users source code will likely have a valid package.json with assemblyscript as a dependency, you could skip it all together

This would ultimately mean that the user would need to provide their own AssemblyScript (e.g. via a mounted `node_modules` directory), which is fine if that's the expectation, and additionally it would make the container more flexible as it wouldn't be restricting the version of AssemblyScript a customer wants to utilise.

What about `npm install --global npm@latest`? Is there any concern in having this built into the image? Could this cause conflicts between the user's local code (that they mount into the running container). Maybe we shouldn't do this and just rely on the `FROM node:15.14.0` providing whatever version of `npm` comes with that version of node?

Dockerfile-assemblyscript
```diff
@@ -0,0 +1,19 @@
...
+
+RUN apt-get update && apt-get install -y curl jq
+RUN export FASTLY_CLI_VERSION=$(curl --silent https://api.github.com/repos/fastly/cli/releases/latest | jq -r .tag_name | cut -d 'v' -f 2) \
+  GOARCH=$(dpkg --print-architecture) \
+  && curl -vL "https://github.com/fastly/cli/releases/download/v${FASTLY_CLI_VERSION}/fastly_v${FASTLY_CLI_VERSION}_linux-$GOARCH.tar.gz" -o fastly.tar.gz \
+  && curl -vL "https://github.com/fastly/cli/releases/download/v${FASTLY_CLI_VERSION}/fastly_v${FASTLY_CLI_VERSION}_SHA256SUMS" -o sha256sums \
+  && dlsha=$(shasum -a 256 fastly.tar.gz | cut -d " " -f 1) && expected=$(cat sha256sums | awk -v pat="$dlsha" '$0~pat' | cut -d " " -f 1) \
+  && if [[ "$dlsha" != "$expected" ]]; then echo "shasums don't match" && exit 1; fi
+
+RUN tar -xzf fastly.tar.gz --directory /usr/bin && rm fastly.tar.gz
```
Integralist, line 12: @phamann I've addressed your feedback (see above updates). Let me know what you think.

andersjanmyr:
Any plans when this will be merged?

Integralist:
Hi @andersjanmyr 

We are in the process of scheduling an internal security review (regarding getting access to GitHub Packages for the CLI project). 

Unfortunately I don't have an ETA on when this will be completed but I'm actively engaging with the relevant team.

phamann (approved):
👍🏻

Integralist:
Hi @andersjanmyr thanks for your patience.

I'm going to merge this PR today. 

In the next few days I hope to cut a new CLI release so we should hopefully see these Dockerised CLI releases on https://ghcr.io

PR #264, opened 2021-04-23
Replace master with main.

Integralist:


kailan (approved):
As long as the build passes, looks good! We will need to update devhub ingestion to use the new branch name. I'll do this now.

Integralist:
Thanks @kailan yeah the CI is having issues with `macos-latest` no matter what I try so I'm going to attempt switching from latest to `macos-11.0` which is more recent than 'latest' but I want to see if the CI builds with it.

Issue #250, opened 2021-04-13
Init a compute project without creating a service and objects


phamann:
Currently, `fastly compute init` requires a valid API token and will provision a new Compute@Edge service for the user on their account and add any objects (backends, domains etc) that they provide during the wizard flow to the initial service version.

While this is normally the desired outcome there are situations when a user wants to scaffold out a project for local without touching the API:

- They are intending on using Terraform to create and manage the compute service
- They only want local development / experimentation before creating any object in their companies production account
- They have no or limited network (aeroplane etc)

Therefore, it would be nice if we could provide a mechanism to scaffold out a compute project locally without creating a service. Either via a flag or changing the default behaviour to not touch the API. 

This will become a more common pattern when we have a local development server an no need to publish to a service to test code.

PR #266, opened 2021-04-26
Move service creation to `fastly compute deploy`.

Integralist:
Currently when a user runs fastly `compute init`, we create a service on the Fastly platform, but that service is not needed until the user wants to _deploy_ their project.  We should move the service creation to the `deploy` subcommand.

Fixes https://github.com/fastly/cli/issues/250


pkg/compute/deploy.go
```diff
@@ -41,137 +56,437 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
 	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
 	c.CmdClause.Flag("version", "Number of version to activate").Action(c.Version.Set).IntVar(&c.Version.Value)
 	c.CmdClause.Flag("path", "Path to package").Short('p').StringVar(&c.Path)
+	c.CmdClause.Flag("domain", "The name of the domain associated to the package").StringVar(&c.Domain)
+	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.Backend)
+	c.CmdClause.Flag("backend-port", "A port number for the package backend").UintVar(&c.BackendPort)
```
Integralist, line 61: 5. Here are the flags we're migrating from `init` to `deploy`.

pkg/compute/deploy.go
```diff
@@ -41,137 +56,437 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
...
 func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
-	var progress text.Progress
-	if c.Globals.Verbose() {
-		progress = text.NewVerboseProgress(out)
-	} else {
-		progress = text.NewQuietProgress(out)
+	// Exit early if no token configured.
+	_, s := c.Globals.Token()
+	if s == config.SourceUndefined {
+		return errors.ErrNoToken
```
Integralist, line 71: 6. This logic was originally only in the `init` function but is moved from there to here as we no longer make API requests in `init`. I've also added this token check to the pre-existing `publish` subcommand.

pkg/compute/deploy.go
```diff
@@ -41,137 +56,437 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
...
-	if c.Path == "" {
-		progress.Step("Reading package manifest...")
+	undoStack := common.NewUndoStack()
+	defer func() { undoStack.RunIfError(out, err) }()
 
-		name, source := c.manifest.Name()
-		if source == manifest.SourceUndefined {
-			return fmt.Errorf("error reading package manifest")
+	serviceID, source := c.manifest.ServiceID()
+	if source != manifest.SourceUndefined {
```
Integralist, line 98: 7. If the `fastly.toml` manifest already has a Service ID set, then we'll go ahead and clone the latest version.

pkg/compute/deploy.go
```diff
@@ -41,137 +56,437 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
...
+		version, err = serviceVersion(serviceID, c.Globals.Client, c.Version, progress)
+		if err != nil {
+			return err
+		}
+	} else {
```
Integralist, line 103: 8. Otherwise we'll skip to here and create the service, domain and backend.

pkg/compute/deploy.go
```diff
@@ -41,137 +56,437 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
...
 	progress.Step("Validating package...")
+	if err := validate(path); err != nil {
+		return err
+	}
+
+	cont, err := pkgCompare(c.Globals.Client, serviceID, version.Number, path, progress, out)
+	if err != nil {
+		return err
+	}
+	if !cont {
```
Integralist, line 170: 9. We'll short-circuit the `deploy` subcommand if we discover the actual wasm binary hasn't been changed since the last time it was uploaded.

pkg/compute/init.go
```diff
@@ -4,33 +4,25 @@ import (
...
 	"github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/cli/pkg/filesystem"
 	"github.com/fastly/cli/pkg/text"
-	"github.com/fastly/go-fastly/v3/fastly"
+
 	"gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
-const (
```
Integralist, line 29: 10. The changes in `init` are actually much simpler as we're primarily deleting code, and the only additional thing I have done is refactored the remaining logic into separate functions.

pkg/compute/publish.go
```diff
@@ -17,8 +17,11 @@ type PublishCommand struct {
 	deploy   *DeployCommand
 
 	// Deploy fields
-	path    common.OptionalString
-	version common.OptionalInt
+	path        common.OptionalString
```
Integralist, line 20: 11. Now `init` and `deploy` have been updated we need to ensure `publish` is kept in-sync as it proxies through to the `deploy` subcommand.

pkg/service/delete.go
```diff
@@ -37,6 +39,25 @@ func (c *DeleteCommand) Exec(in io.Reader, out io.Writer) error {
 	}
 	c.Input.ID = serviceID
 
+	if c.force {
```
Integralist, line 42: 12. I discovered no one has actually tried deleting a service using the CLI as they would discover that it would fail with a message to say the activated service can't be deleted. So I've copied the technique we use in the Terraform provider which is to accept a flag to force deletion. All this does is first _disable_ the service before attempting to delete it.

Integralist (commented):
Here is a brief rundown of the changes in this PR.

pkg/service/delete.go
phamann: Nit: this belongs in its own PR.

phamann (approved):
:+1: LGTM - As far as I can tell it seems to just shuffle the code around and do some refactoring and no new functionality is added. 

I'm also going to quickly test locally to test the new workflow manually.

pkg/app/run_test.go
```diff
@@ -158,6 +158,7 @@ SUBCOMMANDS
     Delete a Fastly service
 
     -s, --service-id=SERVICE-ID  Service ID
+    -f, --force                  Force the deletion of the service
```
Integralist, line 161:
```suggestion
    -f, --force                  Force deletion of an active service.
```

pkg/compute/deploy.go
```diff
@@ -41,13 +56,47 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
...
+	}
+
+	var (
+		domain, backend string
+		backendPort     uint
+	)
+
+	serviceID, sidSrc := c.manifest.ServiceID()
+	if sidSrc == manifest.SourceUndefined {
+		text.Output(out, "You don't currently have a service defined. This utility will walk you through defining the resources necessary to create a Compute@Edge service.")
```
Integralist, line 81:
```suggestion
		text.Output(out, "There is no Fastly service associated with this package. To connect to an existing service add the Service ID to the fastly.toml file or otherwise press enter to create one now.")
```

pkg/compute/init.go
```diff
@@ -178,331 +134,317 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+		return fmt.Errorf("error initializing package: %w", err)
+	}
+
+	progress.Done()
+
+	text.Break(out)
+
+	text.Description(out, fmt.Sprintf("Initialized package %s to", text.Bold(m.Name)), abspath)
+	text.Description(out, "To compile the package, run", "fastly compute build")
+	text.Description(out, "To deploy the package, run", "fastly compute deploy")
```
Integralist, line 198:
```
	fastly compute publish
```

kailan (approved):
All works as expected for me! My only suggestion would be for `compute publish` to run the interactive service creation prompt _before_ building the package, but it's probably not worth the effort.

PR #267, opened 2021-04-27
Add isBool to command flags

kailan:
This is to support https://github.com/kailan/figly/commit/0e9c2f2a8513e7f1e909fea6f6bc739200be071d

By adding this flag, we can determine if args are booleans or take a value based on the output of `fastly help --format json`.


Integralist (approved):

doramatadora (approved):
🚀  I'm happy with this addition

PR #268, opened 2021-05-07
Clean-out service_id from manifest when deleting a service.

Integralist:
**Problem**: When you delete a service, the `service_id` in the manifest persists. Which (depending on what subcommands you execute next) could cause an unexpected error.
**Solution**: When deleting a service, be sure to clean-out the Service ID value from the manifest.

### DIFF
--- a/pkg/service/delete.go
+++ b/pkg/service/delete.go
@@ -66,6 +66,16 @@ func (c *DeleteCommand) Exec(in io.Reader, out io.Writer) error {
 		}
 	}
 
+	if err := c.manifest.File.Read(manifest.Filename); err != nil {
+		return fmt.Errorf("error reading package manifest: %w", err)
+	}
+
+	c.manifest.File.ServiceID = ""
+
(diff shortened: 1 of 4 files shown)


phamann:
This seems to be based off of #266 and thus makes it hard to review. Is this intentional?

pkg/service/delete.go
```diff
@@ -66,6 +66,16 @@ func (c *DeleteCommand) Exec(in io.Reader, out io.Writer) error {
 		}
 	}
 
+	if err := c.manifest.File.Read(manifest.Filename); err != nil {
+		return fmt.Errorf("error reading package manifest: %w", err)
```
kailan, line 70: Would this not fail when running delete outside of a project directory? I would expect the command to still work for normal VCL services or via passing in a flag

Integralist: Ooh good point. Let me have a think about the implementation 👍🏻

kailan (changes requested):

kailan (approved):
Fantastic!

PR #272, opened 2021-05-10
Fix --force flag not being respected.

Integralist:
**Problem**: an error message about the directory not being empty would be displayed regardless of whether you passed `--force`.


kailan (approved):
Looks good 👍

PR #275, opened 2021-05-12
Update RELEASE process

Integralist:
**Problem**: The changelog generator is constantly non-deterministic/flaky and will always suggest moving prior entries around even when they shouldn't be moved. This means it's a very tedious process to git patch apply only select changes.
**Solution**: Remove the generator. It's very trivial for me to look at the commits in `main` since the last release was cut and to manually generate my own changelog entry.
**Notes**: I intend on making a similar change to go-fastly. I've also made the RELEASE document in this repo consistent with our Terraform RELEASE document (and again in go-fastly I'll look to make that one consistent).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,6 @@
 project_name: fastly
 release:
+  draft: true
   prerelease: auto
   extra_files:
     - glob: "dist/usage.json"
@@ -92,4 +93,4 @@ checksum:
 snapshot:
   name_template: "{{ .Tag }}-next"
 changelog:
-  sort: asc
+  skip: true
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,6 +15,6 @@
 
 Step 4. we utilize [semantic versioning](https://semver.org/) and only include relevant/significant changes within the CHANGELOG.
 
-Step 7. causes a [github action](https://github.com/fastly/cli/blob/main/.github/workflows/tag_release.yml) to be triggered which produces a 'draft' release.
+Step 7. triggers a [github action](https://github.com/fastly/cli/blob/main/.github/workflows/tag_release.yml) that produces a 'draft' release.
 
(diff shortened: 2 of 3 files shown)


RELEASE.md
```diff
@@ -1,31 +1,20 @@
...
-1. Push the new tag:
-    * `git push origin vx.x.x`
-1. Go to GitHub and check that the release was successful:
-    * Check the release CI job status via the [Actions](https://github.com/fastly/cli/actions?query=workflow%3ARelease) tab
-    * Check the release exists with valid assets and changelog: https://github.com/fastly/cli/releases
-1. Announce release internally via Slack
-1. Celebrate :tada:
+Step 4. we utilize [semantic versioning](https://semver.org/) and only include relevant/significant changes within the CHANGELOG.
+
+Step 7. causes a [github action](https://github.com/fastly/cli/blob/main/.github/workflows/tag_release.yml) to be triggered which produces a 'draft' release.
```
doramatadora, line 18:
```suggestion
Step 7. triggers a [github action](https://github.com/fastly/cli/blob/main/.github/workflows/tag_release.yml) that produces a 'draft' release.
```

RELEASE.md
```diff
@@ -1,31 +1,20 @@
...
+
+Step 10. `#api-clients`, `#ecp-languages`.
```
doramatadora, line 20:
```suggestion
Step 10. Fastly make internal announcements in the Slack channels: `#api-clients`, `#ecp-languages`.
```

.goreleaser.yml
```diff
@@ -1,5 +1,6 @@
 project_name: fastly
 release:
+  draft: true
```
Integralist, line 3: This will cause the release to be a draft, and thus allow me to manually copy/paste the relevant CHANGELOG content into it before publishing. This is how the Terraform repo works.

doramatadora (approved):
A couple of suggestions, otherwise 👍🏼

PR #278, opened 2021-05-12
compute/deploy: clear Service ID from manifest when service is deleted

Integralist:
**Problem**: during the `deploy` logic flow, if there is an error at any point, we clean up by deleting the service/domain/backend. But we don't clear the Service ID from the manifest, which means on the next run of `deploy` a different flow branch is executed and the user can get into an unexpected state where they can't activate their service as the service defined in the manifest is being used and that service has since been deleted.


kailan (approved):
Very elegant 👍

PR #279, opened 2021-05-12
Validate package before prompting inside deploy flow.

Integralist:
**Problem**: Currently when running `compute deploy` an error, caused by not first running `build`, will cause the service/domain/backend resources to be deleted and thus the user has to re-enter the information when they re-run deploy. That is a frustrating experience.
**Solution**: Move the package validation check to the top of the deploy flow so we can short-circuit any prompts.


kailan (approved):
Looks good 👍

doramatadora (approved):
LGTM!

PR #281, opened 2021-05-13
pkg/compute/rust: update messaging for 'rustup self update'

### LINKED ISSUE #269: Update rustup error to cover package managers
If `rustup` was installed via a package manager (e.g. homebrew) and needs to be updated, the provided instructions will not work:

```
error: self-update is disabled for this build of rustup
error: you should probably use your system package manager to update rustup
```

We might want to reword the instructions to cover this case.

Reported from @gaastonsr via #137.

Integralist:
Fixes #269

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -155,7 +155,7 @@ func (r Rust) Verify(out io.Writer) error {
 	if !rustupConstraint.Check(rustupVersion) {
 		pre := "To fix this error, run the following command"
 		cmd := text.Bold("rustup self update")
-		alt := fmt.Sprintf("%s If you installed rustup using a package manager, refer to its documentation to update it.", text.Bold("INFO:"))
+		alt := fmt.Sprintf("%s If you installed rustup using a package manager, you may need to follow your package manager's documentation to update the rustup package.", text.Bold("INFO:"))
 
 		return errors.RemediationError{
 			Inner:       fmt.Errorf("rustup constraint not met: %s", r.config.File.Language.Rust.RustupConstraint),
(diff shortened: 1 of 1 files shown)


pkg/compute/rust.go
```diff
@@ -153,9 +153,13 @@ func (r Rust) Verify(out io.Writer) error {
 		return fmt.Errorf("error parsing rustup constraint: %w", err)
 	}
 	if !rustupConstraint.Check(rustupVersion) {
+		pre := "To fix this error, run the following command"
+		cmd := text.Bold("rustup self update")
+		alt := fmt.Sprintf("%s If you installed rustup using a package manager, refer to its documentation to update it.", text.Bold("INFO:"))
```
kailan, line 158: perhaps "package manager such as Homebrew"?

Integralist: I had hoped to avoid calling out specific package managers (although admittedly the issue we're addressing has been reported because of Homebrew).

fgsch: What are we trying to accomplish by naming a specific package manager?

Personally I'd prefer if we don't, as this might be the case with any of them and/or might change in the future without our knowledge.

triblondon: I'm less worried about that and more about implying that we want the user to update their package manager 😂 .  I might say don't mention homebrew unless we can be sure that we're only doing that on MacOS platforms.

```suggestion
		alt := fmt.Sprintf("%s If you installed rustup using a package manager, you may need to follow your package manager's documentation to update the rustup package.", text.Bold("INFO:"))
```

kailan (approved):

PR #282, opened 2021-05-13
pkg/compute/pack: implement pack subcommand

Integralist:
**Problem**: We want users to experiment with the C@E platform, which means being able to deploy their own WASM binaries compiled using languages not officially supported (yet).

**Solution**: Implement `fastly compute pack`, which will accept a `--path` to a WASM binary and will scaffold the required structure so that `fastly compute deploy` will know what to do. 

**Notes**: I've also updated `fastly compute init` to provide a new language option of "other" as this means we can piggyback off the existing logic to build a compatible fastly.toml manifest (which `fastly compute deploy` will require).


pkg/compute/init.go
```diff
@@ -185,17 +190,26 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
 
 	progress.Done()
 
 	text.Break(out)
 
 	text.Description(out, fmt.Sprintf("Initialized package %s to", text.Bold(m.Name)), abspath)
-	text.Description(out, "To publish the package (build and deploy), run", "fastly compute publish")
+
+	if language.Name == "other" {
+		text.Description(out, "To package a WASM binary for deployment, run", "fastly compute pack")
```
kailan, line 207: how about "a pre-compiled WASM binary"? Concerned that this implies the build will output a wasm binary.

pkg/compute/compute_integration_test.go
```diff
@@ -1678,6 +1775,34 @@ func makeDeployEnvironment(t *testing.T, manifestContent string) (rootdir string
...
+	t.Helper()
+
+	rootdir, err := os.MkdirTemp("", "fastly-pack-*")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := os.MkdirAll(rootdir, 0700); err != nil {
+		t.Fatal(err)
+	}
```
fgsch, line 1788: This is not needed - `os.MkdirTemp` will create the directory with 0700.

Integralist: Nice, thanks! Will remove.

Integralist:
@phamann I appreciate you're busy so are you happy for someone else to review this? e.g. @fgsch

pkg/compute/init.go
```diff
@@ -118,6 +117,10 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			StarterKits: c.Globals.File.StarterKits.AssemblyScript,
 			Toolchain:   NewAssemblyScript(),
 		}),
+		NewLanguage(&LanguageOptions{
+			Name:        "other",
+			DisplayName: "Other (pre-compiled WASM binary)",
```
triblondon, line 122:
```suggestion
			DisplayName: "Other ('bring your own' build process)",
```

triblondon:
"Initialized package" => "Initialized project" ?

fgsch:
Would be possible to use a smaller binary?

pkg/compute/init.go
Integralist: @fgsch I see you've disagreed with @triblondon's suggestion. Do you have an alternative for consideration (or can you elaborate on why the suggestion isn't ideal)?

fgsch: I like the original "pre-compiled WASM binary" text.

Integralist:
> Would be possible to use a smaller binary?

I had used the binary compiled using https://github.com/fastly/compute-starter-kit-rust-default as I was not familiar with generating a WASM binary myself. 

But I've managed to replace the `main.wasm` binary with a plain text file as a _real_ binary isn't necessary for validating the tests.

pkg/app/run_test.go
```diff
@@ -322,6 +322,11 @@ COMMANDS
         --backend-port=BACKEND-PORT
                                  A port number for the package backend
 
+  compute pack --path=PATH
+    Package a pre-compiled WASM binary for a Fastly Compute@Edge service
+
+    -p, --path=PATH  Path to a custom pre-compiled WASM binary
```
fgsch, line 328: Not a blocker but I'd probably zap "custom" from the text above.

pkg/compute/pack.go
```diff
@@ -0,0 +1,100 @@
...
+// NewPackCommand returns a usable command registered under the parent.
+func NewPackCommand(parent common.Registerer, globals *config.Data) *PackCommand {
+	var c PackCommand
+	c.Globals = globals
+
+	c.manifest.File.SetOutput(c.Globals.Output)
+	c.manifest.File.Read(manifest.Filename)
+
+	c.CmdClause = parent.Command("pack", "Package a pre-compiled WASM binary for a Fastly Compute@Edge service")
+	c.CmdClause.Flag("path", "Path to a custom pre-compiled WASM binary").Short('p').Required().StringVar(&c.path)
```
fgsch, line 33: Same here wrt "custom". Not a blocker.

pkg/compute/rust.go
```diff
@@ -80,7 +80,7 @@ func (m *CargoMetadata) Read() error {
 	return nil
 }
 
-// Rust is an implements Toolchain for the Rust language.
+// Rust is implements a Toolchain for the Rust language.
```
fgsch, line 83: Mind zapping the "is" as well? Not a blocker.

fgsch (approved):
Tested "compute pack" deploying an existing service using an unsupported language, a supported language (rust) and also initialising a project using "Other".

Code looks and works as advertised ✔️

fgsch:
Left a few non blocking comments but looks great to me. 

Thank you for working on this!

pkg/compute/init.go
```diff
@@ -185,19 +190,28 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	if language.Name == "other" {
+		text.Description(out, "To package a pre-compiled WASM binary for deployment, run", "fastly compute pack")
+		text.Description(out, "To deploy the package, run", "fastly compute deploy")
+	} else {
+		text.Description(out, "To publish the package (build and deploy), run", "fastly compute publish")
+	}
+
 	text.Description(out, "To learn about deploying Compute@Edge projects using third-party orchestration tools, visit", "https://developer.fastly.com/learning/integrations/orchestration/")
-	text.Success(out, "Initialized package %s", text.Bold(m.Name))
+	text.Success(out, "Initialized project %s", text.Bold(m.Name))
```
phamann, line 214: Why have we changed the terminology here from package to project? I personally feel this is out of the scope of this PR and we need to have a wider conversation with product and UX before changing our terminology as this will need to change in many other places in the user journey.

pkg/compute/init.go
```diff
@@ -118,6 +117,10 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			StarterKits: c.Globals.File.StarterKits.AssemblyScript,
 			Toolchain:   NewAssemblyScript(),
 		}),
+		NewLanguage(&LanguageOptions{
+			Name:        "other",
+			DisplayName: "Other ('bring your own' WASM binary)",
```
phamann, line 122: > WebAssembly (abbreviated Wasm)

From https://webassembly.org/ its Wasm not WASM 😉  
```suggestion
			DisplayName: "Other ('bring your own' Wasm binary)",
```

Similar comments elided for the rest of the PR, there are many references, some `WASM` and some `wasm`.

phamann (changes requested):
Looking good - have requested some changes:

1) Question regarding an important change in terminology that has wider implications than just the CLI. Therefore, I suggest we revert for the time being,
2) Nitty ask to use the correct Wasm instead of WASM.

pkg/compute/init.go
Integralist: Oh that's annoying 🙂 as I could have sworn I Google'd the naming and saw WASM used 🤦🏻 

Thanks for flagging this @phamann 👍🏻

phamann (approved):
LGTM

PR #283, opened 2021-05-18
Replace archived go-git dependency

Integralist:
**Problem**: the go-git dependency doesn't support SSH, resulting in strange low-level authentication errors.
**Solution**: replace go based implementation of git with a subprocess call to the user's `git` binary.

## Example

```
$ fastly compute init --name test -d "thing" -a "$(whoami)" -f git@github.com:fastly/compute-starter-kit-rust-default.git -l Rust
Creating a new Compute@Edge project.

Press ^C at any time to quit.


✓ Initializing...
✓ Checking if git is installed...
✓ Fetching package template...
✓ Updating package manifest...
✓ Initializing package...

Initialized package test to:
        /Users/integralist/Code/go/testing-cli-cloningrepo

        fastly compute publish

To learn about deploying Compute@Edge projects using third-party orchestration tools, visit:
        https://developer.fastly.com/learning/integrations/orchestration/


SUCCESS: Initialized package test
```

Now if I `tree` the current directory...

```
$ tree
.
├── Cargo.lock
├── Cargo.toml
├── LICENSE
├── README.md
├── SECURITY.md
├── fastly.toml
├── rust-toolchain
└── src
    └── main.rs
```

If `git` is not installed you should see:

```
✗ Initializing...

ERROR: `git` not found in $PATH.

The Fastly CLI requires a local installation of git.  For installation instructions for your operating system see:

        $ https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
```


pkg/compute/init.go
```diff
@@ -171,6 +169,15 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		progress = text.NewQuietProgress(out)
 	}
 
+	progress.Step("Checking if git is installed...")
+	_, err = exec.LookPath("git")
+	if err != nil {
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("`git` not found in $PATH"),
+			Remediation: fmt.Sprintf("To fix this error, follow the installation instructions for your operating system:\n\n\t$ %s", text.Bold("https://git-scm.com/book/en/v2/Getting-Started-Installing-Git")),
```
triblondon, line 177:
```suggestion
			Remediation: fmt.Sprintf("The Fastly CLI requires a local installation of git.  For installation instructions for your operating system see:\n\n\t$ %s", text.Bold("https://git-scm.com/book/en/v2/Getting-Started-Installing-Git")),
```

triblondon (approved):

pkg/compute/init.go
```diff
@@ -171,6 +169,15 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		progress = text.NewQuietProgress(out)
 	}
 
+	progress.Step("Checking if git is installed...")
```
triblondon, line 172: Don't feel like we need to print this to the output.  Just print the error if it's not there.

PR #285, opened 2021-05-19
Update dead link in historical stats command

kailan:


Integralist (approved):

Integralist:
@kailan have approved although the test is failing as I don't want to block you. But ultimately you just need to update the link in the help output that the test is validating.

PR #286, opened 2021-05-21
Surface any cargo metadata errors

fgsch:
While here also fix a logic bug.

### DIFF
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -64,17 +64,15 @@ type CargoMetadata struct {
 // Read the contents of the Cargo.lock file from filename.
 func (m *CargoMetadata) Read() error {
 	cmd := exec.Command("cargo", "metadata", "--quiet", "--format-version", "1")
-	stdout, err := cmd.StdoutPipe()
+	stdoutStderr, err := cmd.CombinedOutput()
 	if err != nil {
+		if len(stdoutStderr) > 0 {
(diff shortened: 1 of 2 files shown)


fgsch:
Example out before and after:

Before:

```
ERROR: error reading cargo metadata: EOF.
```

After:

```
ERROR: error reading cargo metadata: error: failed to parse manifest at `/private/tmp/b/Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  expected an equals, found an identifier at line 11 column 3
.
```

pkg/compute/rust.go
```diff
@@ -64,17 +64,15 @@ type CargoMetadata struct {
 // Read the contents of the Cargo.lock file from filename.
 func (m *CargoMetadata) Read() error {
 	cmd := exec.Command("cargo", "metadata", "--quiet", "--format-version", "1")
-	stdout, err := cmd.StdoutPipe()
+	stdoutStderr, err := cmd.CombinedOutput()
```
Integralist, line 67: @phamann do you have any historical context around the use of `.StdoutPipe()`?

Integralist (approved):
Going to approve as there's been no further feedback re: historical reason for using `StdoutPipe`. I checked the original PR that introduced the change and that doesn't indicate any specific reasoning for using that over `CombinedOutput`.

Issue #141, opened 2020-07-09
Operations on current and active version


jregistr:
👋  Hello Team!
Fantastic CLI! and I'm excited about its future.
During my usage over the last few days, I came up w/ some small (perhaps) ideas I thought I'd share with y'all. Please feel free to 👍  or 👎  .

For the subcommands that require a version number, it'd be nice if I can simply set it to use the latest version or 
the currently active version. 
Here are some potential ideas:
```bash
# Specify some defined keywords: LATEST | ACTIVE
fastly logging syslog list -s '...' --version=LATEST

# Flags for latest and active
fastly logging syslog list -s '...' --version-latest
```

I'd be happy to take a crack at implementation. Lemme know what y'all think. 
Thanks!

PR #290, opened 2021-05-27
Overload `--version` flag to accept multiple values

Integralist:
**Problem**: identifying the correct service version to use is a manual process which slows down a user's workflow.
**Solution**: overload the `--version` flag and implement a new `--autoclone` flag (fixes: https://github.com/fastly/cli/issues/141).
**Notes**: the implementation can be summarised as follows:

```
--version=VERSION   Number of service version, 'latest', 'active'
                    or 'editable'
--autoclone         Automatically clone the identified service
                    version if it's 'locked' or 'active'
```

- Every command that implements `--version` will call a `serviceVersion.Parse` function.
  - This function makes an API call for a list of all available service versions.
  - The returned list is sorted in descending order (i.e. newest versions first).
  - If the `--version` flag is set with the value `latest`:
    - We consider the selected version to be the latest version available.
    - The version _could_ be 'active' or 'locked', so `--autoclone` might need to be specified too.
    - We do not expect the returned version to be older than the last active version.
  - If the `--version` flag is set with the value `active`:
    - We consider the selected version to be the latest 'active' version.
  - If the `--version` flag is set with a numerical value (e.g. `1`, `2`, `3` etc):
    - We consider the selected version to be the given version.
  - If the `--version` flag is set with the value `editable`:
    - We consider the selected version to be the latest _editable_ version.
    - If we find an active version before an editable version, then we return an error.
  - If the `--version` flag is _not_ set (or an empty value is provided):
    - We consider the selected version to be the latest _editable_ version.
    - If we find an active version before an editable version, then we return an error.
- Once the version has been identified, if `--autoclone` is defined, we call a `autoclone.Parse` function.
  - This function checks if the version is 'active' or 'locked' and will clone it if necessary.
  - The `autoclone.Parse` function should only be called inside of commands where an 'editable' version is required.
  - The `--autoclone` flag should only be defined on commands where an 'editable' version is required.

### Version not set, no editable versions available

The user has a service with a single version that is 'active'. They execute `backend create` without passing the `--version` flag. 

In this scenario the CLI would internally call `getLatestEditableVersion` and, because there are no editable versions available, it would return an error message such as "error retrieving an editable service version".

The resolution for the user would be to pass both the `--version` flag and the `--autoclone` flag. The value of the `--version` flag could be either `latest`, `active` or `1`. With `--autoclone` also specified this would cause a new editable service version `2` to be created, and the backend created within that new version.

On subsequent calls to `backend create` the user could then omit the `--version` and `--autoclone` flags as the CLI would infer calling `getLatestEditableVersion` and (unless the user had 'activated' the newly cloned version) would result in the newly cloned/editable version being identified and the new backend resource created within it.

Alternatively, on subsequent calls the user could specify either `--version=latest` or `--version=editable`. Because each CLI invocation will acquire an updated versions list, which is then sorted in descending order, this should ensure (regardless of whether the value `latest` or `editable` was provided) the version returned was the newly cloned/editable version and thus `--autoclone` wouldn't be needed on the subsequent calls.

### Version not set, editable version available

The user has a service with a single version that is editable (i.e. it is neither 'locked' nor 'active'). They execute `backend create` multiple times without passing the `--version` flag.

In this scenario the CLI would internally call `getLatestEditableVersion` on each CLI invocation (resulting in the editable version `1` being identified each time), which means the three backends would be created within the same editable version.

Alternatively, the user could prefer to be _explicit_ about their intention and for each CLI invocation they could pass either `--version=editable`, which also would internally call `getLatestEditableVersion`, or `--version=latest`. Both would result in the editable version `1` being identified each time.

### DIFF
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -42,11 +41,12 @@ type DeployCommand struct {
 
 	// NOTE: these are public so that the "publish" composite command can set the
 	// values appropriately before calling the Exec() function.
-	Path        string
-	Version     common.OptionalInt
-	Domain      string
-	Backend     string
-	BackendPort uint
+	Path           string
+	Domain         string
+	Backend        string
+	BackendPort    uint
+	ServiceVersion common.OptionalServiceVersion
+	AutoClone      common.OptionalAutoClone
 }
 
 // NewDeployCommand returns a usable command registered under the parent.
@@ -60,12 +60,12 @@ func NewDeployCommand(parent common.Registerer, client api.HTTPClient, globals *
 	// NOTE: when updating these flags, be sure to update the composite command:
 	// `compute publish`.
 	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
-	c.CmdClause.Flag("version", "Number of version to activate").Action(c.Version.Set).IntVar(&c.Version.Value)
+	c.NewServiceVersionFlag(common.ServiceVersionFlagOpts{Dst: &c.ServiceVersion.Value, Optional: true, Action: c.ServiceVersion.Set}, "true")
+	c.NewAutoCloneFlag(c.AutoClone.Set, &c.AutoClone.Value)
 	c.CmdClause.Flag("path", "Path to package").Short('p').StringVar(&c.Path)
 	c.CmdClause.Flag("domain", "The name of the domain associated to the package").StringVar(&c.Domain)
 	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.Backend)
 	c.CmdClause.Flag("backend-port", "A port number for the package backend").UintVar(&c.BackendPort)
-
 	return &c
 }
 
@@ -116,17 +116,23 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 
 		text.Break(out)
 	} else {
+		version, err = c.ServiceVersion.Parse(serviceID, c.Globals.Client)
+		if err != nil {
+			return err
+		}
+		version, err = c.AutoClone.Parse(version, serviceID, c.Globals.Client)
+		if err != nil {
+			return err
+		}
+
 		// We define the `ok` variable so that the following call to
-		// `validateservice` will be able to shadow multiple variables.
+		// `validateservice` will be able to shadow `invalidType`.
 		var ok bool
 
 		// Because a service_id exists in the fastly.toml doesn't mean it's valid
 		// e.g. it could be missing either a domain or backend resource. So we
 		// check and allow the user to configure these settings before continuing.
-		//
-		// The returned version will only be nil if the initial service version
-		// request failed.
-		version, ok, invalidType, err = validateService(serviceID, c.Globals.Client, c.Version)
+		ok, invalidType, err = validateService(serviceID, c.Globals.Client, version)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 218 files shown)


pkg/compute/deploy.go
```diff
@@ -320,95 +326,57 @@ func pkgPath(path string, name string, source manifest.Source) (string, error) {
...
 	}
 
-	return v, true, resourceNone, nil
+	return true, resourceNone, nil
 }
 
-// serviceVersion returns the version for the given service.
-func serviceVersion(serviceID string, client api.Interface, versionFlag common.OptionalInt) (*fastly.Version, error) {
+// checkServiceID validates the given Service ID maps to a real service.
+func checkServiceID(serviceID string, client api.Interface, version *fastly.Version) error {
```
Integralist, line 372: The `compute deploy` command was originally implementing similar logic to what is being proposed in this PR. Specifically it would attempt to acquire an 'ideal' service version to deploy a package into. This logic has been superseded by the new overloaded `--version` and `--autoclone` flag parsers.

pkg/domain/describe.go
```diff
@@ -26,7 +27,7 @@ func NewDescribeCommand(parent common.Registerer, globals *config.Data) *Describ
 	c.manifest.File.Read(manifest.Filename)
 	c.CmdClause = parent.Command("describe", "Show detailed information about a domain on a Fastly service version").Alias("get")
 	c.CmdClause.Flag("service-id", "Service ID").Short('s').StringVar(&c.manifest.Flag.ServiceID)
-	c.CmdClause.Flag("version", "Number of service version").Required().IntVar(&c.Input.ServiceVersion)
+	c.NewServiceVersionFlag(common.ServiceVersionFlagOpts{Dst: &c.serviceVersion.Value})
```
Integralist, line 30: Some commands don't require `--autoclone` because even if a service version is identified as 'locked' or 'active' it won't matter as far as the command behaviour. 

For example, when listing or describing resources, you can still do that with locked/active service versions.

pkg/testutil/api.go
```diff
@@ -0,0 +1,84 @@
+package testutil
```
Integralist, line 1: This is the new `testutil` package file that contains the shared test data now consumed by all tests.

Integralist (commented):
I've commented on all the sections I think are worth 'highlighting'

The rest of the changes break down into:

1. integrating the new flags and calling their associated behaviours within each command file.
2. updating the tests for every command affected by these changes.

triblondon:
> The returned list is sorted by the UpdatedAt field.

I think we should sort it by version number descending.

> If the --version flag is set with the value latest:
> We consider the selected version to be the latest version that isn't itself 'active'.
> The version could be 'locked', so --autoclone might need to be specified too (depending on the situation).

I'd prefer this to be the version with the highest number.

> If the --version flag is set with the value editable:
> If the --version flag is not set (or an empty value is provided
> We consider the selected version to be the latest editable version.

Confirm: 'latest' here means the one with the highest number?)

I'd also like to see a guard here: we should never automatically select a version that is lower than the active.  If you want to do something with a version that is lower than the active, you can specify it explicitly.

Integralist:
For posterity, as I'll be updating the documented behaviours in the PR description, below is the _original_ PR implementation...

- Every command that implements `--version` will call a `serviceVersion.Parse` function.
  - This function makes an API call for a list of all available service versions.
  - The returned list is sorted by the `UpdatedAt` field.
  - If the `--version` flag is set with the value `active`:
    - We consider the selected version to be the latest 'active' version.
  - If the `--version` flag is set with the value `latest`:
    - We consider the selected version to be the latest version that isn't itself 'active'.
    - The version _could_ be 'locked', so `--autoclone` might need to be specified too (depending on the situation).
  - If the `--version` flag is set with a numerical value (e.g. `1`, `2`, `3` etc):
    - We consider the selected version to be the given version.
  - If the `--version` flag is set with the value `editable`:
    - We consider the selected version to be the latest _editable_ version.
  - If the `--version` flag is _not_ set (or an empty value is provided):
    - We consider the selected version to be the latest _editable_ version.
- Once the version has been identified, if `--autoclone` is defined, we call a `autoclone.Parse` function.
  - This function checks if the version is 'active' or 'locked' and will clone it if necessary.
  - The `autoclone.Parse` function should only be called inside of commands where an 'editable' version is required.

Integralist:
@triblondon while I wait for a technical review, would you mind approving from the high-level 'flow' perspective.

Integralist:
This PR is being closed in favour of two separate PRs. The first will define the new flags, and the second will implement them across the code base.

Here is the first PR: https://github.com/fastly/cli/pull/300

PR #291, opened 2021-06-02
Fix bug in parsing port number for backend

Integralist:
When prompting the user for a backend port, we incorrectly try to convert an empty string into an integer.


triblondon (approved):
Looks reasonable

PR #292, opened 2021-06-03
fix bug in `publish` command caused by lack of flag override

Integralist:
**Problem**: A report came in to say that using `publish` with the `--service-id` flag was causing an error stating the relevant service could not be found, while switching to the `deploy` command (which publish proxies to) the flag would work as expected. 

This was caused by the `publish` command not overriding the manifest field used by the `deploy` command it was proxying to.

### DIFF
--- a/pkg/compute/deploy.go
+++ b/pkg/compute/deploy.go
@@ -38,10 +38,10 @@ const (
 // DeployCommand deploys an artifact previously produced by build.
 type DeployCommand struct {
 	common.Base
-	manifest manifest.Data
 
 	// NOTE: these are public so that the "publish" composite command can set the
 	// values appropriately before calling the Exec() function.
+	Manifest    manifest.Data
 	Path        string
 	Version     common.OptionalInt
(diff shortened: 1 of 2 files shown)


triblondon (approved):

gaastonsr:
Is there a chance this regressed? I'm using Fastly CLI version `0.31.0` and running

```sh
$ fastly compute publish --service-id=MY_SERVICE_ID
✓ Initializing...
✓ Building package using rust toolchain...
✓ Creating package archive...

SUCCESS: Built rust package compute-edge-service (pkg/compute-edge-service.tar.gz)


ERROR: the Fastly API returned 404 Not Found: Record not found (Cannot find service 'SERVICE_ID_I_WANT_TO_OVERRIDE').
```

Integralist:
@gaastonsr Apologies, yes this looks to be a regression caused by https://github.com/fastly/cli/pull/302 which accidentally undid these changes. 

I'll get a PR worked on to resolve this.

PR #293, opened 2021-06-03
Move undo behaviour from common to own package

Integralist:
**Problem**: I'm increasingly stumbling into issues with the `common` package being too broad/generic. This materialises as an "import cycle error". It is also considered bad practice to have folders named `common` or `utils` etc.
**Solution**: Move these behaviours into their own packages.
**Example**: I would like to add a remediation error type to some code in our `config` package. This isn't possible because importing the `errors` package will import `text` which will import `common` and the common package has already been imported earlier in the import cycle:

```
package github.com/fastly/cli/cmd/fastly
        imports github.com/fastly/cli/pkg/app
        imports github.com/fastly/cli/pkg/backend
        imports github.com/fastly/cli/pkg/common << command stuff like: Base/Register/StreamingExec + lots more other misc stuff.
        imports github.com/fastly/cli/pkg/config
        imports github.com/fastly/cli/pkg/errors
        imports github.com/fastly/cli/pkg/text
        imports github.com/fastly/cli/pkg/common: import cycle not allowed
```


fgsch (approved):
👍

triblondon (approved):

PR #300, opened 2021-06-04
Define new --version and --autoclone flags

Integralist:
**Problem**: Identifying the correct service version to use is a manual process which slows down a user's workflow.
**Solution**: Overload the `--version` flag and implement a new `--autoclone` flag (fixes: https://github.com/fastly/cli/issues/141).
**Notes**: This PR defines the new flags (and tests for the internal behaviours) but doesn't integrate them yet (see follow-up PRs).

The implementation can be summarised as follows:

```
--version=VERSION   Number of service version, 'latest', 'active'
                    or 'editable'
--autoclone         Automatically clone the identified service
                    version if it's 'locked' or 'active'
```

- Every command that implements `--version` will call a `serviceVersion.Parse` function.
  - This function makes an API call for a list of all available service versions.
  - The returned list is sorted in descending order (i.e. newest versions first).
  - If the `--version` flag is set with the value `latest`:
    - We consider the selected version to be the latest version available.
    - The version _could_ be 'active' or 'locked', so `--autoclone` might need to be specified too.
    - We do not expect the returned version to be older than the last active version.
  - If the `--version` flag is set with the value `active`:
    - We consider the selected version to be the latest 'active' version.
  - If the `--version` flag is set with a numerical value (e.g. `1`, `2`, `3` etc):
    - We consider the selected version to be the given version.
  - If the `--version` flag is set with the value `editable`:
    - We consider the selected version to be the latest _editable_ version.
    - If we find an active version before an editable version, then we return an error.
  - If the `--version` flag is _not_ set (or an empty value is provided):
    - We consider the selected version to be the latest _editable_ version.
    - If we find an active version before an editable version, then we return an error.
- Once the version has been identified, if `--autoclone` is defined, we call a `autoclone.Parse` function.
  - This function checks if the version is 'active' or 'locked' and will clone it if necessary.
  - The `autoclone.Parse` function should only be called inside of commands where an 'editable' version is required.
  - The `--autoclone` flag should only be defined on commands where an 'editable' version is required.

### Version not set, no editable versions available

The user has a service with a single version that is 'active'. They execute `backend create` without passing the `--version` flag. 

In this scenario the CLI would internally call `getLatestEditableVersion` and, because there are no editable versions available, it would return an error message such as "error retrieving an editable service version".

The resolution for the user would be to pass both the `--version` flag and the `--autoclone` flag. The value of the `--version` flag could be either `latest`, `active` or `1`. With `--autoclone` also specified this would cause a new editable service version `2` to be created, and the backend created within that new version.

On subsequent calls to `backend create` the user could then omit the `--version` and `--autoclone` flags as the CLI would infer calling `getLatestEditableVersion` and (unless the user had 'activated' the newly cloned version) would result in the newly cloned/editable version being identified and the new backend resource created within it.

Alternatively, on subsequent calls the user could specify either `--version=latest` or `--version=editable`. Because each CLI invocation will acquire an updated versions list, which is then sorted in descending order, this should ensure (regardless of whether the value `latest` or `editable` was provided) the version returned was the newly cloned/editable version and thus `--autoclone` wouldn't be needed on the subsequent calls.

### Version not set, editable version available

The user has a service with a single version that is editable (i.e. it is neither 'locked' nor 'active'). They execute `backend create` multiple times without passing the `--version` flag.

In this scenario the CLI would internally call `getLatestEditableVersion` on each CLI invocation (resulting in the editable version `1` being identified each time), which means the three backends would be created within the same editable version.

Alternatively, the user could prefer to be _explicit_ about their intention and for each CLI invocation they could pass either `--version=editable`, which also would internally call `getLatestEditableVersion`, or `--version=latest`. Both would result in the editable version `1` being identified each time.

### DIFF
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -71,7 +71,7 @@ func (sv *OptionalServiceVersion) Parse(sid string) (*fastly.Version, error) {
 	case "latest":
 		return vs[0], nil
 	case "active":
-		v, err = getLatestActiveVersion(vs)
+		v, err = getActiveVersion(vs)
 	case "":
 		return vs[0], nil
 	default:
@@ -119,8 +119,8 @@ func (ac *OptionalAutoClone) Parse(v *fastly.Version, sid string) (*fastly.Versi
 	return v, nil
 }
 
-// getLatestActiveVersion returns the active service version.
-func getLatestActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
+// getActiveVersion returns the active service version.
+func getActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
 	for _, v := range vs {
 		if v.Active {
 			return v, nil
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -1,14 +1,12 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/fastly/cli/pkg/api"
-	fsterr "github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/go-fastly/v3/fastly"
 	"github.com/fastly/kingpin"
 )
@@ -71,10 +69,8 @@ func (sv *OptionalServiceVersion) Parse(sid string, c api.Interface) (*fastly.Ve
 		return vs[0], nil
 	case "active":
 		v, err = getLatestActiveVersion(vs)
-	case "editable":
-		v, err = getLatestEditableVersion(vs)
 	case "":
-		v, err = getLatestEditableVersion(vs)
+		return vs[0], nil
 	default:
 		v, err = getSpecifiedVersion(vs, sv.Value)
 	}
@@ -123,36 +119,6 @@ func getLatestActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
 		}
 	}
 	return nil, fmt.Errorf("no active service version found")
-}
-
-// getLatestEditableVersion returns the latest editable service version.
-//
-// The returned version will be ahead of the latest active version. If there is
-// no editable version above the latest active, an error will be returned.
-//
-// When no --version flag is provided, this algorithm helps handle cases where
-// a command (such as `backend create`) is executable multiple times, as the
-// latest editable version will be reused and subsequently will contain each
-// new backend created from the prior executions.
-//
-// There could be a scenario where a customer has a single service version
-// which is activated and so there would be no match for an editable version
-// without us first cloning it. The act of cloning should be a decision made by
-// the user (i.e. --autoclone) and so this function will return an error if no
-// editable version available.
-func getLatestEditableVersion(vs []*fastly.Version) (*fastly.Version, error) {
-	for _, v := range vs {
-		if v.Active {
-			break
-		}
-		if !v.Locked {
-			return v, nil
-		}
-	}
-	return nil, fsterr.RemediationError{
-		Inner:       errors.New("error retrieving an editable service version: no editable version found"),
-		Remediation: "Please retry with --version=[latest|active] and --autoclone",
-	}
 }
 
 // getSpecifiedVersion returns the specified service version.
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -24,10 +24,10 @@ type ServiceVersionFlagOpts struct {
 	Action   kingpin.Action
 }
 
-// NewServiceVersionFlag defines a --version flag that accepts multiple values
+// SetServiceVersionFlag defines a --version flag that accepts multiple values
 // such as 'latest', 'active' and numerical values which are then converted
 // into the appropriate service version.
-func (b Base) NewServiceVersionFlag(opts ServiceVersionFlagOpts, args ...string) {
+func (b Base) SetServiceVersionFlag(opts ServiceVersionFlagOpts, args ...string) {
 	clause := b.CmdClause.Flag("version", "'latest', 'active', or the number of a specific version")
 	if !opts.Optional {
 		clause = clause.Required()
@@ -37,9 +37,9 @@ func (b Base) NewServiceVersionFlag(opts ServiceVersionFlagOpts, args ...string)
 	clause.StringVar(opts.Dst)
 }
 
-// NewAutoCloneFlag defines a --autoclone flag that will cause a clone of the
+// SetAutoCloneFlag defines a --autoclone flag that will cause a clone of the
 // identified service version if its found to be active or locked.
-func (b Base) NewAutoCloneFlag(action kingpin.Action, dst *bool) {
+func (b Base) SetAutoCloneFlag(action kingpin.Action, dst *bool) {
 	b.CmdClause.Flag("autoclone", "If the selected service version is not editable, clone it and use the clone.").Action(action).BoolVar(dst)
 }
 
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -122,7 +122,7 @@ func getLatestActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
 			return v, nil
 		}
 	}
-	return nil, fmt.Errorf("error locating latest active service version")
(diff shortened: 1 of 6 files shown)


pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+type OptionalServiceVersion struct {
+	OptionalString
+}
+
+// Parse returns a service version based on the given user input.
+func (sv *OptionalServiceVersion) Parse(sid string, c api.Interface) (*fastly.Version, error) {
+	vs, err := c.ListVersions(&fastly.ListVersionsInput{
+		ServiceID: sid,
+	})
+	if err != nil || len(vs) == 0 {
```
fgsch, line 58: I wonder if we should behave differently for the `len(vs) == 0` case.
If the user did not use `--version` what do we do in that case?

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// getLatestActiveVersion returns the latest active service version.
+func getLatestActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
+	for _, v := range vs {
+		if v.Active {
+			return v, nil
+		}
+	}
+	return nil, fmt.Errorf("error locating latest active service version")
```
fgsch, line 125: It might make sense to reword this error to something like `no active service version found` to make it clear that it was not a problem reading the versions.

pkg/cmd/flags.go
fgsch: Actually this probably doesn't make (i.e. we should always have at least on version) so nevermind.

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+// which is activated and so there would be no match for an editable version
+// without us first cloning it. The act of cloning should be a decision made by
+// the user (i.e. --autoclone) and so this function will return an error if no
+// editable version available.
+func getLatestEditableVersion(vs []*fastly.Version) (*fastly.Version, error) {
+	for _, v := range vs {
+		if v.Active {
+			break
+		}
+		if !v.Active && !v.Locked {
```
fgsch, line 148: I believe you can drop the `!v.Active` check here.

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+
+	for _, v := range vs {
+		if v.Number == i {
+			return v, nil
+		}
+	}
+
+	return nil, fmt.Errorf("error getting specified service version: %s", version)
```
fgsch, line 171: I'd probably reword this to make it clear that the version was not found as opposed to some other error.

fgsch:
Looks good to me. Left a few minor comments around error messages.

fgsch (approved):
Looks good to me.

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+	Dst      *string
+	Optional bool
+	Action   kingpin.Action
+}
+
+// NewServiceVersionFlag defines a --version flag that accepts multiple values
+// such as 'latest', 'active' and numerical values which are then converted
+// into the appropriate service version.
+func (b Base) NewServiceVersionFlag(opts ServiceVersionFlagOpts, args ...string) {
+	clause := b.CmdClause.Flag("version", "Number of service version, 'latest', 'active' or 'editable'")
```
triblondon, line 33: This sounds a bit like "Number of [service version, latest, active or editable]", not "[Number of service version], latest, active or editable".  Resolve that ambiguity:

```suggestion
	clause := b.CmdClause.Flag("version", "'latest', 'active', 'editable', or the number of a specific service version")
```

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+		clause = clause.Required()
+	} else {
+		clause = clause.Action(opts.Action)
+	}
+	clause.StringVar(opts.Dst)
+}
+
+// NewAutoCloneFlag defines a --autoclone flag that will cause a clone of the
+// identified service version if its found to be active or locked.
+func (b Base) NewAutoCloneFlag(action kingpin.Action, dst *bool) {
```
triblondon, line 44: What's the logic behind adding 'New' into the name of these functions?

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+	b.CmdClause.Flag("autoclone", "Automatically clone the identified service version if it's 'locked' or 'active'").Action(action).BoolVar(dst)
```
triblondon, line 45: Would it be too long to say this?

```suggestion
	b.CmdClause.Flag("autoclone", "If the selected service version is not editable, clone it and use the clone.").Action(action).BoolVar(dst)
```

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		return version, nil
+	}
+
+	// Treat the function as a no-op if the version is editable.
+	return v, nil
+}
+
+// getLatestActiveVersion returns the latest active service version.
+func getLatestActiveVersion(vs []*fastly.Version) (*fastly.Version, error) {
```
triblondon, line 119: Nit: there can only be one active version, so it's not the latest one, it's just the active

triblondon (commented):
This is looking good.  I'm still slightly unsure about the utility of 'editable'.  Otherwise and with a couple of small copy edits, this looks great.

## Change default from 'editable' to 'latest'?

For the `backend create` scenario, the default behaviour we're creating requires that the first and second invocations use different flags:

```
fastly backend create --version=active --autoclone <-- creates the new draft version
fastly backend create <-- uses the draft version created in the previous step
```

If you thought you maybe already had a draft but didn't and tried to run the first command without flags, it would error, so fair enough.  But if you thought you didn't have a draft and actually you do, then that first command will create another draft from the active, possibly losing changes you've made to your existing draft.

If we were to change the default behaviour so that if you don't specify a version, you get `latest` instead of `active`, then this improves the experience in all scenarios, because regardless of whether you have an open draft or not, your command is the same:

```
fastly backend create --autoclone <-- clones from latest if latest is not editable
fastly backend create --autoclone <-- uses editable draft
```

## Special case for 'compute deploy'?

Are we retaining the special case for deploy that autoclones and activates by default?  Is there an opportunity to hook it into the new code rather than retaining dedicated code for that behaviour?  Or maybe we wait and do this if/when we ship `--activate`?

pkg/cmd/flags.go
```diff
@@ -0,0 +1,172 @@
...
+//
+// The returned version is either the same as the input argument `v` or it's a
+// cloned version if the input argument was either active or locked.
+func (ac *OptionalAutoClone) Parse(v *fastly.Version, sid string, c api.Interface) (*fastly.Version, error) {
+	// if user didn't provide --autoclone flag
+	if !ac.Value && (v.Active || v.Locked) {
+		return nil, fmt.Errorf("service version %d is not editable", v.Number)
+	}
+	if ac.Value && (v.Active || v.Locked) {
+		version, err := c.CloneVersion(&fastly.CloneVersionInput{
```
triblondon, line 104: Does this produce any output?  If not, I'd consider dropping in here a line of output that says:

> Service version {a} is not editable, so it was automatically cloned because `--autoclone` is enabled.  Now operating on version {b}...

pkg/cmd/flags.go
```diff
@@ -85,25 +88,30 @@ func (sv *OptionalServiceVersion) Parse(sid string, c api.Interface) (*fastly.Ve
...
+		version, err := ac.Client.CloneVersion(&fastly.CloneVersionInput{
 			ServiceID:      sid,
 			ServiceVersion: v.Number,
 		})
 		if err != nil {
 			return nil, fmt.Errorf("error cloning service version: %w", err)
 		}
+		msg := fmt.Sprintf("Service version %d is not editable, so it was automatically cloned because --autoclone is enabled. Now operating on version %d.", v.Number, version.Number)
+		text.Output(ac.Out, msg)
+		text.Break(ac.Out)
```
fgsch, line 114: Shouldn't this be limited to when verbose output was requested?

triblondon (approved):

pkg/cmd/flags.go
Integralist: Agreed. Will modify

PR #302, opened 2021-06-09
Implement new --version and --autoclone flags

Integralist:
The new `--version` and `--autoclone` flags were already present in the codebase, but were not yet _integrated_. This PR integrates the new flags across all CLI commands.

At a high-level this PR consists generally of the following changes:

- Most commands now implement `SetServiceVersionFlag` and `SetAutoCloneFlag`.
  - **NOTE**: Some subcommand behaviours (e.g. describe/list) don't require `--autoclone` (so no `SetAutoCloneFlag` used).
- All affected commands use a single function (`cmd.ServiceDetails`) for acquiring the Service ID and Service Version. 
  - **NOTE**: As a result a few commands could, at the same time, be corrected to not mark `--service-id` as 'required'.
- All tests that mock a Service Version now use a single/consistent implementation (`pkg/testutil/api.go`).
  - **NOTE**: These tests needed updating anyway to both mock the Service Version API and to add `--autoclone`.
- The `OptionalServiceVersion.Parse`/`OptionalAutoClone.Parse` functions now accept a HTTP client.
  - **NOTE**: The original implementation in `main` had the client on the receiver type but I realised that wouldn't have worked.


pkg/compute/deploy.go
```diff
@@ -8,7 +8,6 @@ import (
 	"net"
```
Integralist, line 8: Need to autoclone.

triblondon (approved):
I ran through this with @Integralist on a 1:1.  It's big so we were not able to look at it line by line but the patterns made sense and I had a couple of points:

* In the `compute deploy` command, we need to make `--autoclone` the default behaviour since this is a composite operation, which already includes cloning and activating of the resulting version.  
* I'd also like to make sure that we are clear in the source the difference between "API commands" and "composite/special" commands, as the way we apply coding patterns in each is quite different, and those that map to API endpoints directly should rightly be more formulaic and - perhaps in future - generatable?!

Integralist:
I've been testing the code locally with no problems and am just waiting for Kailan to test locally (and to approve once they're happy) and then I'll get this PR merged.

kailan:
Tested and working as expected 🎉 

Chatted with @Integralist regarding defaults for these flags, but this is a discussion that he has already had with others so I will be quiet and approve!

kailan (approved):
Tested and working as expected 🎉 

Chatted with @Integralist regarding defaults for these flags, but this is a discussion that he has already had with others so I will be quiet and approve!

Issue #301, opened 2021-06-07
Feature request: take service ID from FASTLY_SERVICE_ID 


peterbourgon:
I tend to use the CLI in a "session" where I am working on various properties of a single service. Basically every time I invoke `fastly` I have to pass a service ID as a parameter.

If I happen to be in a directory with a `fastly.toml` manifest file, I can put the service ID in there, where it's used as a default. Otherwise, I have to specify `--service-id` with each command. This immediately becomes very tedious, for what I hope are obvious reasons.

It should be possible to specify a service ID via a `FASTLY_SERVICE_ID` env var. In case the service ID is specified in multiple ways, the priority order would be

1. File (lowest priority)
2. Env var
3. Flag (highest priority)

Let me know if you're down to clown, I can do the PR.

Integralist:
This makes sense to me. Happy for you to contribute a PR 👍🏻

PR #306, opened 2021-06-12
Support FASTLY_SERVICE_ID environment variable

peterbourgon:
Addresses #301.

```
$ fastly logs tail --service-id abc
ERROR: the Fastly API returned 404 Not Found: Record not found (Cannot find service 'abc').

$ env FASTLY_SERVICE_ID=abc fastly logs tail
ERROR: the Fastly API returned 404 Not Found: Record not found (Cannot find service 'abc').

$ env FASTLY_SERVICE_ID=abc fastly logs tail --service-id def
ERROR: the Fastly API returned 404 Not Found: Record not found (Cannot find service 'def').
```

How would you like this to be tested?

### DIFF
--- a/pkg/backend/create.go
+++ b/pkg/backend/create.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/fastly/cli/pkg/cmd"
 	"github.com/fastly/cli/pkg/config"
+	"github.com/fastly/cli/pkg/env"
 	"github.com/fastly/cli/pkg/text"
 	"github.com/fastly/go-fastly/v3/fastly"
 )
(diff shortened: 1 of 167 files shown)


pkg/backend/create.go
```diff
@@ -27,7 +28,7 @@ func NewCreateCommand(parent cmd.Registerer, globals *config.Data) *CreateComman
 	c.Globals = globals
 	c.CmdClause = parent.Command("create", "Create a backend on a Fastly service version").Alias("add")
 
-	c.CmdClause.Flag("service-id", "Service ID").Short('s').Required().StringVar(&c.Input.ServiceID)
+	c.CmdClause.Flag("service-id", "Service ID").Short('s').Envar(env.ServiceID).Required().StringVar(&c.Input.ServiceID)
```
peterbourgon, line 31: Should we also call out the env var in the help text, as `-token` does?

peterbourgon:
Or maybe helper function(s) would be better?

```go
func RegisterOptionalServiceID(c *kingpin.CmdClause, dst *string) {
    c.Flag("service-id", "Service ID").Short('s').Envar(env.ServiceID).StringVar(dst)
}

func RegisterRequiredServiceID(c *kingpin.CmdClause, dst *string) {
    c.Flag("service-id", "Service ID").Short('s').Envar(env.ServiceID).Required().StringVar(dst)
}
```

Integralist:
@peterbourgon 

> Or maybe helper function(s) would be better?

I think I'd prefer that approach 👍🏻 

As a side note, in `main` there's been some changes where by the Service ID flag is no longer explicitly required. This was to support falling back to reading the value from the fastly manifest using the [`ServiceID()`](https://github.com/fastly/cli/blob/c8dbff77b7407272789ea3e1239320b0c254fcda/pkg/compute/manifest/manifest.go#L78-L88) function.

Integralist:
I went ahead and implemented this feature here: https://github.com/fastly/cli/pull/320 as I need to cut a new CLI release soon and wanted this feature included in that release. This PR had looked to have gone stale with a bunch of changes in the `main` branch and so it was easier for me to implement this myself as I had the context around the changes that have been happening.

Thank you for the (internal) suggestion to implement this feature, I appreciate you taking the time to look into this.

peterbourgon:
Oh sorry!! This had entered my Backburner and I let it slide :( Thank you!

PR #312, opened 2021-06-18
Embed app config into compiled CLI binary

Integralist:
**Problem**: The logic for how to handle the local application config file, as well as how to deal with multiple _blocking_ calls to load the config from a remote endpoint, has become complex and is prone towards hard to debug errors.

**Solution**: Embed the latest version of the app config into the compiled CLI binary. This means we remove two blocking remote calls, and we can failover to the embedded config whenever there is an issue trying to read the config from the user's local disk. 

PLUS, we'll have tests for the config logic now! 🎉 

**Notes**: We'll use a GitHub Action to ensure when cutting a release that the binary has the latest version (at that time) of the configuration file embedded.

**Logic flow**:

- Attempt to read the config from the user's local disk.
  - If that fails, i.e. there is no local config, resort to using the embedded config.
- We then unmarshal the config into an in-memory struct.
  - If that fails, and we were using... 
    - **The embedded config**:
      - Exit the program with a remediation error because we don't expect our embedded config to be invalid toml.
    - **The user's local disk config**:
      - Prompt the user, requesting confirmation to use the embedded config (this is because we couldn't parse their local config, meaning we won't have access to their `User.Token`/`User.Email` and so that data would be lost if we switched to embedded config without asking the user first). 
        - If reading the user's input fails, then return an error.
        - If the user wants to continue, then resort to using the embedded config, and if _that_ fails:
          - Exit the program with a remediation error because we don't expect our embedded config to be invalid toml.
        - If the user doesn't want to continue, then return a remediation error that tells them they'll need to manually resolve any configuration syntax errors.
- If using the embedded config: ensure `LastChecked` and `Version` are set to the current time and current CLI version.
- We then write the in-memory struct back to disk.
  - This is for the scenario where we're relying on the embedded config.
- If the config looks to be in a legacy format, then we'll transition them to the embedded config format.
- If the config shows the CLI `Version` as not matching the CLI version currently running, then we'll transition them to the embedded config format (because it's possible their config might not be compatible with the updated CLI version).
- If the config is stale, then we'll fetch the latest config version asynchronously from the remote api endpoint.


.github/workflows/pr_test.yml
```diff
@@ -82,3 +82,18 @@ jobs:
       env:
         TEST_COMPUTE_INIT: true
         TEST_COMPUTE_BUILD: true
+  config:
```
Integralist, line 85: Move to step before: https://github.com/fastly/cli/blob/06b2ff56e365f98086d049408e5c89206825cae3/.github/workflows/tag_release.yml#L21 to prevent user having to run `make config`.

cmd/fastly/static/config.toml
```diff
@@ -0,0 +1,29 @@
+[fastly]
```
Integralist, line 1: Remove this file as it should only be pulled at CI release time.

pkg/config/data_test.go
```diff
@@ -36,3 +45,276 @@ func TestConfigLoad(t *testing.T) {
...
+	} else {
+		testutil.AssertErrorContains(t, err, config.ErrInvalidConfig.Error())
+	}
+
+	// Validate that legacy configuration can be migrated to the static one
+	// embedded in the CLI binary.
+	var out bytes.Buffer
+	f.Read(fpath, strings.NewReader(""), &out)
+	f.UseStatic(staticConfig, fpath)
+	if f.CLI.LastChecked == "" || f.CLI.Version == "" {
```
Integralist, line 216: Ensure we're able to validate that we have modern config that needs updating to latest version and so we validate the tokens aren't lost.

Integralist (commented):
Had a Zoom call with @kailan to review and have some things to double check before requesting a re-review.

kailan (approved):
All functioning as expected on my machine! Walked through code and CI process with @Integralist and suggested some changes, which have now been implemented.

Ship it! 🚀

PR #318, opened 2021-06-22
Resolve issues with `vcl snippet`

Integralist:
**Problem**: before cutting a new release I wanted to manually validate the new `vcl snippet` commands and found that the API was behaving differently to how I had implemented the logic (which was based on incorrect logic in go-fastly).
**Solution**: [update go-fastly](https://github.com/fastly/go-fastly/pull/281) to have correct logic, and then ensure the CLI logic consumes the new go-fastly release.
**Notes**: I think the best thing for a reviewer to do is to review the logic changes in the command's themselves, and not focus on the tests (which are just reflecting the command logic changes -- as the diff makes it hard to understand the changes at a glance).


Integralist:
@kailan FYI -- hold off reviewing this for the moment as there's more changes I have locally that are blocked by https://github.com/fastly/go-fastly/pull/281

triblondon (approved):
Reviewed on zoom

PR #329, opened 2021-07-02
`app.Run` to accept a struct

Integralist:
We currently have **_nine_** separate positional arguments needed to be passed to `app.Run` and it’s not clear from the many tests (or the business logic side) what the arguments actually mean. So I’ve cleaned this up by switching to a single argument: a `RunOpts` struct type.

To support this change I’ve updated `pkg/testutil/args.go` to import the `app` package so it can access the new `RunOpts` struct. 

**Side notes**

Because the `app` package is the root of the import flow we need to ensure our tests don’t use ‘white-box testing’ (i.e. has access to internal identifiers) otherwise we’ll get a cyclic import error.

This means some `_test.go` files need to have their package name changed from `foobar` to `foobar_test` (a pattern in Go for moving from white-box testing to black-box testing). 

This now means those test files need to explicitly import the package they’re testing and reference the package accordingly in the test code.

Some benefits fall out of this, the first is it means our command variables are consistently exported (currently we have a mix-mash of styles of exported and unexported properties due to different requirements).

Additionally there are duplicated mocked responses that can be deleted now thanks to sharing between the unit and integration tests.

Also the `pkg/testutil/args.go` implementation for testing the run function now has become much simpler/cleaner.

**Change Summary**

- Pass a struct to the run function.
- Update the positional argument references inside the run function.
- Update all tests to use the updated `testutil.NewRunOpts` (which replaces `testutil.NewAppRunArgs`).
- Update those commands affected by the cyclic import issue, so they have exported references.
- Delete a bunch of duplicated mocked API response data.

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -24,6 +24,10 @@ import (
 var cfg []byte
 
 func main() {
+	// Some configuration options can come from env vars.
+	var env config.Environment
+	env.Read(parseEnv(os.Environ()))
+
 	// All of the work of building the set of commands and subcommands, wiring
 	// them together, picking which one to call, and executing it, occurs in a
 	// helper function, Run. We parameterize all of the dependencies so we can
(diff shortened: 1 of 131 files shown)


Integralist (commented):
@triblondon left some comments to help you while you scan through these changes.

triblondon (approved):
Some nice DRYing here.  LGTM 👍

Issue #157, opened 2020-09-24
Support more than one Rust toolchain version


phamann:
Currently we force a single Rust toolchain version when compiling Rust based Compute@Edge programs via `fastly compute build`. Ideally we should support a SemVer range (minimum -> maximum) to be more flexible for user and CI toolchain environments.

pchickey:
Ideally we would always support the `stable` rust version - this special version points to the latest stable release that rustup knows about. So maybe we should express a disjunction of constaints?

PR #330, opened 2021-07-05
Support multiple versions of Rust

Integralist:
Fixes https://github.com/fastly/cli/issues/157

This PR creates a new property `toolchain_constraint` and will implement a config versioning check.

**Context**:
When a new CLI release is cut we embed into the binary a copy of the [current config](https://github.com/fastly/cli-config/blob/main/src/config.toml). This ensures the CLI is runnable, for first time users, with config that is known to work with the CLI code. After a predetermined TTL (currently 5mins) the config is considered stale and will be refreshed asynchronously from the [remote endpoint](https://developer.fastly.com/api/internal/cli-config).

This is fine when the loaded configuration is a simple value change (e.g. changing a TTL from 5mins to 10mins) but it becomes a problem when the change being made is fundamentally a breaking change, such as changing `toolchain_version` from a simple semver (i.e. `1.49.0`) to a 'range' (i.e. `>= 1.49.0 < 2.0.0`) as we do in this PR.

We can't just update the remote config to use a range because the [current CLI code](https://github.com/fastly/cli/blob/main/pkg/compute/rust.go#L181) won't work with that. So it means we need the CLI code to be updated first to work with a range. We also can't replace the `toolchain_version` value with a range because users of an older CLI version would experience a runtime error from the binary's old logic. We also don't want to have to wait for users to upgrade to a newer version of the CLI, and so the best option is for us to continue to publish `toolchain_version` while now publishing a new `toolchain_constraint` property. This means users of an older CLI version will find the binary still functions correctly while users who upgrade the CLI version will start using the new property.

**Config Versioning**

To help identify whether the config is valid, is to introduce a `config_version` property (loosely based on semver, where a bump in the value indicates a _breaking_ change and thus an updated CLI version will be required).  

I've added a `config_version` property and set it to `1`. In the future we would cut a new CLI release with logic that checks the `config_version` value coming from the remote config endpoint, and if it matches the version in the _embed_ configuration (i.e. the configuration the CLI worked with at the time of cutting the release), then we know the remote config is still safe to use. 

Otherwise (e.g. if `config_version` was updated to `2`, along with breaking changes in the config values), then we inform the user that the remote configuration no longer works with their current version of the CLI and they need to upgrade using `fastly update`.

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -82,7 +82,7 @@ func main() {
 	//
 	// To prevent issues we'll replace the config with what's embedded into the CLI,
 	// as we know that is compatible with the code currently being executed.
-	if err == config.ErrLegacyConfig || !file.ValidConfig() {
+	if err == config.ErrLegacyConfig || !file.ValidConfig(verboseOutput, out) {
 		err = file.UseStatic(cfg, config.FilePath)
 		if err != nil {
 			fsterrors.Deduce(err).Print(os.Stderr)
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -14,7 +14,6 @@ import (
 	"github.com/fastly/cli/pkg/check"
 	"github.com/fastly/cli/pkg/config"
 	fsterrors "github.com/fastly/cli/pkg/errors"
-	"github.com/fastly/cli/pkg/revision"
 	"github.com/fastly/cli/pkg/sync"
 	"github.com/fastly/cli/pkg/text"
 	"github.com/fastly/cli/pkg/update"
@@ -79,12 +78,12 @@ func main() {
 	// There are two scenarios we now want to look out for...
 	//
 	// 1. The config is using a legacy format.
-	// 2. The config is from an older CLI version.
+	// 2. The config is invalid (e.g. major version bump) for CLI version running.
 	//
 	// To prevent issues we'll replace the config with what's embedded into the CLI,
 	// as we know that is compatible with the code currently being executed.
-	if err == config.ErrLegacyConfig || file.CLI.Version != revision.SemVer(revision.AppVersion) {
-		err := file.UseStatic(cfg, config.FilePath)
+	if err == config.ErrLegacyConfig || !file.ValidConfig() {
+		err = file.UseStatic(cfg, config.FilePath)
 		if err != nil {
 			fsterrors.Deduce(err).Print(os.Stderr)
 			os.Exit(1)
--- a/pkg/compute/rust.go
+++ b/pkg/compute/rust.go
@@ -167,9 +167,9 @@ func (r *Rust) Verify(out io.Writer) error {
 	// We use rustup to assert that the toolchain is installed by streaming the output of
 	// `rustup toolchain list` and looking for a toolchain whose prefix matches our desired
 	// version.
-	fmt.Fprintf(out, "Checking if Rust %s is installed...\n", r.config.File.Language.Rust.ToolchainVersion)
+	fmt.Fprintf(out, "Checking if Rust %s is installed...\n", r.config.File.Language.Rust.ToolchainConstraint)
 
-	rustConstraint, err := semver.NewConstraint(r.config.File.Language.Rust.ToolchainVersion)
+	rustConstraint, err := semver.NewConstraint(r.config.File.Language.Rust.ToolchainConstraint)
 	if err != nil {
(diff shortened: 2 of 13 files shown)


Integralist:
@phamann as per our conversation earlier, some consideration needs to be given to configuration versioning (hence I've left this as a DRAFT PR).

cmd/fastly/main.go
```diff
@@ -79,12 +78,12 @@ func main() {
...
 	//
 	// 1. The config is using a legacy format.
-	// 2. The config is from an older CLI version.
+	// 2. The config is invalid (e.g. major version bump) for CLI version running.
 	//
 	// To prevent issues we'll replace the config with what's embedded into the CLI,
 	// as we know that is compatible with the code currently being executed.
-	if err == config.ErrLegacyConfig || file.CLI.Version != revision.SemVer(revision.AppVersion) {
-		err := file.UseStatic(cfg, config.FilePath)
+	if err == config.ErrLegacyConfig || !file.ValidConfig() {
```
Integralist, line 85: I'm wondering if we should still check explicitly for `ErrLegacyConfig`? 

The reason I'm asking is because the `ValidConfig` function will check the user's config for a `ConfigVersion` value (which in the case of a legacy config will be an empty string) and will compare it to the CLI's embedded config. 

In the case of a user having a 'legacy config', the user's config will be considered invalid and we'll use the embedded config instead. Thus the check for `ErrLegacyConfig` _before_ calling `ValidConfig()` is technically redundant.

The reason I left it in is because we still want to _return_ an `ErrLegacyConfig` type (earlier in the logic), and so it's kinda nice that we _explicitly_ highlight it as a scenario we want to be handling in this latter section of the code.

pkg/compute/rust.go
```diff
@@ -206,15 +195,15 @@ func (r Rust) Verify(out io.Writer) error {
 	// user should be the person making the local configuration change.
 	//
 	/* #nosec */
-	cmd = exec.Command("rustup", "target", "list", "--installed", "--toolchain", r.config.File.Language.Rust.ToolchainVersion)
+	cmd = exec.Command("rustup", "target", "list", "--installed", "--toolchain", r.toolchain.String())
```
kailan, line 198: Maybe we need to consider updating [this guide](https://developer.fastly.com/learning/compute/#install-local-dependencies) to explicitly state that you need `rustup` – installing Rust and Cargo manually will cause issues here.

> Compiling Rust applications for Compute@Edge requires Rust 1.49.0 and the wasm32-wasi toolchain. Install Rust and its dependencies so you can begin building Compute@Edge services:

kailan (approved):
Looks great! 🚀

pkg/compute/rust.go
Integralist: 👍🏻 yup, looks like that should be updated

kailan: Ticket opened ✨

pkg/compute/rust.go
```diff
@@ -383,6 +384,48 @@ func (r Rust) Build(out io.Writer, verbose bool) error {
...
+		if v[0] == "stable" {
+			continue
+		}
+
+		r.toolchain, err = semver.NewVersion(v[0])
+		if err != nil {
+			return fmt.Errorf("error parsing rust toolchain version: %w", err)
+		}
+
+		if ok := rustConstraint.Check(r.toolchain); ok {
```
Integralist, line 413: @kailan do you think we should go in _reverse_ for this logic? e.g. if my local rustup revealed...

```
stable-x86_64-apple-darwin (default)
1.49.0-x86_64-apple-darwin
1.50.0-x86_64-apple-darwin
```

...then it would mean that `1.49.0` would be used even though I have a newer rust version available (1.50.0) that's still within the supported range.

It's not massively important. It comes down to whether we feel it's worth pushing the user to the newest version of rust they happen to have available. So maybe something I revisit in a separate PR.

The downside though, is it would require more code to be written to support that selection process.

Thoughts?

kailan: I believe we should always use the latest version available. I could see confusion arise when a user installs a new Rust version with rustup, but then doesn't actually see their applications being compiled with it – especially because this selection behaviour isn't visible to the user.

Alternatively, make it clear to the user on each `compute build` which version of Rust is being used.

kailan: OR actually we could just let rustup decide which version to use and drop the `+1.49.0` flag from cargo commands. Then the behaviour can be configured using the `rustup default` command etc.

PR #336, opened 2021-07-07
Improve GitHub Actions intermittent test timeouts

Integralist:
## Problem

We've seen regular, but intermittent, test timeouts in the GitHub Actions CI.

The source of the problem appears to be due to `exec.Command`’s `Run` method (or at least the command we're passing to it, which I believe for the issues I've seen it occur would be a `cargo build` command).

```
github.com/fastly/cli/pkg/exec.Streaming.Exec(0x1ea1c8c, 0x5, 0xc0000ca090, 0x9, 0x9, 0xc000240c00, 0x5d, 0x5d, 0x0, 0xb9e7678, ...)
	/Users/runner/work/cli/cli/pkg/exec/exec.go:53 +0x2de
github.com/fastly/cli/pkg/compute.Rust.Build(0x2038ee0, 0xc000552390, 0xc0001ce000, 0xb9e7678, 0xc002c8b200, 0xc0004a2b00, 0x104dac5, 0x1dc0180)
	/Users/runner/work/cli/cli/pkg/compute/rust.go:356 +0x7d8
github.com/fastly/cli/pkg/compute.(*BuildCommand).Exec(0xc000504000, 0x20388a0, 0xc00011e140, 0x2037c40, 0xc000210708, 0x0, 0x0)
	/Users/runner/work/cli/cli/pkg/compute/build.go:174 +0xace
github.com/fastly/cli/pkg/compute.(*PublishCommand).Exec(0xc0001ce9c0, 0x20388a0, 0xc00011e140, 0x2037c40, 0xc000210708, 0x203d718, 0xc0001ce9c0)
	/Users/runner/work/cli/cli/pkg/compute/publish.go:86 +0x145
github.com/fastly/cli/pkg/compute_test.TestPublish(0xc0000a8300)
	/Users/runner/work/cli/cli/pkg/compute/compute_integration_test.go:1313 +0x170a
```

## Solution

> **NOTE**: These aren't _solutions_ in that they don't resolve the timeout issue. These are attempts at trying to help us better identify the root cause of the timeout when they happen.

- Update the `TestPublish` test code to log the stdout (this only prints when the test fails).
  - This is so when it does fail randomly again, we'll be able to see what was happening.
- Implement `exec.CommandContext` so we can set a timeout around the subprocess.
  - This helps us to short-circuit the test run (rather than waiting the full 10mins).

## Notes

I did some Googling and it seems there’s a few open issues related to this:

- https://github.com/golang/go/issues/38824
- https://github.com/golang/go/issues/38768

One suggestion I stumbled across was to set `GO_TEST_TIMEOUT_SCALE` as some architectures can have problems with the default value. But we've _NOT_ done that here as just increasing the timeout isn't really a solution and considering the entire test run normally only takes 4mins (which is well under the 10min limit) and the fact it only occurs occasionally suggests a problem we should properly identify.

Another suggestion was to split up code into more granular packages to avoid timeouts, but that's not an appropriate solution because the tests aren't timing out because of a code logic error but because the command itself being executed (`cargo build` in this case) is taking too long (well, as far as I'm concerned _something_ is stalling randomly at the point of running the command).

### DIFF
--- a/pkg/compute/publish.go
+++ b/pkg/compute/publish.go
@@ -28,6 +28,7 @@ type PublishCommand struct {
 	lang       cmd.OptionalString
 	includeSrc cmd.OptionalBool
 	force      cmd.OptionalBool
+	timeout    cmd.OptionalInt
 }
 
 // NewPublishCommand returns a usable command registered under the parent.
@@ -45,6 +46,7 @@ func NewPublishCommand(parent cmd.Registerer, globals *config.Data, build *Build
 	c.CmdClause.Flag("language", "Language type").Action(c.lang.Set).StringVar(&c.lang.Value)
 	c.CmdClause.Flag("include-source", "Include source code in built package").Action(c.includeSrc.Set).BoolVar(&c.includeSrc.Value)
 	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
+	c.CmdClause.Flag("timeout", "Numerical timeout, in minutes, for the build compilation step").Action(c.timeout.Set).IntVar(&c.timeout.Value)
 
 	// Deploy flags
 	c.RegisterServiceIDFlag(&c.manifest.Flag.ServiceID)
@@ -81,6 +83,9 @@ func (c *PublishCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	}
 	if c.force.WasSet {
 		c.build.Force = c.force.Value
+	}
+	if c.timeout.WasSet {
+		c.build.Timeout = c.timeout.Value
 	}
 
 	err = c.build.Exec(in, out)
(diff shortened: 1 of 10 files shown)


pkg/compute/publish.go
```diff
@@ -45,6 +46,7 @@ func NewPublishCommand(parent cmd.Registerer, globals *config.Data, build *Build
 	c.CmdClause.Flag("language", "Language type").Action(c.lang.Set).StringVar(&c.lang.Value)
 	c.CmdClause.Flag("include-source", "Include source code in built package").Action(c.includeSrc.Set).BoolVar(&c.includeSrc.Value)
 	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
+	c.CmdClause.Flag("timeout", "Numerical timeout, in minutes, for the build compilation step").Action(c.timeout.Set).IntVar(&c.timeout.Value)
```
triblondon, line 49:
```suggestion
	c.CmdClause.Flag("timeout", "Timeout, in seconds, for the build compilation step").Action(c.timeout.Set).IntVar(&c.timeout.Value)
```

triblondon (approved):
Approved on the basis of changing the units of the timeout from minutes to seconds.

PR #340, opened 2021-07-08
Log a record of errors to disk

Integralist:
To aid debugging, when running the CLI, track any errors in a log file.


pkg/errors/log.go
```diff
@@ -0,0 +1,99 @@
+package errors
```
Integralist, line 1: Here is where the core implementation is.

Integralist:
The new testdata files I'm using for the tests aren't comitted? So trying to figure out why.

triblondon (approved):

PR #343, opened 2021-07-09
Implement comment flag to compute deploy and publish (#328)

### LINKED ISSUE #328: fastly compute publish and deploy should accept a comment
**Version**

Fastly CLI version 0.32.0 (802bf7b)
Built with go version go1.16.5 linux/amd64

**What happened**

I leave a comment on every Fastly service version when working with VCL services to note what has changed in that version. It'd be great if we could use the same mechanism with C@E services.

Fastly service versions can have a comment:

* UI example: https://docs.fastly.com/en/guides/working-with-services#adding-comments-to-service-versions
* API example: https://developer.fastly.com/reference/api/services/version/
* `fastly service-version update` already supports `--comment` option: https://developer.fastly.com/reference/cli/service-version/update/

Perhaps`fastly compute publish` and `fastly compute deploy` could support a `--comment` option?

I don't think it makes sense to autogenerate a comment?

bbutkovic:
I added support for passing comments to compute deploy & publish. The command line for both commands will now accept a --comment flag that updates the new version with the desired comment. This closes #328.

### DIFF
--- a/pkg/compute/compute_deploy_test.go
+++ b/pkg/compute/compute_deploy_test.go
@@ -357,6 +357,27 @@ func TestDeploy(t *testing.T) {
 				"Deployed package (service 123, version 4)",
 			},
 		},
+		{
+			name: "success with comment",
+			args: args("compute deploy --token 123 -s 123 -p pkg/package.tar.gz --version 2 --comment foo"),
+			api: mock.API{
+				GetServiceFn:      getServiceOK,
+				ListVersionsFn:    testutil.ListVersions,
+				CloneVersionFn:    testutil.CloneVersionResult(3),
+				ListDomainsFn:     listDomainsOk,
+				ListBackendsFn:    listBackendsOk,
+				GetPackageFn:      getPackageOk,
+				UpdatePackageFn:   updatePackageOk,
+				ActivateVersionFn: activateVersionOk,
+				UpdateVersionFn:   updateVersionOk,
+			},
+			manifest: "name = \"package\"\nservice_id = \"123\"\n",
+			wantOutput: []string{
+				"Uploading package...",
+				"Activating version...",
+				"Deployed package (service 123, version 3)",
+			},
+		},
 	} {
 		t.Run(testcase.name, func(t *testing.T) {
 			// We're going to chdir to a deploy environment,
--- a/pkg/compute/compute_deploy_test.go
+++ b/pkg/compute/compute_deploy_test.go
@@ -363,7 +363,7 @@ func TestDeploy(t *testing.T) {
 			api: mock.API{
 				GetServiceFn:      getServiceOK,
 				ListVersionsFn:    testutil.ListVersions,
-				CloneVersionFn:    testutil.CloneVersionResult(3),
+				CloneVersionFn:    testutil.CloneVersionResult(4),
 				ListDomainsFn:     listDomainsOk,
 				ListBackendsFn:    listBackendsOk,
 				GetPackageFn:      getPackageOk,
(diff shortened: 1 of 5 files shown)


pkg/compute/compute_deploy_test.go
```diff
@@ -357,6 +357,27 @@ func TestDeploy(t *testing.T) {
 				"Deployed package (service 123, version 4)",
 			},
 		},
+		{
+			name: "success with comment",
+			args: args("compute deploy --token 123 -s 123 -p pkg/package.tar.gz --version 2 --comment foo"),
+			api: mock.API{
+				GetServiceFn:      getServiceOK,
+				ListVersionsFn:    testutil.ListVersions,
+				CloneVersionFn:    testutil.CloneVersionResult(3),
```
Integralist, line 366:
```suggestion
				CloneVersionFn:    testutil.CloneVersionResult(4),
```

Although, as far as mocking is concerned, it doesn't really matter what we return here, I would prefer the code to assert expectations that resemble the real life outcome. So, as `testutil.ListVersions` will return a list of three versions (where version 2 is 'locked') the real life outcome is that the locked version would be cloned and result in the output being: 

```
"Deployed package (service 123, version 4)"
```

pkg/compute/compute_deploy_test.go
```diff
@@ -357,6 +357,27 @@ func TestDeploy(t *testing.T) {
...
+				GetPackageFn:      getPackageOk,
+				UpdatePackageFn:   updatePackageOk,
+				ActivateVersionFn: activateVersionOk,
+				UpdateVersionFn:   updateVersionOk,
+			},
+			manifest: "name = \"package\"\nservice_id = \"123\"\n",
+			wantOutput: []string{
+				"Uploading package...",
+				"Activating version...",
+				"Deployed package (service 123, version 3)",
```
Integralist, line 378:
```suggestion
				"Deployed package (service 123, version 4)",
```

As per my above comment.

pkg/compute/deploy.go
```diff
@@ -291,6 +293,18 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
 
+	if c.Comment.WasSet {
+		_, err = c.Globals.Client.UpdateVersion(&fastly.UpdateVersionInput{
+			ServiceID:      serviceID,
+			ServiceVersion: version.Number,
+			Comment:        &c.Comment.Value,
+		})
+
+		if err != nil {
+			text.Error(out, "failed to set comment for version %d: %e", version.Number, err)
```
Integralist, line 304: If the use case for a comment is to annotate a release of a service/package (as per https://github.com/fastly/cli/issues/328) then I wonder if we should set the expectation to be that a failure to update the service with the given comment to result in the deploy flow to be stopped (i.e. don't then go ahead and activate the service).

/cc @acme what are your thoughts on this?

Integralist (commented):
Thanks @bbutkovic this looks good.

I've just a couple of minor comments.

pkg/compute/deploy.go
bbutkovic: I thought about this as well and ultimately decided that setting a comment wasn't critical, but that really depends on what makes most sense. It's no issue to change this to fail the flow instead of just logging the error.

pkg/compute/compute_deploy_test.go
bbutkovic: I agree, this was an oversight on my end.

pkg/compute/deploy.go
acme: Thanks for your contribution! I feel that if a comment has been passed as an option, then it is a critical part of the new service version, much like a version control commit message.

Integralist: Thanks @acme 

@bbutkovic if you wouldn't mind making those changes please, and then I'll re-review. Many thanks!

bbutkovic: Makes sense, thank you for the input! I'll make the changes soon.

Integralist (approved):

PR #344, opened 2021-07-14
Add an --addr flag to `compute serve` 

phamann:
Per user testing feedback from @dkegel-fastly; this adds an `--addr` flag to `compute serve` which allows a user to configure the IPv4 address which Viceroy listens on. This mirrors the flag exposed by Viceroy. 

### Note:
⚠️ This is a PR into @Integralist's working branch and not `main`.

### DIFF
--- a/pkg/compute/serve.go
+++ b/pkg/compute/serve.go
@@ -25,6 +25,7 @@ import (
 type ServeCommand struct {
 	cmd.Base
 
+	addr             string
 	build            *BuildCommand
 	env              cmd.OptionalString
 	file             string
@@ -50,6 +51,7 @@ func NewServeCommand(parent cmd.Registerer, globals *config.Data, build *BuildCo
 	c.manifest.File.SetOutput(c.Globals.Output)
 	c.manifest.File.Read(manifest.Filename)
 
+	c.CmdClause.Flag("addr", "The IPv4 address to listen on").Default("127.0.0.1:7676").StringVar(&c.addr)
 	c.CmdClause.Flag("env", "The environment configuration to use (e.g. stage)").Action(c.env.Set).StringVar(&c.env.Value)
 	c.CmdClause.Flag("file", "The Wasm file to run").Default("bin/main.wasm").StringVar(&c.file)
 	c.CmdClause.Flag("force", "Skip verification steps and force build").Action(c.force.Set).BoolVar(&c.force.Value)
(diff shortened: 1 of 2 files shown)


pkg/compute/serve.go
```diff
@@ -28,6 +28,7 @@ type ServeCommand struct {
 	build            *BuildCommand
 	env              cmd.OptionalString
 	file             string
+	addr             string
```
Integralist, line 31: Similarly I've been sorting keys (separate from any embedded structs, e.g. `cmd.Base`) as it makes it easier to find things.

pkg/compute/serve.go
```diff
@@ -52,6 +53,7 @@ func NewServeCommand(parent cmd.Registerer, globals *config.Data, build *BuildCo
 
 	c.CmdClause.Flag("env", "The environment configuration to use (e.g. stage)").Action(c.env.Set).StringVar(&c.env.Value)
 	c.CmdClause.Flag("file", "The Wasm file to run").Default("bin/main.wasm").StringVar(&c.file)
+	c.CmdClause.Flag("addr", "The IPv4 address to listen on").Default("127.0.0.1:7676").StringVar(&c.addr)
```
Integralist, line 56: Do we really want to couple the CLI to Viceroy like this? If there's a new release of Viceroy that uses a different port for example, then it means we'd need to release a new CLI version. Might be best to only set the `--addr` flag if a value is provided rather than providing a default.

Integralist:
Looks like [@dkegel-fastly's patch](https://github.com/fastly/cli/pull/252#issuecomment-879482413) follows with my own review comments here.

pkg/compute/serve.go
phamann: Yeah I did spend some time thinking about this. As we always pass this arg (with default value or user provded one) Its better to think about it that we're explicitly controlling the default value from the CLI and that is independent over whatever Viceroy decides to use as its default. I.e. if Viceroy updates their default, the CLI doesn't need to change as its always passing 7878 explicitly.

Per my comment on your PR, I "personally" prefer to use flag defaults as it then generates the documentation so users can discover what the default is rather than having magic defaults.

Integralist: > Its better to think about it that we're explicitly controlling the default value from the CLI and that value is independent from whatever Viceroy decides to use as its default. I.e. if Viceroy updates their default, the CLI doesn't need to change as its always passing 7878 explicitly.

OK this makes sense 👍🏻

Integralist (approved):
LGTM and agree with your feedback.

PR #347, opened 2021-07-20
Support for beta JS SDK

Integralist:
As part of this PR I moved around some existing test files so they were placed inside language specific sub directories. Previously Rust files and AssemblyScript files were in the same directory, which was not only confusing but also problematic when we introduced JavaScript as it uses the same toolchain files (e.g. `package.json` etc).


pkg/compute/compute_build_test.go
```diff
@@ -382,8 +382,8 @@ func TestBuildAssemblyScript(t *testing.T) {
 			fastlyManifest: `
 			manifest_version = 1
 			name = "test"
-			language = "javascript"`,
-			wantError: "unsupported language javascript",
```
Integralist, line 386: lulz

pkg/compute/compute_build_test.go
```diff
@@ -407,8 +407,97 @@ func TestBuildAssemblyScript(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// Create test environment
+			rootdir := testutil.NewEnv(testutil.EnvOpts{
+				T: t,
+				Copy: []testutil.FileIO{
+					{Src: filepath.Join("testdata", "build", "javascript", "package.json"), Dst: "package.json"},
+					{Src: filepath.Join("testdata", "build", "javascript", "fastly-js-compute-0.0.1.tgz"), Dst: "fastly-js-compute-0.0.1.tgz"},
```
Integralist, line 499: This file is temporary until the SDK repo is made public and we can rely on installing the binary from npm.

triblondon (approved):
Nits:

* Spelling "dont't"
* Remove package-lock.json files if possible (they don't seem to be used)

pkg/commands/compute/language_javascript.go
```diff
@@ -0,0 +1,190 @@
...
+
+	p, err = getNpmBinPath()
+	if err != nil {
+		return errors.RemediationError{
+			Inner:       fmt.Errorf("could not determine npm bin path"),
+			Remediation: fmt.Sprintf("To fix this error, run the following command:\n\n\t$ %s", text.Bold("npm install --global npm@latest")),
+		}
+	}
+
+	path, err := exec.LookPath(filepath.Join(p, "ecp-js"))
```
tschneidereit, line 136: `ecp-js` was renamed to `js-compute-runtime`, so this and the other mentions should be updated accordingly

pkg/commands/compute/language_javascript.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	fmt.Fprintf(out, "Found package.json at %s\n", fpath)
+
+	// 3) Check if `ecp-js` is installed.
+	//
+	// ecp-js is the JavaScript compiler. We first check if the required
+	// dependency exists in the package.json and then whether the ecp-js binary
+	// exists in the npm bin directory.
```
tschneidereit: Tiny nit: this is now called `js-compute-runtime`, but I guess we're also not explcitly testing for it anymore?

PR #350, opened 2021-07-22
Implement `acl` command

Integralist:
**NOTE**: As part of this PR I've re-ordered the commands to be alphabetically ordered (as it makes more sense from a user perspective and also easier to manage from a code maintenance perspective). This has meant a large number of line changes in `pkg/app/run.go` and in its test file `pkg/app/run_test.go` but no actual logic changes, so those two files are what are driving up the change set size.


kailan (approved):
This works perfectly, and testing is extensive. Ship it 🚀 

Am I right in thinking ACL entries will be the next command?

Issue #241, opened 2021-03-30
fastly compute init should use SNI and Override host for origins


acme:
Would it be possible for `fastly compute init` when adding an origin over TLS to use the hostname for `SNI hostname` and `Override host` too? All the origins I talk to require SNI and to pass the real host, not the funny-three-words host, so I then have to go into the UI to fix it.

fgsch:
I've run into this problem today while testing a C@E service created with the CLI. 
Adding the backend through the UI did set the `SNI hostname` (and `Certificate hostname`) correctly.

fgsch:
@acme would #352 do the job for your use case?

acme:
Thanks, this now works great for me! 👏

PR #352, opened 2021-07-23
Implement --override-host and --ssl-sni-hostname in compute deploy & publish

fgsch:
Related and partially addresses #241.


Integralist:
@fgsch I noticed this same test failing in one of my own PRs recently and I think this should fix it...

```diff
diff --git a/pkg/commands/acl/acl_test.go b/pkg/commands/acl/acl_test.go
index 34430dde..0c1f453b 100644
--- a/pkg/commands/acl/acl_test.go
+++ b/pkg/commands/acl/acl_test.go
@@ -299,7 +299,7 @@ func TestACLUpdate(t *testing.T) {
        scenarios := []testutil.TestScenario{
                {
                        Name:      "validate missing --name flag",
-                       Args:      args("acl update --version 3"),
+                       Args:      args("acl update --new-name beepboop --version 3"),
                        WantError: "error parsing arguments: required flag --name not provided",
                },
                {
```

I _think_ Kingpin might be parsing the flags in a nondeterministic fashion and so sometimes the test fails because it doesn't fail on the first flag defined (`--name`), so by ensuring we set the other required flag `--new-name` we can be sure the test will fail only on the flag we are expecting it to.

I'm going to put this change into a separate PR now and I'll ping it to you so you can review it, then that can be rebased into your branch.

Integralist:
https://github.com/fastly/cli/pull/353

pkg/commands/compute/compute_deploy_test.go
```diff
@@ -378,6 +378,62 @@ func TestDeploy(t *testing.T) {
...
+			api: mock.API{
+				CreateServiceFn:   createServiceOK,
+				GetPackageFn:      getPackageOk,
+				UpdatePackageFn:   updatePackageOk,
+				CreateDomainFn:    createDomainOK,
+				CreateBackendFn:   createBackendExpect("host.com", 80, "", ""),
+				ActivateVersionFn: activateVersionOk,
+				ListDomainsFn:     listDomainsOk,
+			},
+			manifest: "name = \"package\"\n",
```
Integralist, line 393: NIT: I'd prefer to use back ticks so we don't have to escape the double quotes (repeat this suggestion throughout).

```suggestion
			manifest: `name = "package"\n`,
```

pkg/commands/compute/compute_deploy_test.go
```diff
@@ -378,6 +378,62 @@ func TestDeploy(t *testing.T) {
 				"Deployed package (service 123, version 4)",
 			},
 		},
+		{
+			name: "success with --backend and no --backend-port",
+			args: args("compute deploy --token 123 --backend host.com"),
```
Integralist, line 383: NIT: I prefer to order flags alphabetically (even in test code) just so it's easier to scan for flags (repeat this suggestion throughout).

```suggestion
			args: args("compute deploy --backend host.com --token 123"),
```

pkg/commands/compute/compute_mocks_test.go
```diff
@@ -97,3 +97,16 @@ func (v versionClient) Do(req *http.Request) (*http.Response, error) {
 	}
 	return rec.Result(), nil
 }
+
+func createBackendExpect(address string, port uint, overrideHost string, sslSNIHostname string) func(*fastly.CreateBackendInput) (*fastly.Backend, error) {
```
Integralist, line 101: NIT: I'd prefer to have this code inside of `compute_deploy_test.go` until such a time that the specific behaviour is necessary to put here for sharing across other test code.

pkg/commands/compute/deploy.go
```diff
@@ -44,12 +44,19 @@ type DeployCommand struct {
...
 	Comment        cmd.OptionalString
 	ServiceVersion cmd.OptionalServiceVersion
 }
 
+// Backend represents the configuration parameters for a backend
+type Backend struct {
+	Address        string
+	Port           uint
+	OverrideHost   string
+	SSLSNIHostname string
```
Integralist, line 57: NIT: I prefer to order struct fields:

```suggestion
	Address        string
	OverrideHost   string
	Port           uint
	SSLSNIHostname string
```

pkg/commands/compute/deploy.go
```diff
@@ -68,8 +75,10 @@ func NewDeployCommand(parent cmd.Registerer, client api.HTTPClient, globals *con
 	})
 	c.CmdClause.Flag("path", "Path to package").Short('p').StringVar(&c.Path)
 	c.CmdClause.Flag("domain", "The name of the domain associated to the package").StringVar(&c.Domain)
-	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.Backend)
-	c.CmdClause.Flag("backend-port", "A port number for the package backend").UintVar(&c.BackendPort)
+	c.CmdClause.Flag("backend", "A hostname, IPv4, or IPv6 address for the package backend").StringVar(&c.Backend.Address)
+	c.CmdClause.Flag("backend-port", "A port number for the package backend").UintVar(&c.Backend.Port)
+	c.CmdClause.Flag("override-host", "The hostname to override the Host header").StringVar(&c.Backend.OverrideHost)
+	c.CmdClause.Flag("ssl-sni-hostname", "The hostname to use at the start of the TLS handshake").StringVar(&c.Backend.SSLSNIHostname)
```
Integralist, line 81: NIT: With new commands I've added I've been sorting the flags into alphabetical order, so might be worth doing that while we're here.

pkg/commands/compute/publish.go
```diff
@@ -21,6 +21,8 @@ type PublishCommand struct {
 	domain         cmd.OptionalString
 	backend        cmd.OptionalString
 	backendPort    cmd.OptionalUint
+	overrideHost   cmd.OptionalString
+	sslSNIHostname cmd.OptionalString
```
Integralist, line 25: NIT: I'd prefer to order these Deploy fields.

Integralist (commented):
Thanks @fgsch I just had a couple of small NIT comments, but otherwise this LGTM (and the new `Backend` struct type helps clean up the code nicely 👍🏻).

It's probably worth, for posterity, to clarify in a comment here why we're implementing this in the flag configuration only (i.e. no interactive prompts), which is because both the override host and sni hostname settings are optional, and there's no satisfactory way for the user to indicate that they neither want the options set nor do they want the defaults. For example, if they just press ENTER, then they'll get the defaults and they might want nothing set at all. 

We could codify a pattern such as entering NONE means neither the default value, nor the configuration itself is set. But I don't know if these settings are common enough to justify adding that extra logic (we can always go back and add this behaviour if we find users would prefer to be prompted).

pkg/commands/compute/publish.go
fgsch: Done.

pkg/commands/compute/deploy.go
fgsch: Done.

pkg/commands/compute/deploy.go
fgsch: Done.

pkg/commands/compute/compute_mocks_test.go
fgsch: Done.

pkg/commands/compute/compute_deploy_test.go
fgsch: Done.

pkg/commands/compute/compute_deploy_test.go
fgsch: Done.

Integralist (approved):

PR #360, opened 2021-07-30
Refactor code to help CI performance

Integralist:
## Problem

The tests time out constantly, and they need to be rerun until they _don't_ time out (and subsequently pass). 

Looking at the time it takes to run the tests (and remember we have a 10 minute timeout _per 'package'_), all packages take around 3 seconds to run (on a cold start average) with the exception of the 'compute' package which takes 9-10 mins (which is the sole reason why the tests time out, and practically all of that time is spent on `TestPublish` in particular).

## Changes

Things done in this PR that are directly related to performance...

1. Reduce the amount of IO when setting up tests (e.g. create test environment once per test).
2. Cache rust resources.
3. Remove `TestPublish` test.

## Notes

Bullet point `1.` made little to no difference in the overall runtime of tests, but was still worth doing for the sake of cleaning up some of the test code.

Bullet point `2.` also made little to no difference in the overall runtime of the tests, and this seems to be because the compute tests are primarily integration tests and they validate multiple scenarios where we provide mocked lock files for simulating different dependency conditions and how the code should behave when getting an unrecognised dependency. Meaning the effectiveness of caching a static set of resource doesn't help the runtime of those tests.

Bullet point `3.` I'm doing being the value of this test doesn't match the time overhead involved with having to re-run the CI tests over and over whenever they timeout (which happens multiple times per PR). When looking at the test it simply validates that calling `compute publish` (with variations of the available flags exposed) will show the `build` and `deploy` commands were executed successfully (of which all API calls are mocked to succeed). This is a nice test to have, but not at the expense of hours (†) of wasted time.

> † We have three platforms we run our tests on. The tests (as a whole) take around 15mins to run, and considering I have to re-run the CI tests any where between 4-6 times per PR open to get the timeout to not trigger, means a lot of wasted time.

I've tried rewriting `TestPublish` into a unit test (e.g. mock the compute command and pass in mocked build/deploy commands) but this didn't work because of the CLI design, and it would have meant adding logic to the business logic just for the sake of the test validation (e.g. "if this is a test, then ignore this error here and continue onto the next line").

By deleting `TestPublish` I consistently cut the compute package test run in half! We could have kept the test, but it would have meant pushing it into its own package so it didn't fall under the same 10min limit of the 'compute' package. But this approach is frowned upon generally by the Go community (i.e. you shouldn't need to restructure your packages to fix a slow test).


pkg/commands/compute/compute_publish_test.go
```diff
@@ -16,242 +15,137 @@ import (
 )
 
 func TestPublish(t *testing.T) {
+	// We're going to chdir to a temp environment,
```
Integralist, line 18: The changes to this function can be summarised as follows...

1. Deduplicate a lot of the test configuration.
2. Define the test configuration outside of the test harness.
3. Define a single temp environment rather than having a new environment for each test scenario.

pkg/commands/compute/compute_unit_test.go
```diff
@@ -83,78 +83,54 @@ func getFlags(cmd *kingpin.CmdClause) reflect.Value {
 }
 
 func TestCreatePackageArchive(t *testing.T) {
-	for _, testcase := range []struct {
```
Integralist, line 86: The changes to this function can be summarised as follows...

Remove the 'table test' structure, as it's redundant for a single test scenario. It just makes the code harder to read.

pkg/commands/compute/compute_unit_test.go
```diff
@@ -183,6 +159,32 @@ func TestFileNameWithoutExtension(t *testing.T) {
 }
 
 func TestGetIgnoredFiles(t *testing.T) {
+	// we're going to chdir to a build environment,
```
Integralist, line 162: The changes to this function can be summarised as follows...

1. Define a single temp environment rather than having a new environment for each test scenario.
2. For scenario specific files, create those within the temp environment separately.

pkg/commands/compute/compute_unit_test.go
```diff
@@ -252,6 +228,32 @@ func TestGetIgnoredFiles(t *testing.T) {
 }
 
 func TestGetNonIgnoredFiles(t *testing.T) {
+	// We're going to chdir to a build environment,
```
Integralist, line 231: The changes to this function can be summarised as follows...

Define a single temp environment rather than having a new environment for each test scenario.

pkg/testutil/env.go
```diff
@@ -23,8 +23,6 @@ type EnvOpts struct {
 
 // NewEnv creates a new test environment and returns the root directory.
 func NewEnv(opts EnvOpts) (rootdir string) {
-	opts.T.Helper()
```
Integralist, line 26: A `Helper` marks the calling function as a test helper function so, when printing file and line information, that function will be skipped. I decided I wanted to see that output.

.github/workflows/pr_test.yml
```diff
@@ -4,48 +4,50 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout code
+    - name: "Checkout code"
```
Integralist, line 7: The changes to this file can be summarised as follows...

1. Add `name` field if missing.
2. Reword some `name` fields.
3. Tweak Golang caching.
4. Add Rust caching.

pkg/commands/compute/build_test.go
```diff
@@ -444,6 +444,33 @@ func TestBuildJavaScript(t *testing.T) {
 		t.Skip("Set TEST_COMPUTE_BUILD_JAVASCRIPT or TEST_COMPUTE_BUILD to run this test")
 	}
 
+	// We're going to chdir to a build environment,
```
Integralist, line 447: The changes to this file can be summarised as follows...

Define a single temp environment rather than having a new environment for each test scenario.

pkg/commands/compute/compute_test.go
```diff
@@ -83,78 +83,54 @@ func getFlags(cmd *kingpin.CmdClause) reflect.Value {
 }
 
 func TestCreatePackageArchive(t *testing.T) {
-	for _, testcase := range []struct {
```
Integralist, line 86: The changes to this file can be summarised as follows...

Remove the 'table test' structure as it was redundant (there was only ever one scenario).

pkg/commands/compute/compute_test.go
```diff
@@ -183,6 +159,32 @@ func TestFileNameWithoutExtension(t *testing.T) {
 }
 
 func TestGetIgnoredFiles(t *testing.T) {
+	// we're going to chdir to a build environment,
```
Integralist, line 162: The changes to this function can be summarised as follows...

1. Define a single temp environment rather than having a new environment for each test scenario.
2. Fix file path/name ambiguity that could fail on Windows if not careful.

pkg/commands/compute/compute_test.go
```diff
@@ -252,6 +228,32 @@ func TestGetIgnoredFiles(t *testing.T) {
 }
 
 func TestGetNonIgnoredFiles(t *testing.T) {
+	// We're going to chdir to a build environment,
```
Integralist, line 231: The changes to this function can be summarised as follows...

Define a single temp environment rather than having a new environment for each test scenario.

pkg/commands/compute/deploy_test.go
```diff
@@ -16,25 +16,59 @@ import (
 )
 
 func TestDeploy(t *testing.T) {
+	// NOTE: Some tests don't provide a Service ID via any mechanism (e.g. flag
```
Integralist, line 19: The changes to this function can be summarised as follows...

1. Define a single temp environment rather than having a new environment for each test scenario.
2. Replace short flag `-s` with the long flag `--service-id`.
3. Use flag for setting Service ID and not the manifest (makes bullet point 1 easier).
4. Recreate manifest for each scenario, as the mutation of the manifest affects the test expectations.

triblondon (approved):
👍

Issue #363, opened 2021-08-04
Rustup toolchain check prevents deploy if newest installed toolchain is incompatible


ccampbell:
**Version**

```
Fastly CLI version 0.36.0 (dd0b7d9)
Built with go version go1.16.6 linux/amd64
```

**What happened**

I recently updated my project’s rust-toolchain to use 1.54.0 which is [not currently supported in the CLI](https://github.com/fastly/cli/blob/dd0b7d95750beb23e735151f73945e69660dbff1/cmd/fastly/static/config.toml#L13). It returns the following error which is expected:
> ERROR: rust toolchain 1.54.0 is incompatible with the constraint >= 1.49.0 < 1.54.0

However, the issue is that even after changing my local `rust-toolchain` file to use 1.53.0 which _is_ compatible, the deployment still fails because the [check for the toolchain version](https://github.com/fastly/cli/blob/28370af78585c98d884dd78162a8eb697c4e59b0/pkg/commands/compute/language_rust.go#L410) explicitly uses the last version you have installed locally based on `rustup toolchain list`. For me the list (after updating `rust-toolchain`) is:
```
1.43.0-x86_64-apple-darwin
1.46.0-x86_64-apple-darwin
1.49.0-x86_64-apple-darwin
1.51.0-x86_64-apple-darwin (default)
1.53.0-x86_64-apple-darwin (override)
1.54.0-x86_64-apple-darwin
```

I think it may make sense to use `rustup show active-toolchain` to pull in the toolchain version since that one should respect the one actually being used by the project. Output of that command is:

```
1.53.0-x86_64-apple-darwin (overridden by '/path/to/project/rust-toolchain')
```

berniechiu:
Just encountered this issue when we have a CI build, but we've only installed a single version, using the `rustup toolchain list` shouldn't matter

This is our example CI
```yaml
env:
  CARGO_TERM_COLOR: always
  FASTLY_CLI_VERSION: v0.27.2

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - uses: actions-rs/toolchain@v1.0.6
      with:
        toolchain: 1.49.0
        target: wasm32-wasi
...
...
    - name: Build wasm package
      run: ./fastly compute build
      env:
        FASTLY_API_TOKEN: ${{ secrets.FASTLY_API_TOKEN }
```

[...]

PR #365, opened 2021-08-09
Respect active rust toolchain instead of the latest version

i110:
The current implementation determines rust version by executing `rustup toolchain list` and pick the last line as the latest version, but should it use the active one?

I updated fastly cli and it started to fail with the following error (even though my rust-toochain is `1.49.0`)
```
$ fastly compute build
(snip)
Checking if Rust >= 1.49.0 < 1.54.0 is installed...

ERROR: rust toolchain 1.54.0 is incompatible with the constraint >= 1.49.0 < 1.54.0.

To fix this error, run the following command with a version within the given range >= 1.49.0 < 1.54.0:

        $ rustup toolchain install <version>
```

### DIFF
--- a/pkg/commands/compute/build_test.go
+++ b/pkg/commands/compute/build_test.go
@@ -213,7 +213,7 @@ func TestBuildRust(t *testing.T) {
 			client: versionClient{
 				fastlyVersions: []string{"0.6.0"},
 			},
-			wantError:            "rust toolchain 1.49.0 is incompatible with the constraint >= 1.0.0 < 1.40.0",
+			wantError:            "rust 1.49.0 is incompatible with the constraint >= 1.0.0 < 1.40.0",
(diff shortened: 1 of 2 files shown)


i110:
This fixes https://github.com/fastly/cli/issues/363

i110:
CI failed, and seeing the FAIL logs I noticed that we should separately handle `toolchain name` (e.g. `stable`) and `rustc version` (e.g. `1.49.0`). Addressed in https://github.com/fastly/cli/pull/365/commits/c465b48d1189f43d7939eb2af72e92d62877df45

Integralist:
Thanks for taking the time to work on this issue and to open this PR. 

Unfortunately there were additional design and implementation changes (discussed internally) which we decided to implement in a separate PR, and so we'll close this PR in favour of https://github.com/fastly/cli/pull/371. 

Thank you again for your contribution, it was appreciated.

Issue #364, opened 2021-08-06
Unexpected "permission denied" while running "fastly compute serve"


dnechay:
Hey
Playing around with your nice Compute@Edge engine and faced the next issue, any help is appreciated.

### Environment
```
NAME="Linux Mint"
VERSION="20.2 (Uma)"
ID=linuxmint
ID_LIKE=ubuntu
PRETTY_NAME="Linux Mint 20.2"
VERSION_ID="20.2"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.linuxmint.com/"
BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
PRIVACY_POLICY_URL="https://www.linuxmint.com/"
VERSION_CODENAME=uma
UBUNTU_CODENAME=focal
```

### Fastly CLI Version

Fastly CLI version 0.36.0 (dd0b7d9)
Built with go version go1.16.6 linux/amd64

### What happened

1. `fastly compute init` (for javascript and from https://github.com/fastly/compute-starter-kit-javascript-default)
2. `fastly compute serve`

**Expected behavior**
Get the local dev env running

**Actual Behavior**
Successfully built but weren't able to start `viceroy`
![Screenshot from 2021-08-06 12-31-29](https://user-images.githubusercontent.com/15381780/128490563-250f65fd-dc1e-4b1c-a99a-fc2931f17cba.png)

**Steps taken**
- tried to reinstall CLI and run commands one more time - no results
- tried to run `serve` command with `sudo`, but locally I use `nvm` to manage Node.js installations, so under `sudo` there is no `npm` and `fastly` isn't able to build package; also IMO `serve` command should not require any superuser permission

kailan:
Hey @dnechay, thanks for reporting!

There is a known workaround for this: `chmod +x ~/.config/fastly/viceroy`

I'll leave the issue open to track this until a proper fix next week 👍

dnechay:
> chmod +x ~/.config/fastly/viceroy

Thanks @kailan , this workaround works for me!

Integralist:
Just a quick update to state that it looks like we already attempt to set the permissions on the downloaded binary:

https://github.com/fastly/cli/blob/3209162a6f8e689766d8197e6bcc4fb785ce8dfa/pkg/commands/update/versioner.go#L190-L193

[...]

PR #368, opened 2021-08-11
Add executable permissions to Viceroy binary AFTER moving it

Integralist:
Fixes https://github.com/fastly/cli/issues/364

Executable permissions are reset on the Viceroy binary file once it has been moved from its initial tmp directory to the intended destination directory.

### DIFF
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -168,6 +168,16 @@ func getViceroy(progress text.Progress, out io.Writer, versioner update.Versione
 		}
 	}
 
+	// G302 (CWE-276): Expect file permissions to be 0600 or less
+	// gosec flagged this:
+	// Disabling as the file was not executable without it and we need all users
+	// to be able to execute the binary.
+	/* #nosec */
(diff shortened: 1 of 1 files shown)


kailan (approved):
Good find!

pkg/commands/compute/serve.go
```diff
@@ -168,6 +168,16 @@ func getViceroy(progress text.Progress, out io.Writer, versioner update.Versione
 		}
 	}
 
+	// G302 (CWE-276): Expect file permissions to be 0600 or less
+	// gosec flagged this:
+	// Disabling as the file was not executable without it and we need all users
+	// to be able to execute the binary.
+	/* #nosec */
+	err = os.Chmod(bin, 0777)
```
fgsch, line 176: Why not 0555?

Integralist: For now I'm just keeping it consistent to what we've used elsewhere. I'll make a note to revisit this as 555 would probably be fine.

Integralist:
The customer has tested this code and validated it has resolved their issue.

PR #370, opened 2021-08-12
Refactor `compute deploy` logic to better support [setup] configuration

Integralist:
This is a large refactor of the existing `compute deploy` logic.

The original logic had some issues...

1. It wasn't designed for supporting more than a few types of `[setup]` fields, so when looking to introduce extra fields like ACL, Dictionaries, Logging Endpoints (and potentially more in future) I realised the logic needed to be rewritten.

2. The logic that was checking which backends within the `[setup]` configuration were missing was flawed and would cause the same backend to be added multiple times. The bug was only surfaced upon refactoring the existing logic.

3. The specific user flow we had meant we needed to duplicate a bunch of calls to the same functions. For example, the functions `cfgDomain ` and `configureBackends` would need to be called _three_ separate times each, where as now they're only called once. This was caused primarily by the design of the original `validateService` function and because we would have a separate "new service" flow that was subsequently needed to be uniquely separate from the "existing service" flow.

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -235,6 +235,14 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		}
 	}
 
+	// RESOURCE CREATION...
+	//
+	// We can't create resources earlier in the logic flow as it requires the use
+	// of a text.Progress which overwrites the current line (i.e. it would cause
+	// any text prompts to be hidden) and so we prompt for as much information as
+	// possible at the top of the Exec function. After we have all the information,
+	// then we proceed with the creation of resources.
+
 	var (
 		progress text.Progress
 		desc     string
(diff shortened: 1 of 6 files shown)


pkg/commands/compute/deploy.go
```diff
@@ -658,49 +463,181 @@ func updateManifestServiceID(m *manifest.File, manifestFilename string, progress
...
 }
 
-// pkgCompare compares the local package hashsum against the existing service
-// package version and exits early with message if identical.
-func pkgCompare(client api.Interface, serviceID string, version int, path string, progress text.Progress, out io.Writer) (bool, error) {
-	p, err := client.GetPackage(&fastly.GetPackageInput{
-		ServiceID:      serviceID,
-		ServiceVersion: version,
-	})
+// manageExistingServiceFlow handles BLAH.
```
kailan, line 466: BLAH

pkg/commands/compute/deploy.go
```diff
@@ -91,330 +84,215 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	}
 
-		if !ok {
-			invalidService = true
+	// RESOURCE CONFIGURATION...
 
-			text.Output(out, "Service '%s' is missing required domain or backend(s). These must be added before the Compute@Edge service can be deployed.", serviceID)
+	if !hasDomain || !hasRequiredBackends {
+		if !c.AcceptDefaults {
+			text.Output(out, "Service '%s' is missing required resources. These must be added before the Compute@Edge service can be deployed. Please follow the prompts.", serviceID)
```
kailan, line 179: If initialising a new service and then changing the host of a backend on manage.fastly.com this error will appear on subsequent deploys - suggest modifying fastly.toml file

kailan (approved):
🚀

PR #371, opened 2021-08-12
Update rust toolchain validation steps

Integralist:
Fixes https://github.com/fastly/cli/issues/363

There were issues with the original logic that used `rustup toolchain list` as well as using `rustup show active-toolchain` (as that only revealed, for example, `stable-x86_64-...` and so wasn't useful for checking against a semver constraint such as `>= 1.54.0`).

**The new validation steps are**:

1. Validate `rustc` installed.
2. Validate `rustc --version` meets the constraint.
3. Validate `wasm32-wasi` target is added to the relevant toolchain.
4. Validate `cargo` is installed.
5. Validate `fastly-sys` crate version.
6. Validate `fastly` crate version (optional upgrade suggestion).

This PR also need to be coordinated around a change to the CLI config so that the starter kits it references use the `main` branch for each of the Rust starter kit (rather than using various tags or non-main branches). This is because the latest release of each Rust starter kit is now using a `stable` channel:

- https://github.com/fastly/compute-starter-kit-rust-default/pull/44
- https://github.com/fastly/compute-starter-kit-rust-beacon-termination/pull/15
- https://github.com/fastly/compute-starter-kit-rust-static-content/pull/11

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -55,7 +55,7 @@ jobs:
       matrix:
         go-version: [1.17.x]
         node-version: [12]
-        rust-toolchain: [1.54.0]
+        rust-toolchain: [stable]
         platform: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.x
+        go-version: 1.17.x
     - name: "Restore golang bin cache"
       id: go-bin-cache
       uses: actions/cache@v2
@@ -53,7 +53,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.16.x]
+        go-version: [1.17.x]
         node-version: [12]
         rust-toolchain: [1.54.0]
         platform: [ubuntu-latest, macos-latest, windows-latest]
@@ -85,6 +85,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         toolchain: ${{ matrix.rust-toolchain }}
+        override: true
     - name: "Add wasm32-wasi Rust target"
       run: rustup target add wasm32-wasi --toolchain ${{ matrix.rust-toolchain }}
     - name: "Restore cargo cache"
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -85,20 +85,11 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         toolchain: ${{ matrix.rust-toolchain }}
-        override: true
     - name: "Add wasm32-wasi Rust target"
       run: rustup target add wasm32-wasi --toolchain ${{ matrix.rust-toolchain }}
-    - name: "Restore cargo cache"
-      uses: actions/cache@v2
-      with:
-        path: |
-          ~/.rustup/
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    - name: "Validate Rust toolchain"
+      run: rustup show && rustup target list --installed --toolchain stable
+      shell: bash
     - name: "Install Node"
       uses: actions/setup-node@v2
       with:
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -112,3 +112,4 @@ jobs:
       env:
         TEST_COMPUTE_INIT: true
         TEST_COMPUTE_BUILD: true
+        TEST_COMPUTE_DEPLOY: true
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -106,6 +106,9 @@ jobs:
     - name: "Download latest app config"
       run: |
         make config
+    - name: "Validate rust environment"
+      run: rustup show && rustc --print sysroot
+      shell: bash
     - name: "Test suite"
       run: make test
       shell: bash
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -106,9 +106,6 @@ jobs:
     - name: "Download latest app config"
       run: |
         make config
-    - name: "Validate rust environment"
-      run: rustup show && rustc --print sysroot
-      shell: bash
     - name: "Test suite"
       run: make test
       shell: bash
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 1. Merge all PRs intended for the release.
 2. Rebase latest remote main branch locally (`git pull --rebase origin main`).
-3. Ensure all analysis checks and tests are passing (`time TEST_COMPUTE_BUILD=1 TEST_COMPUTE_INIT=1 make all`).
+3. Ensure all analysis checks and tests are passing (`time TEST_COMPUTE_INIT=1 TEST_COMPUTE_BUILD=1 TEST_COMPUTE_DEPLOY=1 make all`).
 4. Open a new PR to update CHANGELOG ([example](https://github.com/fastly/cli/pull/273))<sup>[1](#note1)</sup>.
 5. Merge CHANGELOG.
 6. Rebase latest remote main branch locally (`git pull --rebase origin main`).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 1. Merge all PRs intended for the release.
 2. Rebase latest remote main branch locally (`git pull --rebase origin main`).
-3. Ensure all analysis checks and tests are passing (`make all`).
+3. Ensure all analysis checks and tests are passing (`time TEST_COMPUTE_BUILD=1 TEST_COMPUTE_INIT=1 make all`).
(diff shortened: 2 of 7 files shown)


acfoltzer (changes requested):
First I just want to explicitly say that I don't think I've done a good job as Rust SME being clear about what we want the outcomes to be for this. This PR isn't too far off, but I think it'd be useful to write out what I have in mind here. @pchickey please pitch in as well as you have a closer connection to the issue of WASI snapshots.

1. We want our users to be able to use whatever version of `rustc` they'd like, as long as it is within the constraints set by the dynamic config. Currently, that means any version `>= 1.54.0`.
2. (Corollary of 1) If a user's `rustc` is within the constraints set by the dynamic config, but is not the latest `stable`, we don't want to bother them with a prompt to update.
3. In the `rust-toolchain` files bundled with starter kits, we want to be able to specify the `stable` toolchain so new users will get an up-to-date compiler without us having to do a PR every six weeks to bump the explicit version.

The main way this PR conflicts with those goals is the use of `+stable` in the `cargo build` invocation. Instead, we should not specify any toolchain when invoking `cargo`; we should only validate that the current default version is within the constraints.

The biggest problem I can see with this is it makes the remediation message trickier. `rustup update stable` is a perfectly good remediation if we are forcing `cargo +stable` and find it's outdated. However, point 2 means that if a user has 1.54.0 but the latest stable is 1.55.0, we don't want to prompt them with any remediation at all. 

But what if the user has 1.53.0, the latest stable is 1.55.0, and the constraint is `>= 1.54.0`? We could tell them to update their `rust-toolchain` file, but to what? What if they don't even have a `rust-toolchain` file and we're just getting whatever their `rustup default` is? What if they don't have `rustup` installed, and have installed Rust some other way?

Integralist:
Thanks for your feedback @acfoltzer 

So it seems although `rustup` is the preferred way to install rust (and as a side-effect of that, cargo) we should not presume it will be used by all users. 

Thinking about how to simplify the logic for validating the rust environment, the following seems to make sense to me...

- Lookup `rustc` in `$PATH`.
  - If there is an error: tell the user `rustc` is not found and they should ensure it's installed.
  - I'm presuming the remediation message should NOT suggest using `rustup` and leave the _how_ to the user.

- Execute `rustc --version`.
  - Compare the version output to the CLI config `toolchain_constraint` field.
    - Need to ensure our parsing logic accounts for `x.x.x-nightly` (e.g. don't just split on spaces).
  - If the constraint isn't met: tell the user their constraint isn't met and leave it to the user to choose the approach.

- Lookup `cargo` in `$PATH`.
  - If there is an error: tell the user it's not found and they should ensure it's installed.
  - I'm presuming the remediation message should NOT suggest _how_ to install `cargo`.

**What do you think of this?**. Any suggested changes to the flow?

I feel like this flow will work for when user's have `rustup` but are doing funky things like `rustup override set <…>`, as that will be reflected when running either `rustup show active-toolchain` and `rustc --version` (i.e. either of those latter commands will show the version set to be the override, and the `override` subcommand looks to also take priority over whatever is set in the `rust-toolchain` file).

Integralist:
@acfoltzer I've pushed changes that reflect the new flow I described earlier. Let me know your thoughts.

Integralist:
OK I ran into issues where the CLI was running `rustc --print sysroot` via a subprocess and, although the CI shell could run the same command and get the correct toolchain (i.e. 1.54.0), it seemed that the subprocess from the CLI would see 'stable' (i.e. the `rustup override set 1.54.0` appeared to not be set when running within the subprocess).

To try and work around that issue I changed the logic so that when checking the `wasm32-wasi` target I would use `rustup target add wasm32-wasi --toolchain 1.54.0` if `rustup` was available (which should be 99% of the time, and would be the case for the CI tests), otherwise I'd fallback to using the low-level check of looking inside `$(rustc --print sysroot)/lib/rustlib` for a `wasm32-wasi` directory in case `rustup` wasn't installed.

Unfortunately, although the tests pass fine on my machine, it seems the CI tests are still failing. It looks like they're getting _past_ the validation stage, and onto the 'build' stage, but are failing due to a new error which seems to hint towards being related to my earlier comments about the subprocess causing the toolchain override to be lost:

```
error[E0463]: can't find crate for `core`

= note: the `wasm32-wasi` target may not be installed
= help: consider downloading the target with `rustup target add wasm32-wasi`
```

This suggests that the `cargo build` is not using the correct toolchain.

Integralist:
So I'm now in a situation where using `rustup target list --installed --toolchain stable` is working on Ubuntu but appears to be failing for macOS and Windows. The latter two OS' suggest that the wasm32-wasi target isn't available to the stable toolchain (which again is confusing because we install the target to the stable toolchain -- hence the Ubuntu tests pass. If we weren't doing that, then the tests for Ubuntu should fail too).

acfoltzer:
D'oh, that CI problem sounds devilish. I'm afraid I don't have much experience with this sort of thing other than on Linux, but might be a bug to report to `action-rs`?

Regarding questions of which checks to perform, I think there are several simplifying assumptions we can make:

1. If a user wants to use anything other than a stable toolchain that is within our version constraints, they need to use `--force`. So, no worries about how to compare a nightly version string against a stable semver version.
2. Likewise if a user wants to use a non-`rustup` install, they should use `--force`. I know I raised the non-`rustup` question above but the more I think about it the more I don't want us to commit to being totally comprehensive about supporting all the various ways our ecosystem languages might be packaged.
3. We should assume `cargo` and `rustc` versions are kept in lockstep, as they are by `rustup`. If someone wants to use a stable cargo with a rustc they compiled themselves, I think `--force` remains okay.

I'm quick to fall back on `--force` for these more complex situations because I suspect anyone wanting to do these more advanced types of workflows will probably be more comfortable using `cargo` commands directly, and only using the CLI at the end to `fastly compute deploy`. That is just my guess, though.

So to go in more detail through the points in your earlier comment https://github.com/fastly/cli/pull/371#issuecomment-899460329:

> * Lookup `rustc` in `$PATH`.
>   
>   * If there is an error: tell the user `rustc` is not found and they should ensure it's installed.
>   * I'm presuming the remediation message should NOT suggest using `rustup` and leave the _how_ to the user.

I think we should point to [this page](https://www.rust-lang.org/learn/get-started) to remediate a missing `rustc`.

> * Execute `rustc --version`.
>   
>   * Compare the version output to the CLI config `toolchain_constraint` field.
>     
>     * Need to ensure our parsing logic accounts for `x.x.x-nightly` (e.g. don't just split on spaces).

Let's make sure we account for nightly so the parser doesn't crash, but we shouldn't try to do a comparison against the toolchain constraint. A nightly or beta or custom toolchain should always just fail and require `--force`.

>   * If the constraint isn't met: tell the user their constraint isn't met and leave it to the user to choose the approach.

Yeah, this is tricky. If our new standard practice is to use `stable` in the `rust-toolchain` file of starter kits, we should probably recommend `rustup update stable` as a remediation.

But that won't fix it if the issue is a `rust-toolchain` file that specifies a more specific version. Maybe something like "Run `rustup update stable`, or ensure your `rust-toolchain` file specifies a version matching $toolchain_constraint".

> * Validate `wasm32-wasi` target is installed.
>   
>   * Look inside `$(rustc --print sysroot)/lib/rustlib` to see if there is a `wasm32-wasi` directory.
>   * If the directory doesn't exist: tell the user they need to install the target.
> 
> > **NOTE**: In the `main` branch we use `rustup target list` to see if the target is installed but as per discussions we don't want to rely on user having `rustup` installed, it means we have to use the above approach instead.

I think `rustup target list --installed` is the right approach here.

> * Lookup `cargo` in `$PATH`.
>   
>   * If there is an error: tell the user it's not found and they should ensure it's installed.
>   * I'm presuming the remediation message should NOT suggest _how_ to install `cargo`.

I think we can skip this step due to simplifying assumption 3 above.

> * Lookup `Cargo.toml`.
>   
>   * If there is an error: tell the user they're missing this file.

I don't think we need this step. `cargo` already gives a suitable message in this case:

```
❯ cargo build
error: could not find `Cargo.toml` in `/tmp/foo` or any parent directory
```

> * Validate `fastly-sys` crate version.
>   
>   * If `fastly_sys_constraint` isn't met: tell user to update the `fastly` crate to the latest version via `Cargo.toml`.
> * Validate `fastly` crate version.
>   
>   * If not using latest fastly crate version: _suggest_ user update the `fastly` crate to the latest version via `Cargo.toml`.

These sound good.

Integralist:
👋🏻 @kailan the PR is ready for review now. 

👋🏻  @acfoltzer I've addressed your feedback, manually tested locally, and the updated code looks to work as intended.

The only thing I've _not_ done from your feedback is remove the check for the `cargo` binary. 

Although the majority of users will be using `rustup`, I would still prefer to check for `cargo` separately so we can handle the scenario where someone needs to avoid `rustup` (for whatever reason) and so they manually install a `rustc` binary and also manually compile `cargo`. We already have to check `rustc --version`, and so the one extra check for whether the `cargo` binary is available is not a big maintenance overhead (IMHO). 

Admittedly the messaging is more targeted towards `rustup` users, but I think that's OK considering that they're likely to be the majority of users, and if you're a user who needs to avoid `rustup`, then I think it's probably safe to say you're a more advanced user and more likely to just use the CLI's `--force` flag to side-step the validation altogether.

**So to summary**, the validation steps are:

1. Validate `rustc` installed.
2. Validate `rustc --version` meets the constraint.
3. Validate `wasm32-wasi` target is added to the relevant toolchain.
4. Validate `cargo` is installed.
5. Validate `fastly-sys` crate version.
6. Validate `fastly` crate version (optional upgrade suggestion).

acfoltzer (approved):
I'm not qualified to weigh in on the Go, but from a Rust perspective this looks like a good way to approach these checks. Thanks for your patience in iterating on this with me :rocket:

kailan (approved):

Issue #367, opened 2021-08-10
the Fastly API returned 404 Not Found: Not found when running compute deploy


makma:
**Version**
```
Fastly CLI version 0.29.0 (511ffdc)
Built with go version go1.16.4 linux/amd64
```

```
Fastly CLI version 0.36.0 (dd0b7d9)
Built with go version go1.16.6 linux/amd64
```

**What happened**
When I try to run `fastly compute deploy -s myId --token mytok --verbose` I've got the output:
```
Fastly API token provided via --token
Fastly API endpoint: https://api.fastly.com
Setting service ID in manifest to "<myId>"...
Uploading package...

ERROR: the Fastly API returned 404 Not Found: Not found.
```

The error message is quite useless. I tried various combinations of service Ids together with (probably) all combinations of tokens with various scopes with no luck. Any guidance is more than appreciated. Thanks.

fastly.toml
```
# This file describes a Fastly Compute@Edge package. To learn more visit:
# https://developer.fastly.com/reference/fastly-toml/

authors = ["Martin"]
description = "myintegration"
language = "rust"
manifest_version = 1
name = "example"
service_id = "myId"
```

makma:
Thank you @Integralist for your response. I'll try your suggestsion and let you know.

Integralist:
Thanks @makma 

re:

> ERROR: error downloading latest release: no asset found for your OS (windows) and architecture (amd64).

Looks like we just don't have that configured:

https://github.com/fastly/cli/blob/main/.goreleaser.yml#L28

So that should be a quick fix (I'll look to get a change merged shortly).

makma:
I found the root cause. I didn't have an enabled c@e feature flag.
Once it's enabled, it works. Nevertheless API could be more descriptive :)

Thank you @Integralist

PR #372, opened 2021-08-16
support Windows arm64

Integralist:
In issue https://github.com/fastly/cli/issues/367 customer also reported they tried to update the CLI on Windows from their old version but discovered there was no arm64 version available.


kailan (approved):
Looks good pending CI!

PR #374, opened 2021-08-18
pkg/commands/logs: respect --endpoint

peterbourgon:
`fastly logs tail` should respect the `--endpoint` flag.


Integralist (commented):
LGTM. Not sure why we have a `pkg/httpcli/client.go` package though? I presume this is an accidental commit.

Integralist (approved):

PR #378, opened 2021-08-25
Add region support for New Relic logging endpoint

kellymclaughlin:
Follows from https://github.com/fastly/go-fastly/pull/292. ~~Marking as Draft until the next release of `go-fastly` and will update the `go.mod` entry at that point and promote to regular PR.~~

### DIFF
--- a/pkg/commands/logging/newrelic/create.go
+++ b/pkg/commands/logging/newrelic/create.go
@@ -34,8 +34,8 @@ func NewCreateCommand(parent cmd.Registerer, globals *config.Data) *CreateComman
 	c.CmdClause.Flag("format", "A Fastly log format string. Must produce valid JSON that New Relic Logs can ingest").StringVar(&c.format)
 	c.CmdClause.Flag("format-version", "The version of the custom logging format used for the configured endpoint").UintVar(&c.formatVersion)
 	c.CmdClause.Flag("placement", "Where in the generated VCL the logging call should be placed").StringVar(&c.placement)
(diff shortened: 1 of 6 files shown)


Integralist:
Ping 🙂

pkg/commands/logging/newrelic/create.go
```diff
@@ -52,6 +53,7 @@ type CreateCommand struct {
 	name              string
 	placement         string
 	responseCondition cmd.OptionalString
+	region            string
```
Integralist, line 56: Apologies for the annoying nit-pick but I've been trying to order the flags alphabetically so could we move this flag and struct field up a line so the response condition flag/field comes after it.

pkg/commands/logging/newrelic/newrelic_test.go
```diff
@@ -64,7 +64,7 @@ func TestNewRelicCreate(t *testing.T) {
 					}, nil
 				},
 			},
-			Args:       args("logging newrelic create --key abc --name foo --service-id 123 --version 3"),
+			Args:       args("logging newrelic create --key abc --name foo --service-id 123 --version 3 --region eu"),
 			WantOutput: "Created New Relic logging endpoint 'foo' (service: 123, version: 3)",
```
Integralist, line 68: I would prefer to remove this flag from the test as we're not able to validate anything other than the flag has been defined (this is why I omitted the other flags and only included flags whose content we could see in the response object and would be displayed back to the user in the success message).

Integralist (changes requested):

Integralist (approved):

Issue #377, opened 2021-08-24
provide an update-config subcommand or cut a new release


softprops:
https://github.com/fastly/cli/pull/330 introduced a dynamic way to resolve a config file which I believe checks once every 5 minutes before polling and downloading the latest copy

At this time the current stable rust version is at 1.54.0. Also at this time the latest fastly cli vesrion (`v0.36.0`)  which [does not permit the current stable rust version](https://github.com/fastly/cli/blob/v0.36.0/cmd/fastly/static/config.toml#L13)

for a ci/cd pipeline its untenable to have to install the cli and have a pipeline wait 5 minutes for the cli to get the latest config before building and deploying.

one workaround to support current stable rust would be to provide a subcommand to explicitly update the config rather than 5 minutes.

another workaround to support current stable rust would simply to cut a new release of the cli from master which [allows for everything at or after thee current stable rust](https://github.com/fastly/cli/blob/3fbc5267d4c8ef49efec2c09654fb3506d441b53/cmd/fastly/static/config.toml#L13)

PR #382, opened 2021-09-02
update CLI config using flag on `update` command

Integralist:
Fixes https://github.com/fastly/cli/issues/377


kailan (approved):
A very helpful addition, thanks!

softprops:
when do we think we can see this in a release?

Integralist:
Hi @softprops I'm hoping to cut a new release in the next day or so.

PR #383, opened 2021-09-03
Add Service ID to `--verbose` output

Integralist:
The `--verbose` flag prints the Fastly API token source as well as the Fastly API endpoint, but it would be helpful for users to also see the Service ID as this can be sourced from multiple locations:

1. Flag
2. Environment
3. Manifest

The motivation behind this change is that we received a report from a user to say they ran into an issue where the Service ID within the manifest file was being overridden by the `FASTLY_SERVICE_ID` environment variable. The root cause was that they had reused the same shell from an earlier test experiment and hadn't notice the environment variable being set. 

Adding the Service ID to the `--verbose` output should help users to pinpoint a possible issue like this.

> **NOTE**: We've added the new display logic within the existing `cmd.ServiceDetails()` function, which is used over ~90% of the files, but we still need to add an explicit call to the new `cmd.DisplayServiceID()` function from within those functions that don't use `cmd.ServiceDetails()`. 
>
> For example, some commands don't require a service version (e.g. versionless resources like Edge Dictionaries and ACL entries) and so we don't call `cmd.ServiceDetails()` as that would also waste time on a bunch of unnecessary service API calls.

### DIFF
--- a/.tmpl/create.go
+++ b/.tmpl/create.go
@@ -47,10 +47,6 @@ type CreateCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *CreateCommand) Exec(in io.Reader, out io.Writer) error {
-	if c.Globals.Verbose() {
-		cmd.DisplayServiceID(c.manifest, out)
-	}
-
 	// Exit early if no token configured.
 	_, s := c.Globals.Token()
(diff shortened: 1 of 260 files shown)


doramatadora (approved):
LGTM 👍🏼

PR #387, opened 2021-09-03
defer container registry

Integralist:
There were issues cutting a new release as it seems we were missing some fundamental configuration/permissions/access to push container images to the GitHub Package Registry.

Due to time constraints, I'm commenting out the relevant code so we can revisit this and figure out the missing steps.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -94,23 +94,23 @@ snapshot:
   name_template: "{{ .Tag }}-next"
 changelog:
   skip: true
-dockers:
-- <<: &build_opts
-    use: buildx
-    goos: linux
-    goarch: amd64
-    image_templates:
-      - "ghcr.io/fastly/cli:{{ .Version }}"
-    build_flag_templates:
-      - "--platform=linux/amd64"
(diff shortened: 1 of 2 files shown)


Integralist:
Hi @andersjanmyr just wanted to make you aware that we had some issues cutting a new CLI release and so we need to do some further investigation. In the meantime I need to comment out the container registry configuration to enable us to cut a new CLI release.

andersjanmyr:
@Integralist Maybe this can help with this issue.
https://github.com/goreleaser/goreleaser/issues/1977

Integralist:
Hi @andersjanmyr thanks for the pointer but I don't believe that would have helped in this case (from what I can tell at least) because that fix you reference was published in goreleaser [`0.150.1`](https://github.com/goreleaser/goreleaser/releases/tag/v0.150.1) while we were already using goreleaser [`0.177.0`](https://github.com/fastly/cli/blob/74a03cca8348826480dab9e49a623bc91a3d6875/.github/workflows/tag_release.yml#L31) at the time of this PR.

PR #389, opened 2021-09-06
Fix download assets to a randomly generated dir

fgsch:
The previous attempt was very broken. Sorry!


Integralist (approved):
LGTM.

Screenshot 2021-09-06 at 16 20 29

PR #391, opened 2021-09-06
Remove Scoop `skip_upload` configuration

Integralist:
[Since moving away from automatically publishing releases](https://github.com/fastly/cli/pull/275) (e.g. since `v0.30.0` we manually curate the CHANGELOG) it turns out publishing an updated manifest file to https://github.com/fastly/scoop-cli has been broken. 

For example, when cutting a new CLI release you see the error `error=release is marked as draft`.

Looking at the https://goreleaser.com/customization/scoop/ documentation there is the setting `skip_upload` which we had set to `auto` (which would prevent the publishing from happening if the release looked like a pre-release). The default value is `false` and so by removing the setting I'm hoping that we'll see the published manifest data to succeed.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -85,7 +85,6 @@ scoop:
     owner: fastly
     name: scoop-cli
   homepage: https://github.com/fastly/cli
-  skip_upload: auto
   description: Fastly CLI
   license: Apache 2.0
 checksum:


Integralist:
OK so after reviewing the goreleaser code https://github.com/goreleaser/goreleaser/blob/b661a57105765b1b754b7fcc1780dcbfef65ea3d/internal/pipe/scoop/scoop.go#L121-L123 it seems that we can't side-step the fact that goreleaser will skip publishing to our scoop repo when our release is in draft. It's just what it does.

Looking at the Homebrew equivalent code https://github.com/goreleaser/goreleaser/blob/b661a57105765b1b754b7fcc1780dcbfef65ea3d/internal/pipe/brew/brew.go#L171-L179 it doesn't have the same logic and so this is why publishing to our Homebrew repo still works.

PR #392, opened 2021-09-07
Add `generated` vcl command

smaeda-ks:
usage:

```
$ fastly vcl generated describe --version=xx --service-id xx
```

### DIFF
--- a/pkg/api/interface.go
+++ b/pkg/api/interface.go
@@ -268,6 +268,8 @@ type Interface interface {
 	GetNewRelic(i *fastly.GetNewRelicInput) (*fastly.NewRelic, error)
 	ListNewRelic(i *fastly.ListNewRelicInput) ([]*fastly.NewRelic, error)
 	UpdateNewRelic(i *fastly.UpdateNewRelicInput) (*fastly.NewRelic, error)
+
+	GetGeneratedVCL(i *fastly.GetGeneratedVCLInput) (*fastly.VCL, error)
 }
 
 // RealtimeStatsInterface is the subset of go-fastly's realtime stats API used here.
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -57,6 +57,7 @@ import (
 	"github.com/fastly/cli/pkg/commands/update"
 	"github.com/fastly/cli/pkg/commands/vcl"
 	"github.com/fastly/cli/pkg/commands/vcl/custom"
+	"github.com/fastly/cli/pkg/commands/vcl/generated"
 	"github.com/fastly/cli/pkg/commands/vcl/snippet"
 	"github.com/fastly/cli/pkg/commands/version"
 	"github.com/fastly/cli/pkg/commands/whoami"
@@ -378,6 +379,8 @@ func Run(opts RunOpts) error {
 	statsRegions := stats.NewRegionsCommand(statsCmdRoot.CmdClause, &globals)
 	updateRoot := update.NewRootCommand(app, opts.ConfigPath, opts.Versioners.CLI, opts.HTTPClient, &globals)
 	vclCmdRoot := vcl.NewRootCommand(app, &globals)
+	vclGeneratedCmdRoot := generated.NewRootCommand(vclCmdRoot.CmdClause, &globals)
+	vclGeneratedDescribe := generated.NewDescribeCommand(vclGeneratedCmdRoot.CmdClause, &globals)
 	vclCustomCmdRoot := custom.NewRootCommand(vclCmdRoot.CmdClause, &globals)
 	vclCustomCreate := custom.NewCreateCommand(vclCustomCmdRoot.CmdClause, &globals)
 	vclCustomDelete := custom.NewDeleteCommand(vclCustomCmdRoot.CmdClause, &globals)
@@ -394,6 +397,9 @@ func Run(opts RunOpts) error {
 	whoamiCmdRoot := whoami.NewRootCommand(app, opts.HTTPClient, &globals)
 
 	commands := []cmd.Command{
+		vclCmdRoot,
+		vclGeneratedCmdRoot,
+		vclGeneratedDescribe,
 		aclCmdRoot,
 		aclCreate,
 		aclDelete,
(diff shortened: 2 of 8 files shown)


Integralist:
ping @triblondon

Integralist:
@smaeda-ks thanks for opening this PR. 

I _think_ (and @triblondon let me know your thoughts) we should change the interface:

```diff
- fastly vcl generated describe
+ fastly vcl generated
```

...because we won't have a need for any of the other 'verbs'.

smaeda-ks:
Interesting. I thought having some sort of subcommand such as "describe" (or, perhaps "show") would be better for the sake of interface consistency.

Integralist:
>  ...would be better for the sake of interface consistency.

That's absolutely understandable. I guess I would still like to reduce unnecessary typing 🙂

**Side note**: I never was a fan of the use of 'describe' in the CLI, and it feels especially awkward (to me at least) when seen in this command's context.

ejthurgo:
I really dislike 'describe' and would have always preferred to go with something like 'show', but consistency should win here.
Bit of a stretch for `vcl generated` but I have a general issue with making assumptions about whether the other verbs are required - they might not be now, but could be later, and the ensuing changes would break backward compatibility for users scripts.

Integralist:
> Bit of a stretch for vcl generated but I have a general issue with making assumptions about whether the other verbs are required - they might not be now, but could be later, and the ensuing changes would break backward compatibility for users scripts.

That makes sense, but in the case of `vcl generated` there are no logical verbs to be used as it's a single API endpoint and so there's nothing like list or delete etc that would make sense in this specific example.

ejthurgo:
Sure yeah, `vcl generated` is a poor example, but we're going to make an exception for this one thing? slippery slope.

Integralist:
> we're going to make an exception for this one thing? slippery slope

I don't feel like this counts as a 'slippery slope' because, in this particular instance, at what point in the future do we think [this API](https://developer.fastly.com/reference/api/vcl-services/vcl/#get-custom-vcl-generated) is going to need a nested hierarchy? 

Unless I'm mistaken (which is very possible), you can only do **_one_** thing with the generated VCL ...display it. You can't edit, list or delete it etc. The user already has things like `fastly vcl custom ...` for that. 

Yes you can GET (i.e. "describe" it, which again is horrible wording in this context), but it's an extra level of hierarchy (and code to maintain) for no real benefit.

If y'all feel having this as an extra subcommand is better, then I would be more onboard with it if it wasn't "describe" but "show" (i.e. `fastly vcl generated show`) because at least it would read nicer IMHO.

triblondon:
This is a great discussion to have because it's a point of principle we need to establish in relation to a bunch of possible commands that don't map to API endpoints, like `diff`.

Currently, we use the command `describe` to mean, "display all the properties of the object".  I don't think we have any cases where describe chooses a single property of the object to display it, therefore using `describe` would be inconsistent with other usage.

However, dropping the verb means ending the command on a past participle, not even a noun, which is not aligned with any existing pattern.

I don't know if we already use `show` elsewhere but that seems appropriate here - and since we are showing the VCL, and the fact that it is generated is a modification to the format, but the entity being displayed is the same, I would consider this:

```
fastly vcl show --generated
```

Integralist:
The use of...

```
fastly vcl show --generated
```

...would also help with the related API https://developer.fastly.com/reference/api/vcl-services/vcl/#get-custom-vcl-generated-highlighted if we ever decided to utilise it because we could add it as an extra flag...

```
fastly vcl show --generated --syntax-highlighted
```

Integralist:
So I'm in favour of:

```
fastly vcl show --generated
```

👍🏻

smaeda-ks:
Thanks. I wasn't expecting this would be so controversial though...

Given we already have other subcommands for `fastly vcl` such as `snippet` and `custom` it doesn't feel adding `fastly vcl show` is the right thing here.

Integralist:
OK I can see both sides of the argument.

Ultimately I want to move the needle on this PR so happy to concede.

smaeda-ks:
Sure

triblondon:
Sorry it's taken a while to circle back to this.  So, let's look at all the child commands of `fastly vcl`:

* `fastly vcl custom <create|delete|describe|list|update>`
* `fastly vcl snippet <create|delete|describe|list|update>`

We did plan to add a bunch more here, to include all the objects that generate VCL, such as:

* `fastly vcl director`
* `fastly vcl pool`
* `fastly vcl cache-settings`
* `fastly vcl condition`
* `fastly vcl header-object`
* ...

But we subsequently reasoned that there is barely any use case for these high-level edge logic concepts being provisioned via the CLI.  You could argue that there's really very little case for them even having API support, since they primarily exist for customers who prefer to control their services via the UI.

Nevertheless, what all these things (`vcl custom` and `vcl snippet` included) have in common is that they are things that contribute VCL to the service, but _do not constitute the entirety of the service's VCL configuration_.  Conversely, *generated VCL* is a derivative of all of these child concepts.  Think of it as custom VCL + snippets + some conditions + header objects + a director or two == the 'complete' VCL.

For that reason, `fastly vcl show` makes sense here.  In fact, since there is no concept of showing the service's entire VCL without "generating" it, there's actually no need for the additional `--generated` flag.  I'm happy to hear other views on that, if there's something else that makes sense to display for `fastly vcl show`, in which case we keep the flag but the command remains the same.

This also helps to be consistent with my preferred distinction between `describe` and `show`:

* `describe` enumerates and displays all metadata of the object.  Most objects are composed entirely of metadata, and so this shows all properties.  Some, like `package`, have a content property that is impractical to display as part of a `describe` command
* `show` outputs a single property value from the object, or a single derived value.

There's an argument that `fastly vcl custom` and `fastly vcl snippet` should exclude the `content` property when doing a `describe`, and we should add a `show` command to these objects to display the content.  But that feels like a discussion for another day :-).  If there are no objections, I'd like to add this command and call it `fastly vcl show`.

smaeda-ks:
Having `show` and other sub-commands such as `snippet` and `custom` on the same level of the hierarchy is also the weird CLI design, IMHO. I still prefer `fastly vcl generated show`, and we do have an endpoint called `/generated_vcl`, so I don't think it's bringing any downsides or inconsistency against the overall principle you're describing.

PR #395, opened 2021-09-07
Teach the manifest about dictionaries

### LINKED ISSUE #393: Rewriting fastly.toml strips out some properties
**Version**

Fastly CLI version 0.37.1 (e100937)

**What happened**

When running `fastly compute serve` I noticed dictionary lookups were causing panics. 
Upon further inspection, I found that the CLI was stripping out some properties (e.g. local_server.dictionaries) when rewriting the fastly.toml and the original input did not have `manifest_version`.

fwiw, the same file works without changes in Viceroy.

fgsch:
Fixes #393.

### DIFF
--- a/pkg/commands/compute/manifest/manifest.go
+++ b/pkg/commands/compute/manifest/manifest.go
@@ -216,13 +216,20 @@ type Mapper map[string]interface{}
 // LocalServer represents a list of backends that should be mocked as per the
 // configuration values.
 type LocalServer struct {
-	Backends map[string]Backend `toml:"backends"`
+	Backends     map[string]Backend    `toml:"backends"`
+	Dictionaries map[string]Dictionary `toml:"dictionaries,omitempty"`
 }
 
 // Backend represents a backend to be mocked by the local testing server.
 type Backend struct {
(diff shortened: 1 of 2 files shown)


pkg/commands/compute/manifest/manifest.go
```diff
@@ -225,6 +226,12 @@ type Backend struct {
 	OverrideHost string `toml:"override_host,omitempty"`
 }
 
+// Dictionary represents a dictionary to be mocked by the local testing server.
+type Dictionary struct {
+	File   string `toml:"file"`
+	Format string `toml:"format"`
```
Integralist, line 232: Does `format` need `omitempty`? 

By this I mean: is it a required property, or will the file format be inferred (or a default used) by Viceroy?

Integralist (approved):
LGTM. One small comment.

pkg/commands/compute/manifest/manifest.go
fgsch: It is required.

Integralist: OK cool 👍🏻

PR #397, opened 2021-09-07
Refactor and improvements

fgsch:
Remove unused code and simplify Download().

### DIFF
--- a/pkg/commands/update/versioner.go
+++ b/pkg/commands/update/versioner.go
@@ -23,8 +23,6 @@ type Versioner interface {
 	Binary() string
 	Download(context.Context, semver.Version) (filename string, err error)
 	LatestVersion(context.Context) (semver.Version, error)
-	Name() string
-	RenameLocalBinary(binName string) error
 	SetAsset(name string)
 }
 
@@ -34,7 +32,6 @@ type GitHub struct {
 	org          string
 	repo         string
 	binary       string // name of compiled binary
-	local        string // name to use for binary once extracted
 	releaseAsset string // name of the release asset file to download
 }
 
@@ -69,23 +66,6 @@ func (g *GitHub) Binary() string {
 // external binaries from within the CLI.
 func (g *GitHub) SetAsset(name string) {
 	g.releaseAsset = name
-}
-
-// RenameLocalBinary will rename the downloaded binary.
-//
-// NOTE: This exists so that we can, for example, rename a binary such as
-// 'viceroy' to something less ambiguous like 'fastly-localtesting'.
-func (g *GitHub) RenameLocalBinary(binName string) error {
-	g.local = binName
-	return nil
-}
-
-// Name will return the name of the binary.
-func (g GitHub) Name() string {
-	if g.local != "" {
-		return g.local
-	}
-	return g.binary
 }
 
 // LatestVersion implements the Versioner interface.
@@ -166,16 +146,6 @@ func (g GitHub) Download(ctx context.Context, version semver.Version) (filename 
 		if err := os.Rename(filepath.Join(dir, g.binary), assetFile); err != nil {
 			return filename, fmt.Errorf("error renaming binary: %w", err)
 		}
-	}
-
-	// TODO: This is racy without the os.CreateTemp() call but does not look like it's used.
-	// Can this be removed and let the Download's caller rename the binary?
-	if g.local != "" {
-		newName := filepath.Join(tmp, g.local)
-		if err := os.Rename(assetFile, newName); err != nil {
-			return filename, fmt.Errorf("error renaming binary: %w", err)
-		}
-		assetFile = newName
 	}
 
 	// G302 (CWE-276): Expect file permissions to be 0600 or less
--- a/pkg/commands/update/versioner.go
+++ b/pkg/commands/update/versioner.go
@@ -81,26 +81,46 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
 func (g GitHub) Download(ctx context.Context, version semver.Version) (filename string, err error) {
 	releaseID, err := g.getReleaseID(ctx, version)
 	if err != nil {
-		return filename, err
+		return "", err
 	}
 
 	release, _, err := g.client.Repositories.GetRelease(ctx, g.org, g.repo, releaseID)
 	if err != nil {
-		return filename, fmt.Errorf("error fetching release: %w", err)
+		return "", fmt.Errorf("error fetching release: %w", err)
 	}
 
 	assetID, err := g.getAssetID(release.Assets)
 	if err != nil {
-		return filename, err
+		return "", err
 	}
 
 	rc, _, err := g.client.Repositories.DownloadReleaseAsset(ctx, g.org, g.repo, assetID, http.DefaultClient)
 	if err != nil {
-		return filename, err
+		return "", err
 	}
 	defer rc.Close()
 
-	var extension string
+	dir, err := os.MkdirTemp("", "fastly-download")
+	if err != nil {
+		return "", fmt.Errorf("error creating release download directory: %w", err)
+	}
+	defer os.RemoveAll(dir)
+
+	dst, err := os.Create(filepath.Join(dir, g.releaseAsset))
+	if err != nil {
+		return "", fmt.Errorf("error creating release asset file: %w", err)
+	}
+
+	_, err = io.Copy(dst, rc)
+	if err != nil {
+		return "", fmt.Errorf("error downloading release asset: %w", err)
+	}
+
+	if err := dst.Close(); err != nil {
+		return "", fmt.Errorf("error closing release asset file: %w", err)
+	}
+
+	tempFile := dst.Name()
 
 	// TODO: We might need to also account for Window users by also checking for
 	// the .zip extension that goreleaser generates:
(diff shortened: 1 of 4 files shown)


pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
 func (g GitHub) Download(ctx context.Context, version semver.Version) (filename string, err error) {
```
peterbourgon, line 81: Yeah, the named return params are just creating ambiguity here, as evidenced by all the 

```go
return filename, <non-nil error>
```

blocks in the method.  Should just remove them e.g. `(string, error)`.

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
-		extension = ".tar.gz"
+	dir, err := os.MkdirTemp("", "fastly-download")
+	if err != nil {
+		return "", fmt.Errorf("error creating release download directory: %w", err)
 	}
+	defer os.RemoveAll(dir)
 
-	tmp := os.TempDir()
-	dst, err := os.CreateTemp(tmp, fmt.Sprintf("%s_%s_*%s", g.binary, version, extension))
+	dst, err := os.Create(filepath.Join(dir, g.releaseAsset))
```
peterbourgon, line 109: Re-using the `dst` var name throughout the function is also creating unnecessary ambiguity, I think...

```suggestion
	assetFile, err := os.Create(filepath.Join(dir, g.releaseAsset))
```

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
+		return "", fmt.Errorf("error downloading release asset: %w", err)
 	}
 
 	if err := dst.Close(); err != nil {
-		return filename, fmt.Errorf("error closing release asset file: %w", err)
+		return "", fmt.Errorf("error closing release asset file: %w", err)
 	}
 
-	assetFile := dst.Name()
+	tempFile := dst.Name()
```
peterbourgon, line 123: Capturing the name of the file in a separate variable creates similar ambiguity, and I don't think represents a benefit over using `assetFile.Name()` directly.

```suggestion
```

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
-		if err != nil {
-			return filename, fmt.Errorf("error creating temp extraction directory: %w", err)
-		}
-		defer os.RemoveAll(dir)
-		if err := archiver.NewTarGz().Extract(assetFile, g.binary, dir); err != nil {
-			return filename, fmt.Errorf("error extracting binary: %w", err)
-		}
-		if err := os.Rename(filepath.Join(dir, g.binary), assetFile); err != nil {
-			return filename, fmt.Errorf("error renaming binary: %w", err)
+		if err := archiver.NewTarGz().Extract(tempFile, g.binary, dir); err != nil {
```
peterbourgon, line 135: So then this would become

```suggestion
		if err := archiver.NewTarGz().Extract(assetFile.Name(), g.binary, dir); err != nil {
```

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
-
-	// TODO: This is racy without the os.CreateTemp() call but does not look like it's used.
-	// Can this be removed and let the Download's caller rename the binary?
-	if g.local != "" {
-		newName := filepath.Join(tmp, g.local)
-		if err := os.Rename(assetFile, newName); err != nil {
-			return filename, fmt.Errorf("error renaming binary: %w", err)
-		}
-		assetFile = newName
+		tempFile = filepath.Join(dir, g.binary)
```
peterbourgon, line 138: And you can skip this

```suggestion
```

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
 	// G302 (CWE-276): Expect file permissions to be 0600 or less
 	// gosec flagged this:
 	// Disabling as the file was not executable without it and we need all users
 	// to be able to execute the binary.
 	/* #nosec */
-	err = os.Chmod(assetFile, 0777)
+	err = os.Chmod(tempFile, 0777)
+	if err != nil {
+		return "", err
+	}
```
peterbourgon, line 149: Which makes this

```suggestion
	if err := os.Chmod(filepath.Join(dir, g.binary), 0777); err != nil {
		return "", fmt.Errorf("changing asset permissions: %w", err)
	}
```

pkg/commands/update/versioner.go
```diff
@@ -101,91 +81,83 @@ func (g GitHub) LatestVersion(ctx context.Context) (semver.Version, error) {
...
-		return filename, err
+		return "", fmt.Errorf("error creating temp file: %w", err)
+	}
+
+	assetFile := dst.Name()
+
+	if err := os.Rename(tempFile, assetFile); err != nil {
+		os.Remove(assetFile)
+		return "", fmt.Errorf("error renaming temp file: %w", err)
 	}
```
peterbourgon, line 161: Why do the extra work of moving the file to PWD vs. just returning the asset file directly from the temp dir?

peterbourgon (commented):
This PR makes things way less ambiguous! 👍 I think there are a few more things that would help, too...

pkg/commands/update/versioner.go
fgsch: No, I can't since this depends on the asset having a .tar.gz extension.

pkg/commands/update/versioner.go
fgsch: See above why I cannot simply use filepath.Join(dir, g.binary) here.

pkg/commands/update/versioner.go
fgsch: Thanks, will do.

pkg/commands/update/versioner.go
peterbourgon: Ah! So `g.binary` will be e.g. `viceroy` and the file returned by this function need to be `something-with.tar.gz`? What has that requirement? If the filename returned by Download has any requirement, it would be good to document that somewhere. (I know this isn't your code! 😅 But if we're tending the garden, and all that...)

fgsch: Not quite. `Download()` should return the asset in its uncompressed form. Since it can be called by multiple instances of the cli, the filename returned should be temporary (and unique); the callee will move the file to its final path and name.

I agree, it'd be a good idea to document this. I will take a look once I finish splitting this PR.

pkg/commands/update/versioner.go
fgsch: Because it makes cleaning up on error easier without shifting the responsibility to the callee.

peterbourgon: Sorry, I'm probably dense, but I don't get it ☹️ The initial file created by `os.Create(filepath.Join(dir, g.releaseAsset))` will already be unique per invocation, right? How does moving it to PWD make it easier to clean up or move more responsibility to the caller? 

If you want to remove it on failure, you could do something like

```go
func Download(...) (_ string, err error) {
    f, err := os.Create(filepath.Join(dir, g.releaseAsset))
    if err != nil { ... }

    defer func(tmpfile string) { 
        if err != nil { os.Remove(tmpfile) }
    }(f.Name())

    ...
```

fgsch: It would be unique per invocation but multiple instances can run in parallel.

fgsch: Just to make sure we're on the same page, `Download()` creates a temporary directory (`dir`) inside $TMPDIR to download and uncompress the asset if needed. It then change the permissions and moves the file to a random file in $TMPDIR, returning it and removing `dir`.

Was your original question why the last part (moving the file to a random file in $TMPDIR) takes place or I misunderstood?

peterbourgon: > Was your original question why the last part (moving the file to a random file in $TMPDIR) takes place or I misunderstood?

Yeah, basically:

> Download() creates a temporary directory (dir) inside $TMPDIR to download and uncompress the asset if needed.

Because `dir` is created by os.MkdirTemp, it will be unique every time, right?

> It then change the permissions and moves the file to a random file in $TMPDIR, returning it and removing dir.

Given the above, isn't this unnecessary?

fgsch: I might be missing something but I don't see how I return the file within `dir` and remove the directory and its content at the same time when returning from `Download()`.

fgsch: Also, for the record, I will be closing this PR in favour of #399 and another PR yet to be opened.

peterbourgon: > I don't see how I can return the file within dir and remove the directory and its content at the same time when returning from Download().

The file itself is handed off to the caller and becomes their responsibility, and unless there is another requirement lurking somewhere, it doesn't matter what directory it's in. The temporary directory itself doesn't represent a resource that necessarily needs to be cleaned up by you, the OS is ultimately able to do that.

fgsch: I think it's pretty bad etiquette to create files in /tmp (or it's equivalent) and don't remove them. YMMV.
Either way, that's a behaviour I'd like to have and/or keep.

peterbourgon: It's your call, of course, but I'll observe that (a) this use case is exactly what OS-level temp dirs are for, and (b) you can reasonably assume the tmpdir is writable, but it's possible the current dir is not.

fgsch:
This was OBE. The remaining changes will be moved to a new PR.

Issue #175, opened 2020-12-03
Log output should be simplified when running in CI


kailan:
Example from GitHub Actions: 

<img width="1884" alt="Screenshot 2020-12-03 at 10 47 53" src="https://user-images.githubusercontent.com/3542064/101001079-9351c000-3556-11eb-902f-ebba851ecfe4.png">

drwilco:
I would say don't use spinners/colors when not a terminal: https://rosettacode.org/wiki/Check_output_device_is_a_terminal#Go

kailan:
This issue can be closed. For the sake of CI, the `--verbose` flag provides the right output.

PR #405, opened 2021-09-15
Use a non-interactive progress writer in non-TTY environments

phamann:
Closes: #175 

### TL;DR
The current default progress writer uses interactive spinners on an event loop; while this style of output is useful in user terminal environments to give constant feedback of progression, it is not ideal in non-TTY environments, such as in continuous integration builds. 

### What?
This PR solves the CI issue by renaming the existing `text.QuietProgress` to `text.InteractiveProgress`, as it's not really "quiet", and then creates a new implementation of `text.QuietProgress` which only writes the step functions and not the intermediary writes to stdout. Lastly, we have a new constructor `text.NewProgress()` which is used throughout the program as a helper to abstract the logic which determines which progress should be used.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,7 +41,7 @@ require (
 	github.com/google/jsonapi v0.0.0-20201022225600-f822737867f6 // indirect
 	github.com/klauspost/compress v1.13.5 // indirect
 	github.com/klauspost/pgzip v1.2.5 // indirect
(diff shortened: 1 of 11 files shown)


phamann:
I've now fixed the failing tests 😄

Integralist (approved):
LGTM.

PR #409, opened 2021-09-17
Implement `auth-token` commands

Integralist:
**NOTE**: This PR also bumps to a new major go-fastly version which fixes some important bugs in the API client. This means the change set has become a lot larger as we have to update all `/v3` references to `/v4`.


triblondon (approved):

Integralist:
This PR is blocked by https://github.com/fastly/go-fastly/pull/304

PR #416, opened 2021-09-22
Remove Rust debug flags

phamann:
### TL;DR
Removes the `debuginfo=2` flag that we pass to `rustc` via a `RUSTFLAGS` env var, as its been found to slow down builds considerably and isn't really used for debugging.


Integralist (approved):

phamann:
Having now discussed this at length internally, this is now ready to merge.

PR #418, opened 2021-09-24
Clarify Starter Kit options in `compute init` flow

Integralist:
Screenshot 2021-09-24 at 15 14 35


Integralist:
@triblondon @kailan this PR can be reviewed (and approved if you're happy with it) but this code is blocked from being published into a new CLI release until https://developer.fastly.com/api/internal/cli-config is serving config with a `description` field.

triblondon (approved):

kailan (approved):

PR #421, opened 2021-09-28
Abstraction layer for `[setup.backends]`

Integralist:
**NOTE**: The below screenshots aren't the final product. For example they will include language in the displayed user messages that will likely have been updated as part of a code review. 

## Scenario: existing service

If you set a Service ID into the `fastly.toml` file, we will not validate the service for missing backends, nor will we prompt the user to create any backends...

Screenshot 2021-09-29 at 16 15 49

## Scenario: new service

If no Service ID, then we'll prompt the user to create a service, including any backend resources required. 

In the case of `[setup]` the user will be prompted for backends to be created that align with that configuration...

Screenshot 2021-09-29 at 16 40 59

If there is no `[setup]` configuration, then the user will be prompted for at least one backend (more can be added)...

Screenshot 2021-09-29 at 16 42 56

Lastly, if the user doesn't enter any information, then an originless backend will be created...

Screenshot 2021-09-29 at 16 44 38


triblondon:
Where you have [setup] information, could you say:

```
Creating a backend called 'backend_name' (backend able to serve /articles path):
Hostname or IP address: [reqbin.com]
Port: [443]

Creating a backend called 'other_backend_name' (backend able to serve /anything path):
Hostname or IP address: [httpbin.org]
Port: [443]
```

Because the `prompt` property for the backend in the [setup] data doesn't relate specifically to the hostname.

Integralist:
@triblondon updated output.

A couple of things to note:

1. The 'prompt' value is on a separate line from the 'Configure a backend called...' line because it gets wrapped and looked a bit weird in some cases. I can put it back onto one line if you prefer.
2. The 'prompt' field I'm going to update in the fastly.toml to be 'description' because it's no longer used as part of the _prompt_ to the user.

Screenshot 2021-10-01 at 12 20 55

triblondon:
👍

Integralist:
@triblondon this is ready for review now.

pkg/commands/compute/init_test.go
```diff
@@ -21,6 +22,10 @@ func TestInit(t *testing.T) {
 		t.Skip("Set TEST_COMPUTE_INIT to run this test")
 	}
 
+	// TODO: Change back to "main" before merging PR.
```
Integralist, line 25: @triblondon in order for the tests to show as passing I needed to use a PR branch for each of the starter kits we pull in. 

So basically once this PR is approved, and all the starter kit PRs are approved (see: links below), then I'll need to coordinate merging both this PR and cut a new CLI release as quickly as possible after the fact.

- https://github.com/fastly/compute-starter-kit-rust-default/pull/49
- https://github.com/fastly/compute-starter-kit-assemblyscript-default/pull/21
- https://github.com/fastly/compute-starter-kit-javascript-default/pull/7
- https://github.com/fastly/compute-starter-kit-rust-beacon-termination/pull/16
- https://github.com/fastly/compute-starter-kit-rust-static-content/pull/13

> NOTE: with the exception of the rust starter kit they're all one line changes, so if you could also take a moment to approve those (when you have a chance).

Integralist:
@kailan adding you to the reviews in case Andrew is busy.

triblondon (approved):
I'm happy with the UX here.

pkg/errors/errors.go
```diff
@@ -41,8 +41,21 @@ var ErrInvalidManifestVersion = RemediationError{
 	Remediation: "Delete `[manifest_version]` from the fastly.toml if present",
 }
 
+// ErrIncompatibleManifestVersion means the manifest_version defined is no
+// longer compatible with the current CLI version.
+var ErrIncompatibleManifestVersion = RemediationError{
+	Inner:       fmt.Errorf("the fastly.toml contains an incompatible manifest_version number"),
+	Remediation: "Update the `manifest_version` in the fastly.toml and refer to https://developer.fastly.com/reference/fastly-toml/#setup-information",
```
Integralist, line 48: @integralist I need to update this message so it points to the next CLI changelog that this code will be released under.

PR #422, opened 2021-09-29
Provide better remediation for unrecognised `manifest_version`

Integralist:
**NOTE**: This PR also fixes two other issues... 

1. An external library function (`toml.Unmarshal()`) manipulates the returned error, resulting in the intended remediation error message being hidden.
2. The 'default' starter kit is no longer showing as the first option.


kailan (approved):
Looks good to me 👍

PR #426, opened 2021-10-01
[WIP] Adding 'Usage' examples to help output

Integralist:


pkg/commands/backend/create.go
```diff
@@ -1,6 +1,7 @@
 package backend
 
 import (
+	_ "embed"
```
Integralist, line 4: This is an example of adding a NOTE to the CLI help output for this particular command. Three things have to happen...

1. Import the embed package
2. Create a variable to hold the embedded file as a string.
3. Define a `Notes()` method that will override the default no-op version in the base object.

If no `Notes()` method defined, then nothing is added to the CLI help output.

pkg/commands/backend/notes/create.txt
```diff
@@ -0,0 +1,2 @@
+API endpoint information:
```
Integralist, line 1: In future we would flesh this out to have more specific information relevant to the flags implemented in the CLI command (e.g. we would say something along the lines of "read the following API documentation to understand the model for the --example flag"), but for now this will suffice as an example of injecting custom information.

triblondon (commented):
I put some comments on this ticket with some higher level thoughts:

https://jira.corp.fastly.com/browse/DEVEX-10?focusedCommentId=716673&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-716673

PR #428, opened 2021-10-06
The `compute init --from` flag should support fiddles

Integralist:
**TODO**: Write tests to validate new behaviour.


pkg/commands/compute/init.go
```diff
@@ -52,7 +55,7 @@ func NewInitCommand(parent cmd.Registerer, client api.HTTPClient, globals *confi
 	c.CmdClause.Flag("description", "Description of the package").Short('d').StringVar(&c.manifest.File.Description)
 	c.CmdClause.Flag("author", "Author(s) of the package").Short('a').StringsVar(&c.manifest.File.Authors)
 	c.CmdClause.Flag("language", "Language of the package").Short('l').StringVar(&c.language)
-	c.CmdClause.Flag("from", "Git repository containing package template").Short('f').StringVar(&c.from)
+	c.CmdClause.Flag("from", "Git repository URL containing package template or Fastly Fiddle zip URL").Short('f').StringVar(&c.from)
```
Integralist, line 58: Tweak messaging so any valid zip is allowed (don't restrict just to fiddle).

pkg/commands/compute/init.go
```diff
@@ -1,9 +1,12 @@
 package compute
```
Integralist, line 1: Append to...

> Creating a new Compute@Edge project.

...information about the use of `--from` so it's clear that using the from flag has done something (as the 'progress' spinner messaging is maybe too subtle otherwise).

PR #429, opened 2021-10-06
Refactor `app.Run()`

Integralist:
Refactor of `app.Run()` function...

- The long list of command instantiations moved to a separate file.
- This refactor reduces the lines of code in `run.go` by ~77%. 
  - This is important as it's the main application entry point (making it much easier to focus on the logic flow).
- I've added a bunch of comments to the code to help clarify some Kingpin specific behaviours.
- We remove the use of `golint` as it has been [marked as deprecated/unmaintained](https://github.com/golang/go/issues/38968) since May 2020.
- The help output won't show error message when using --help flag (see explanation below).

One part of the refactor of `app.Run()` has resulted in the user no longer seeing an error displayed when requesting help output. For example, executing `fastly backend create --help` would display the help for `backend create` but (as far as the current CLI release is concerned) would also have been followed by a bright bold/red error message saying:

```
"ERROR: error parsing arguments: required flag --name not provided." 
```

This error message is unnecessary (and not to mention confusing) because the user has set the `--help` flag so there is no expectation from the user for the command, and all its associated side effects, to be processed (i.e. a user calling `--help` doesn't expect the command to go ahead and successfully create a backend), so there is no need to display flag parsing errors to the user in this situation.

As of this PR branch, commands using `--help` will work without giving these types of error.


.github/workflows/pr_test.yml
```diff
@@ -44,9 +44,6 @@ jobs:
     - name: "Static analysis check"
       run: make staticcheck
       shell: bash
-    - name: "Linting"
-      run: make lint
```
kailan, line 48: I'm guessing this was broken?

pkg/cmd/cmd.go
```diff
@@ -173,3 +173,22 @@ func DisplayServiceID(sid string, s manifest.Source, out io.Writer) {
 	text.Output(out, "Service ID%s: %s", via, sid)
 	text.Break(out)
 }
+
+// ArgsIsHelpJSON determines whether the supplied command arguments are exactly
+// `help --format json`.
+func ArgsIsHelpJSON(args []string) bool {
+	return (len(args) == 3 &&
+		args[0] == "help" &&
+		args[1] == "--format" &&
```
kailan, line 182: Would this work for `fastly help --format=json`?

kailan (approved):
Looks good to me, a couple of comments but I won't block on them as they are just moving existing code around.

PR #439, opened 2021-10-14
Fix shell completion

Integralist:
An earlier refactor of `app.Run()` caused the shell completion flags to return a runtime panic.

This subsequently broke updating the app via Homebrew as it [calls the fastly binary with these flags automatically](https://github.com/fastly/homebrew-tap/blob/master/Formula/fastly.rb#L41-L42).

## TESTS

I've added tests to validate the shell completion output. This means in future if we were to accidentally break the behaviour again we should see it as a CI error and it will prevent us from merging the breaking code.

## NOTES

While implementing a fix for this feature I discovered `app.Parse()` (a Kingpin method) was resetting the values we set for where Kingpin should write its output (we use `app.Writers()` to set `os.Stdout` and `os.Stderr` temporarily to use `io.Discard` for reasons already documented in the CLI code). 

This was a problem because once I started working on a fix I noticed I was not only seeing the shell completion output but it was being followed by our verbose help output. This was happening because when we execute the CLI binary but only pass a shell completion flag, then we're not actually passing a command (e.g. like `fastly acl`), so the Kingpin parser thinks this is an error and provides the help output and tells us to provide a command.

Because Kingpin is internally overriding where it writes output to (it sets it to `os.Stdout` rather than what we had set, which was `io.Discard`) it means I can't stop Kingpin from displaying the help output. So to work around this I ensure the arguments provided to the Kingpin parser include a valid command (along with the shell completion flag). 

An example might look something like:

```bash
fastly --completion-script-bash acl
```

But rather than rely on an actual user feature command like `acl`, I define a 'hidden' command (i.e. it doesn't show up in the help output) called `shellcomplete` that we can safely append to the arguments list. We don't have to worry about this hidden command being accidentally removed in the future as we now have a test to validate the shell autocomplete behaviours.

For posterity I've added these notes as comments in the CLI code.

### DIFF
--- a/pkg/app/commands.go
+++ b/pkg/app/commands.go
@@ -46,6 +46,7 @@ import (
 	"github.com/fastly/cli/pkg/commands/purge"
 	"github.com/fastly/cli/pkg/commands/service"
 	"github.com/fastly/cli/pkg/commands/serviceversion"
+	"github.com/fastly/cli/pkg/commands/shellcomplete"
 	"github.com/fastly/cli/pkg/commands/stats"
 	"github.com/fastly/cli/pkg/commands/update"
 	"github.com/fastly/cli/pkg/commands/user"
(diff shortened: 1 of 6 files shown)


Integralist:
I don't believe so. But we can see that running the relevant commands (e.g. `fastly --completion-script-bash`) no longer cause a runtime panic so we've fixed the issue from that side of things and so there should be no reason for Homebrew to error now.

PR #445, opened 2021-10-19
Fallback to existing viceroy binary in case of network error

Integralist:
This PR ensures that if a user has no internet connection and they want to develop locally (i.e. `compute serve`) then they should be able to without the Viceroy binary check causing an error to be returned.

**NOTE**: This only applies if the user has a pre-existing Viceroy binary installed for us to fallback to. If not, then the network error is still returned.

### DIFF
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -128,7 +128,7 @@ func getViceroy(progress text.Progress, out io.Writer, versioner update.Versione
 
 	bin = filepath.Join(InstallDir, versioner.Binary())
 
-	// NOTE: When checking if viceroy is installed we don't use
+	// NOTE: When checking if Viceroy is installed we don't use
 	// exec.LookPath("viceroy") because PATH is unreliable across OS platforms,
 	// but also we actually install viceroy in the same location as the
 	// application configuration, which means it wouldn't be found looking up by
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -114,63 +114,75 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
 // NOTE: if Viceroy is installed then it is updated, otherwise download the
 // latest version and install it in the same directory as the application
 // configuration data.
-func getViceroy(progress text.Progress, out io.Writer, versioner update.Versioner) (string, error) {
+//
+// In the case of a network failure we fallback to an installed version of the
+// Viceroy binary as long as one is installed and has the correct permissions.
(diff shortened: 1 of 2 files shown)


pkg/commands/compute/serve.go
```diff
@@ -114,63 +114,75 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
 // NOTE: if Viceroy is installed then it is updated, otherwise download the
 // latest version and install it in the same directory as the application
 // configuration data.
-func getViceroy(progress text.Progress, out io.Writer, versioner update.Versioner) (string, error) {
+//
+// In the case of a network failure we fallback to an installed version of the
```
doramatadora, line 118:
```suggestion
// In the case of a network failure we fallback to the latest installed version of the
```

pkg/commands/compute/serve.go
```diff
@@ -114,63 +114,75 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
-	}
+	}(progress)
 
-	asset := fmt.Sprintf(update.DefaultAssetFormat, versioner.Binary(), latest, runtime.GOOS, runtime.GOARCH)
-	versioner.SetAsset(asset)
-
-	bin := filepath.Join(InstallDir, versioner.Binary())
+	bin = filepath.Join(InstallDir, versioner.Binary())
 
+	// NOTE: When checking if viceroy is installed we don't use
```
doramatadora, line 131:
```suggestion
	// NOTE: When checking if Viceroy is installed we don't use
```

pkg/commands/compute/serve.go
```diff
@@ -114,63 +114,75 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	// exec.LookPath("viceroy") because PATH is unreliable across OS platforms,
+	// but also we actually install viceroy in the same location as the
```
doramatadora, line 133:
```suggestion
	// but also we actually install Viceroy in the same location as the
```

pkg/commands/compute/serve.go
```diff
@@ -114,63 +114,75 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	// application configuration, which means it wouldn't be found looking up by
+	// the PATH env var. We could pass the path for the application configuration
+	// into exec.LookPath() but it's arguably simpler to just execute the binary.
```
doramatadora, line 136:
```suggestion
	// into exec.LookPath() but it's simpler to just execute the binary.
```

doramatadora (approved):
Mostly nits

Issue #432, opened 2021-10-07
Fastly CLI `0.39.3` Breaks the `--path` flag


torch2424:
# My Use Case

I'm currently writing a Github Action, and testing with [act](https://github.com/nektos/act), and I noticed that using the latest version of the CLI, it breaks, when my local version was working. I was debugging between the two versions, and I have determined that the `--path` breaks between `0.39.2` and `0.39.3`

## Fastly CLI 0.39.2

```
| > compute-sdk-test-waiting@0.2.0 pack /home/torch2424/Source/compute-sdk-ci-github-action/compute-sdk-test-waiting
| > fastly compute pack --path bin/main.wasm
| 
✓ Initializing...
✓ Copying wasm binary...
✓ Copying manifest...
✓ Creating .tar.gz file...
| Running the C@E test for GET / ...
|  
| Deploying the waiting app....
| /home/torch2424/Source/compute-sdk-ci-github-action
| 
✓ Initializing...
✓ Uploading package...
✓ Activating version...
| 
| Manage this service at:
| 	https://manage.fastly.com/configure/services/2OFt4iXPSaKWADeT4jh4hN
| 
| View this service at:
| 	https://initially-probable-condor.edgecompute.app
| 
| 
| SUCCESS: Deployed package (service 2OFt4iXPSaKWADeT4jh4hN, version 11)
| Deployed!!
| Deploying the module....
[Compute SDK Test Backend/backendtest]   ✅  Success - Compute SDK Test Backend
```

## Fastly CLI 0.39.3

```
| > compute-sdk-test-waiting@0.2.0 pack /home/torch2424/Source/compute-sdk-ci-github-action/compute-sdk-test-waiting
| > fastly compute pack --path bin/main.wasm
| 
✓ Initializing...
✓ Copying wasm binary...
✓ Copying manifest...
✓ Creating .tar.gz file...
| Running the C@E test for GET / ...
|  
| Deploying the waiting app....
| /home/torch2424/Source/compute-sdk-ci-github-action
| 
| ERROR: error reading package manifest.
| 
| Run `fastly compute init` to ensure a correctly configured manifest. See more at https://developer.fastly.com/reference/fastly-toml/
[Compute SDK Test Backend/backendtest]   ❗  ::error::Command failed: pwd && fastly compute deploy --token *** --service-id 2OFt4iXPSaKWADeT4jh4hN --path ./compute-sdk-test-waiting/pkg/compute-sdk-test-waiting.tar.gz
[...]

PR #446, opened 2021-10-20
Implement `--path` lookup fallback for manifest

Integralist:
Fixes https://github.com/fastly/cli/issues/432

**Summary**:  
When running the CLI from a project's parent directory, then use the manifest contained within the `--path`'s archive.

**Screenshot**:  
The following screenshot demonstrates me running the CLI from a parent (non-project) directory. The CLI starts at the `testing-cli` directory (as per the `--path` value). It extracts the manifest from the archive into a temp directory and that manifest is read into memory to be used by the rest of the CLI logic that requires the manifest details.

Next I `cd` to my root user directory and then modify my `--path` flag to again point to the test archive.

Screenshot 2021-10-21 at 11 42 07

### DIFF
--- a/pkg/commands/authtoken/create.go
+++ b/pkg/commands/authtoken/create.go
@@ -101,6 +101,9 @@ func (c *CreateCommand) constructInput() *fastly.CreateTokenInput {
 		// For now, if the user ignores the `HintOptions` and gives a scope such as
 		// 'foobar' (which would be invalid), then the API itself will reject the
 		// request and the error bubbled back to our Exec caller to handle.
+		//
+		// UPDATE: We could try using `EnumsVar` which would then allow setting the
+		// same flag multiple times. For example: --scope global --scope purge_select.
 		input.Scope = fastly.TokenScope(strings.Join(c.scope, " "))
 	}
 	if len(c.services) > 0 {


torch2424:
@Integralist Thank you so much for opening this PR! 😄 I went ahead and left a reply on #432 , and depending on your thoughts there (e.g you think the second option out of my 4 is the way to go), than I'm happy to review and approve the PR 😄 🎉

pkg/commands/compute/deploy.go
```diff
@@ -354,13 +360,22 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 //
 // NOTE: It also validates if the package size exceeds limit:
 // https://docs.fastly.com/products/compute-at-edge-billing-and-resource-limits#resource-limits
-func validatePackage(data manifest.Data, pathFlag string, errLog fsterr.LogInterface) (pkgName, pkgPath string, err error) {
+func validatePackage(data manifest.Data, pathFlag string, errLog fsterr.LogInterface, out io.Writer) (pkgName, pkgPath string, err error) {
 	err = data.File.ReadError()
 	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			err = fsterr.ErrReadingManifest
+		if pathFlag == "" {
```
torch2424, line 366: Just to confirm here: We are going to read the fastly.toml in the current directory, but if we can't find it, we then use the fastly.toml in the tar.gz? 

The reason I asked, is because I realized using the fastly.toml in the tar.gz would be a breaking change, unless we make the default the current behavior, and used this as a fail safe 😄

pkg/commands/compute/deploy.go
```diff
@@ -396,6 +411,74 @@ func validatePackage(data manifest.Data, pathFlag string, errLog fsterr.LogInter
...
+	if err != nil {
+		return err
+	}
+
+	err = data.File.Read(manifestPath)
+	if err != nil {
+		return err
+	}
+
+	text.Info(out, "Using fastly.toml within --path archive:\n\t%s", pathFlag)
```
torch2424, line 443: Oh yay! I'm glad you added this so we can track this 😄 🎉

pkg/commands/compute/deploy_test.go
```diff
@@ -354,6 +354,33 @@ func TestDeploy(t *testing.T) {
 				"Deployed package (service 123, version 3)",
 			},
 		},
+		// NOTE: The following test ensures that if the user runs the CLI from a
+		// directory that isn't a C@E project directory (i.e. it has no manifest
+		// file present) then the deploy command should try to locate a manifest
+		// inside the given package tar.gz archive.
+		{
+			name: "success with path called from non project directory",
```
torch2424, line 362: Thank you for adding a test as well! 😄 🎉

torch2424 (approved):
This all looks great to me! 😄 🎉 

I just had one question, but if it works how I am asking if it works. Than I think this is good to go! (Which is why I am approving. I am pretty sure it's doing what I think it is doing haha! ) 😂 😄 🚢 :shipit:

pkg/commands/compute/deploy.go
Integralist: @torch2424 

So yes, the CLI will try to read the fastly.toml from the current directory (which might be an actual project directory, and so the file will exist, or it might be a parent directory as in your use case scenario). 

If the fastly.toml file **_is_** there in the current directory and we've been able to read it successfully, then `data.File.ReadError()` will be `nil` and so we will just skip over all this new code that tries to read the fastly.toml from inside the package .tar.gz archive file.

BUT it will only do that if the `--path` flag is set. Otherwise if the user doesn't specify a `--path` flag, then we'll just return the original error that occurred when we tried reading the fastly.toml from the user's current directory.

torch2424: Perfect! That's exactly what I had in mind! Thank you! Let's 🚢 😄 🎉

PR #453, opened 2021-10-26
Ignore possible `rustup` output in `rustc --version`. Closes #452

### LINKED ISSUE #452: Rust semantic version check fails initially on Windows
Here is the initial error...

```
integralist@DESKTOP-TEIC5O9:~/testing-rust$ fastly compute serve
✓ Initializing...
✗ Verifying local rust toolchain...

ERROR: error parsing `rustc --version` output "syncing" into a semver: Invalid Semantic Version.

```

But if I then check the output of `rustc --version` I find it to be as expected:

```
integralist@DESKTOP-TEIC5O9:~/testing-rust$ rustc --version
rustc 1.56.0 (09c42c458 2021-10-18)
```

Now if I re-run the previous command it will start compiling...

```
integralist@DESKTOP-TEIC5O9:~/testing-rust$ fastly compute serve
✓ Initializing...
```

Not sure what happened here. Maybe a Rust SME will know more about the "syncing" messaging?

fsly-apayne:
If the `rustc` executable that the CLI invokes is managed by `rustup`,
*and* the Rust toolchain that `rustup` considers active for the PWD is
not installed, then the first several lines of output are from `rustup`
installing the toolchain, and only after it finishes does `rustup`
execute the real `rustc --version` and print its output.

This patch collects and throws away all but the last output lines from
`rustc --version`, since the last line is *always* the version string.

It has been tested to work by using `fastly compute init`, changing the
`rust-toolchain` file to a named version not installed, then running
`fastly compute serve`. This caused an installation, then was able to
correctly parse the output of `rustc --version` and continue.

### DIFF
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -226,10 +226,6 @@ func rustcVersion() (string, error) {
 	// scanner quits, at which point `scanner.Text()` is the last observed line.
 	line := ""
 	for scanner.Scan() {
-		err = scanner.Err()
-		if err != nil {
-			return "", fmt.Errorf("error reading `%s` output: %w", strings.Join(cmd, " "), err)
-		}
 		line = scanner.Text()
 	}
 	err = scanner.Err()
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -222,14 +222,11 @@ func rustcVersion() (string, error) {
 	// (or any Rust distribution executable) will install it. This produces some
 	// `rustup` output before the real `rustc --version` command fires.
 	scanner := bufio.NewScanner(bytes.NewReader(stdoutStderr))
-	// As such, seek the *last* line in the scanner. This loops until the
-	// scanner quits, at which point `scanner.Text()` is the last observed line.
 	line := ""
 	for scanner.Scan() {
 		line = scanner.Text()
 	}
 	err = scanner.Err()
-	// If the scanner never fires, then something has clearly gone wrong
 	if line == "" || err != nil {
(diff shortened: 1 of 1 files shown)


pkg/commands/compute/language_rust.go
```diff
@@ -217,9 +217,20 @@ func rustcVersion() (string, error) {
...
-	reader := bufio.NewReader(bytes.NewReader(stdoutStderr))
-	line, err := reader.ReadString('\n')
-	if err != nil {
+	// Note: when `rustc` is managed by `rustup`, and the toolchain that
+	// `rustup` considers active is not installed, then the first use of `rustc`
+	// (or any Rust distribution executable) will install it. This produces some
+	// `rustup` output before the real `rustc --version` command fires.
+	scanner := bufio.NewScanner(bytes.NewReader(stdoutStderr))
+	// As such, seek the *last* line in the scanner. This loops until the
+	// scanner quits, at which point `scanner.Text()` is the last observed line.
```
Integralist, line 226: I personally don't believe the comments describing what the code does is offering any additional value to the reader, so I'd prefer to remove these lines.
```suggestion
```

pkg/commands/compute/language_rust.go
```diff
@@ -217,9 +217,20 @@ func rustcVersion() (string, error) {
...
+	line := ""
+	for scanner.Scan() {
+		line = scanner.Text()
+	}
+	err = scanner.Err()
+	// If the scanner never fires, then something has clearly gone wrong
```
Integralist, line 232: I personally don't believe this comment describing what the code does is offering any additional value to the reader, so I'd prefer to remove this line.
```suggestion
```

Integralist (approved):
LGTM. 

Thank you for fixing this.

PR #459, opened 2021-11-01
Rename `dictionaryitem` to `dictionary-item`

Integralist:
Also, remove `batchmodify` subcommand and move the logic into the `update` subcommand via the `--file` flag.


triblondon (approved):

eabquina:
Hello, it seems the `update` subcommand with the `--file` option is not working. 

`➜  ~ fastly dictionary-item update --service-id=SERVICE_ID --dictionary-id=DICTIONARY_ID --file=du.json

ERROR: the Fastly API returned 400 Bad Request: Item key can't be blank.`

PR #461, opened 2021-11-02
Rename `--force` to `--skip-verification`

Integralist:
This PR also makes two other corrections... 

1. The `fastly configure` command output when using the `--location` and `--display` flags currently displays a header (e.g. **LOCATION** and **CONFIG**) before the output. If the user only provides one of the flags then a header isn't displayed (as this actually makes it hard to use the location as part of a bash script parsing the path or piping the file content etc), but if the user provides _both_ flags then the headers are still displayed to help distinguish the output.

2. We default the `service create` command's `--type` flag to 'vcl' instead of 'wasm' as the latter isn't GA and so it doesn't make sense for it to be the default behaviour.

### DIFF
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -231,8 +231,8 @@ SUBCOMMANDS
     Create a Fastly service
 
     -n, --name=NAME        Service name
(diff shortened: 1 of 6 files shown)


kailan (approved):
Looks good 👍

PR #464, opened 2021-11-05
Implement `--watch` flag for `compute serve`

Integralist:
The following screen shot shows first my running `fastly compute serve --watch` and then immediately executing `Ctrl-C` to stop the local server. Following that I re-run the original command but in another terminal pane I edit the contents of the source code, which triggers the shutdown of the local server, then the rebuild of the project, and the restarting of the local server.

**See also a live demonstration**: https://asciinema.org/a/1Z4btLzWV5G22bTXQC9nVV99w (**NOTE**: My editor was still loading a rust LSP when I edited the file, so it actually took a second or so for the file to be updated, hence a slightly longer delay in the pane that was running the `compute serve` command from showing the message about reloading viceroy).

Screenshot 2021-11-05 at 20 23 18


Integralist:
**NOTE**: After discussing with @triblondon the language has now changed from...

```
INFO: File system modified: restarting local server
```

To...

```
INFO: Restarting: <file> has been modified
```

Example:

```
INFO: Restarting: ./src/main.rs has been modified
```

triblondon (approved):

PR #470, opened 2021-11-10
Fix error extracting package files from `.tgz` archive

Integralist:
## Problem

When trying to extract files from a `.tgz` archive, the following error was reported:

```
error creating gzip reader: gzip: invalid header.
```

## Solution

This PR refactors the original archive extraction logic to use a third-party package (already used elsewhere in the CLI). 

## Validation

The following examples work as the `--from` value:

- `https://fiddle.fastlydemo.net/fiddle/a9830b49`
- `https://storage.googleapis.com/.../a9830b49-src.tgz`
- `https://github.com/fastly/compute-starter-kit-rust-default/archive/refs/heads/main.zip`

The changes here have been tested successfully on macOS and also using various terminal environments in a Windows 10 VM (Command Prompt, Cygwin, Powershell, WSL).

Screenshot 2021-11-10 at 17 37 27

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -31,10 +31,6 @@ const (
 // https://docs.fastly.com/products/compute-at-edge-billing-and-resource-limits#resource-limits
 var PackageSizeLimit int64 = 50000000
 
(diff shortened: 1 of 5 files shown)


kailan (approved):
Working as expected! Thanks!

PR #479, opened 2021-11-23
Refactor AssemblyScript logic to call out to the JavaScript implementation

Integralist:
The majority of the AssemblyScript logic was copied over to the newer JavaScript logic. They're 99% the same. So I refactored the code such that AssemblyScript delegates to the JavaScript implementation.


pkg/config/data.go
```diff
@@ -62,6 +62,9 @@ const (
 	// ConfigRequestTimeout is how long we'll wait for the CLI API endpoint to
 	// return a response before timing out the request.
 	ConfigRequestTimeout = 5 * time.Second
+
+	// ToolchainJsDefault represents the default JS toolchain.
+	ToolchainJsDefault = "npm"
```
kailan, line 67: I think this may need to be rebased after the latest changes to the build process

kailan (changes requested):

pkg/config/data.go
Integralist: Well technically this is true still, if user provides their own scripts.build value then we don't execute the Initialize() or Verify() methods (as it's up to the user to ensure their local environment is setup). We only execute `Build()` method but will no longer use the npm default, but will use whatever the user specifies in their custom build value.

That said, I have a whole bunch of refactoring to do as part of rebasing `main` 😄

Integralist: Oh actually ignore me. I just noticed this was in the config file. So yes you're right! 😆

kailan (approved):

PR #480, opened 2021-11-23
Implement custom build operation

Integralist:
This feature allows a user to override the default verification and build steps within the CLI with whatever toolchain the user wishes to use.

For example, to avoid using NPM as the default toolchain for a JavaScript project you can set in `fastly.toml`:

```toml
[scripts]
build = "yarn run build"
```

## Example

I tested this by copying https://github.com/fastly/compute-starter-kit-javascript-default and then modifying the `fastly.toml` to include:

```toml
[scripts]
build = "yarn install && yarn run build"
```


kailan (approved):
Looks good and works on my machine! 🚢

PR #484, opened 2021-12-01
Support `compute build` for 'other' language option using `[scripts.build]`

Integralist:
Currently a project that uses an unsupported C@E language can't use the `compute build` command (which means having to do things like using `compute serve --skip-build`). The new `[scripts.build]` configuration in the fastly.toml manifest helps us to better support `compute build` for unofficial languages.

## EXAMPLE

```toml
[scripts]
build = "mkdir -p bin && tinygo build -target wasi -o bin/main.wasm main.go && fastly compute pack --wasm-binary bin/main.wasm"
```

I was able to run `fastly compute serve` and have it 'build' and 'serve' locally via Viceroy.

### DIFF
--- a/pkg/commands/compute/build.go
+++ b/pkg/commands/compute/build.go
@@ -173,21 +173,18 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	// print doesn't get hidden by the progress status.
 	progress.Done()
 
-	customBuildMsg := "This project has a custom build script defined in the fastly.toml manifest.\n"
-
 	if toolchain == "custom" {
-		if c.Globals.Flag.Verbose {
-			text.Info(out, customBuildMsg)
+		if c.Globals.Flag.Verbose || !c.Flags.AcceptCustomBuild {
+			text.Info(out, "This project has a custom build script defined in the fastly.toml manifest.\n")
 			text.Break(out)
 			text.Indent(out, 4, "%s", c.Manifest.File.Scripts.Build)
-			customBuildMsg = ""
 		}
 		if !c.Flags.AcceptCustomBuild {
 			// NOTE: A third-party could share a project with a build command for a
 			// language that wouldn't normally require one (e.g. Rust), and do evil
 			// things. So we should notify the user and confirm they would like to
 			// continue with the build.
-			label := fmt.Sprintf("\n%sAre you sure you want to continue with the build step? [y/N] ", customBuildMsg)
+			label := "\nAre you sure you want to continue with the build step? [y/N] "
 			cont, err := text.Input(out, label, in)
 			if err != nil {
 				return fmt.Errorf("error reading input %w", err)
@@ -199,6 +196,10 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			}
 			text.Break(out)
 		}
+	}
+
+	if c.Globals.Verbose() && c.Flags.AcceptCustomBuild {
+		text.Break(out)
 	}
 
 	progress = text.ResetProgress(out, c.Globals.Verbose())
--- a/pkg/commands/compute/build.go
+++ b/pkg/commands/compute/build.go
@@ -174,13 +174,14 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	// things. So we should notify the user and confirm they would like to
 	// continue with the build.
 	if !c.Flags.AcceptCustomBuild && language.Name == "other" && toolchain == "custom" {
-		label := fmt.Sprintf("This project has a custom build script defined (%s). Are you sure you want to continue with the build step? [y/N] ", c.Manifest.File.Scripts.Build)
+		label := "This project has a custom build script defined in the fastly.toml manifest. Are you sure you want to continue with the build step? [y/N] "
 		cont, err := text.Input(out, label, in)
 		if err != nil {
 			return fmt.Errorf("error reading input %w", err)
 		}
 		contl := strings.ToLower(cont)
 		if contl != "y" && contl != "yes" {
+			text.Info(out, "Stopping the build process.")
 			return nil
 		}
 	}
--- a/pkg/commands/compute/build.go
+++ b/pkg/commands/compute/build.go
@@ -169,30 +169,39 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		}
 	}
 
-	customBuildMsg := "This project has a custom build script defined in the fastly.toml manifest. "
+	// NOTE: We set the progress indicator to Done() so that any output we now
+	// print doesn't get hidden by the progress status.
+	progress.Done()
 
-	if toolchain == "custom" && c.Globals.Flag.Verbose {
-		text.Info(out, fmt.Sprintf("%s\n\n%s\n\n", customBuildMsg, c.Manifest.File.Scripts.Build))
-		customBuildMsg = ""
+	customBuildMsg := "This project has a custom build script defined in the fastly.toml manifest.\n"
+
+	if toolchain == "custom" {
+		if c.Globals.Flag.Verbose {
+			text.Info(out, customBuildMsg)
+			text.Break(out)
+			text.Indent(out, 4, "%s", c.Manifest.File.Scripts.Build)
+			customBuildMsg = ""
+		}
+		if !c.Flags.AcceptCustomBuild {
+			// NOTE: A third-party could share a project with a build command for a
+			// language that wouldn't normally require one (e.g. Rust), and do evil
+			// things. So we should notify the user and confirm they would like to
+			// continue with the build.
+			label := fmt.Sprintf("\n%sAre you sure you want to continue with the build step? [y/N] ", customBuildMsg)
+			cont, err := text.Input(out, label, in)
+			if err != nil {
+				return fmt.Errorf("error reading input %w", err)
+			}
+			contl := strings.ToLower(cont)
+			if contl != "y" && contl != "yes" {
+				text.Info(out, "Stopping the build process.")
+				return nil
+			}
+			text.Break(out)
+		}
 	}
 
-	// NOTE: A third-party could share a project with a build command for a
-	// language that wouldn't normally require one (e.g. Rust), and do evil
-	// things. So we should notify the user and confirm they would like to
-	// continue with the build.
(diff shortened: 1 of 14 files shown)


kailan (approved):
Love this!

Integralist:
If this goes ahead, then we should update [our documentation](https://developer.fastly.com/learning/compute/custom/) and also the `compute init` output so they reflect this change.

As far as [our documentation](https://developer.fastly.com/learning/compute/custom/) is concerned I think it can probably stay the same but have something like the following appended:

> **HINT:** The use of `compute pack` can be replaced by adding a `[scripts.build]` setting within your [`fastly.toml` manifest](https://developer.fastly.com/reference/compute/fastly-toml/).

Currently the `compute init` output shows...

```
Initialized package testing-cli to:
        /Users/integralist/Code/go/testing-cli

To package a pre-compiled Wasm binary for deployment, run:
        fastly compute pack

To deploy the package, run:
        fastly compute deploy
```

I imagine this should change back to the normal output:

```
Initialized package testing-cli to:
        /Users/integralist/Code/go/testing-cli

        fastly compute publish
```

Because if the user of an 'other' language project tries to build without doing anything, then they'll now see (as part of this PR) the error...

```
$ fastly compute build
✓ Initializing...
✓ Verifying local other toolchain...
✗ Building package using other toolchain...

ERROR: error reading custom build instructions from fastly.toml manifest.

Add a [scripts.build] setting for your custom build process. See more at https://developer.fastly.com/reference/fastly-toml/
```

/cc @triblondon

fgsch:
Not specific to this PR but I wonder if we want to let the user know what command we are going to run and give the user the option to abort it or require some flag to continue.  

My thinking here is that a third party could share a project with a build command for language that would normally build without it (e.g. Rust), and do evil things.

Integralist:
@fgsch I think your suggestion is valid and I'll look to add something to that effect.

Integralist:
@fgsch @kailan 

As requested I've added a prompt which informs the user a custom build has been configured (in case they were not aware of it, or what it was doing), and I ask them to confirm they want to go ahead with the build.

I've also added a new `--accept-custom-build` flag to `compute build` so that this prompt can be ignored.

Integralist:
@fgsch @kailan may I ask you both re-review, just to be sure you're still happy with the changes.

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,24 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		}
 	}
 
-	if c.Manifest.File.Scripts.Build != "" {
-		lang = "custom"
+	// NOTE: A third-party could share a project with a build command for a
+	// language that wouldn't normally require one (e.g. Rust), and do evil
+	// things. So we should notify the user and confirm they would like to
+	// continue with the build.
+	if !c.Flags.AcceptCustomBuild && language.Name == "other" && toolchain == "custom" {
```
fgsch, line 176: Wouldn't `language.Name == "other"` limit the check to other languages?

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,24 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+		cont, err := text.Input(out, label, in)
+		if err != nil {
+			return fmt.Errorf("error reading input %w", err)
+		}
+		contl := strings.ToLower(cont)
+		if contl != "y" && contl != "yes" {
+			text.Info(out, "Stopping the build process.")
+			return nil
+		}
 	}
```
fgsch, line 187: Would it be too much to always display `This project has a custom build script defined in the fastly.toml manifest` in verbose mode?

pkg/commands/compute/build_test.go
```diff
@@ -532,3 +533,139 @@ func TestBuildJavaScript(t *testing.T) {
...
+		args                 []string
+		dontWantOutput       []string
+		fastlyManifest       string
+		name                 string
+		stdin                string
+		wantError            string
+		wantOutput           []string
+		wantRemediationError string
+	}{
+		{
```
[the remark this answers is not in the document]
Integralist: done

pkg/commands/compute/build_test.go
```diff
@@ -532,3 +533,152 @@ func TestBuildJavaScript(t *testing.T) {
...
+			fastlyManifest: `
+			manifest_version = 2
+			name = "test"
+			language = "other"`,
+			dontWantOutput: []string{
+				"This project has a custom build script defined in the fastly.toml manifest",
+			},
+			wantError:            "error reading custom build instructions from fastly.toml manifest",
+			wantRemediationError: "Add a [scripts.build] setting for your custom build process",
+		},
```
fgsch, line 594: Could we also add a test for a non-other language?

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,29 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+	// NOTE: A third-party could share a project with a build command for a
+	// language that wouldn't normally require one (e.g. Rust), and do evil
+	// things. So we should notify the user and confirm they would like to
+	// continue with the build.
+	if !c.Flags.AcceptCustomBuild && toolchain == "custom" {
+		customBuildMsg := "This project has a custom build script defined in the fastly.toml manifest. "
+		if c.Globals.Flag.Verbose {
+			text.Info(out, customBuildMsg)
+			customBuildMsg = ""
+		}
```
fgsch, line 181: I wonder if we should always display this in verbose mode for custom build scripts, regardless of the AcceptCustomBuild flag.

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,29 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+		label := fmt.Sprintf("%sAre you sure you want to continue with the build step? [y/N] ", customBuildMsg)
```
fgsch, line 182: Not a blocker: for symmetry, can we check if we printed customBuildMsg above and avoid printing it again?

fgsch: Nevermind this, it's already covered.

pkg/commands/compute/build.go
Integralist: Have fixed locally, will push up shortly

pkg/commands/compute/build_test.go
Integralist: done

pkg/commands/compute/language_other.go
```diff
@@ -0,0 +1,69 @@
...
+			Inner:       err,
+			Remediation: fsterr.ComputeBuildRemediation,
+		}
+	}
+	cmd, args := o.Shell.Build(o.build)
+
+	s := fstexec.Streaming{
+		Command: cmd,
+		Args:    args,
+		Env:     []string{},
```
fgsch, line 57: Question (not a blocker): Any reason we are not passing the existing environment?

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,31 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		}
 	}
 
-	if c.Manifest.File.Scripts.Build != "" {
-		lang = "custom"
+	customBuildMsg := "This project has a custom build script defined in the fastly.toml manifest. "
```
fgsch, line 172: nitpick (not a blocker): add the newline to customBuildMsg (and zap the trailing space), remove the newline from the fmt.Sprintf() below so the prompt further below becomes:

```
This project has a custom build script defined in the fastly.toml manifest.
Are you sure you want to continue with the build step? [y/N] 
```

fgsch (approved):
I've left some non-blocking comments but this looks good to me and any remaining things can be addressed in future PRs if needed.

I still need to test it but I trust your judgement and I will let you make the call.

pkg/commands/compute/language_other.go
Integralist: I'll update this in a separate PR

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,31 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+
+	if toolchain == "custom" && c.Globals.Flag.Verbose {
+		text.Info(out, fmt.Sprintf("%s\n\n%s\n\n", customBuildMsg, c.Manifest.File.Scripts.Build))
+		customBuildMsg = ""
+	}
```
fgsch, line 177:
```suggestion
	if toolchain == "custom" && (c.Globals.Flag.Verbose || !c.Flags.AcceptCustomBuild) {
		text.Info(out, fmt.Sprintf("%s\n\n%s\n\n", customBuildMsg, c.Manifest.File.Scripts.Build))
	}
```

pkg/commands/compute/build.go
```diff
@@ -151,11 +169,31 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+
+	// NOTE: A third-party could share a project with a build command for a
+	// language that wouldn't normally require one (e.g. Rust), and do evil
+	// things. So we should notify the user and confirm they would like to
+	// continue with the build.
+	if toolchain == "custom" && !c.Flags.AcceptCustomBuild {
+		label := fmt.Sprintf("%sAre you sure you want to continue with the build step? [y/N] ", customBuildMsg)
```
fgsch, line 184:
```suggestion
		label := fmt.Sprintf("Are you sure you want to continue with the build step? [y/N] ")
```

fgsch (approved):
Thank you. This is great! 🙏

pkg/commands/compute/build.go
```diff
@@ -192,7 +192,8 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			contl := strings.ToLower(cont)
 			if contl != "y" && contl != "yes" {
 				text.Info(out, "Stopping the build process.")
-				return nil
+				text.Break(out)
+				return fsterr.ErrBuildStopped
```
fgsch, line 196: Is there any way to test this case?

fgsch (approved):
👍

pkg/commands/compute/build.go
Integralist: I actually had to update the tests to make them pass by ensuring I set an expectation for an error to be returned (see my last commit)

PR #489, opened 2021-12-15
Add `viceroy.ttl` to CLI app config

Integralist:
An internal publishing flow means changes such as https://github.com/fastly/cli/pull/488 can't be released until the CLI's application configuration is published to https://developer.fastly.com/

This PR adds a new viceroy configuration field that will be used by the aforementioned PR.


fgsch (approved):
Thank you!

PR #490, opened 2021-12-16
Fix `compute init --from` for Windows

Integralist:
Windows users would see the following error when running `compute init --from <FIDDLE>`:

```
ERROR: rename c6a72613 c6a72613.tgz: The process cannot access the file because it is being used by another process..
```


Integralist:
I've validated this is working using a Windows 10 VM.

kailan (approved):
Works as expected!

Screenshot 2022-01-04 at 11 18 12

PR #491, opened 2021-12-16
Compute languages should copy the parent environment into the subprocess

Integralist:
This is already done in `language_rust.go` but for some reason wasn't done for the other supported languages.

> NOTE**: once https://github.com/fastly/cli/pull/484 is merged, then I'll update this PR to do the same for `language_other.go`


fgsch (approved):
🚢 🇮🇹

PR #493, opened 2021-12-17
Fix the `--watch` flag for AssemblyScript

### LINKED ISSUE #476: --watch doesn't work with AssemblyScript
**Version**

1.11

**What happened**

I'm assuming this is expected behavior.

When running ```fastly compute serve --watch``` on an assemblyscript project I get this error:

```serve.go:504: CreateFile src: The system cannot find the file specified.``` (I added the file/line for your convenience)

I'm assuming this is expected behavior. I just wanted to check.

Integralist:
Fixes #476 

**Problem**: When implementing the `--watch` functionality we neglected to notice that the source code directory for AssemblyScript wasn't `src` (as it is with JavaScript and Rust) and so when it came to using the `--watch` flag we had hardcoded `"src"` to be watched.

**Solution**: We inspect the `--language` flag, falling back to the fastly.toml manifest to identify the language and use that information to identify the correct source code directory to watch.

**Screenshots**: The following screenshot shows that the compiled binary now correctly runs the `compute serve --watch` command and will (when an update is made to a file) trigger the hot reload functionality.

Screenshot 2021-12-17 at 17 08 07

### DIFF
--- a/pkg/commands/compute/build.go
+++ b/pkg/commands/compute/build.go
@@ -124,21 +124,21 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	case "assemblyscript":
 		language = NewLanguage(&LanguageOptions{
 			Name:            "assemblyscript",
-			SourceDirectory: "assembly",
+			SourceDirectory: ASSourceDirectory,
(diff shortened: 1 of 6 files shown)


triblondon (approved):
LGTM

PR #503, opened 2022-01-06
Fix Windows ANSI escape code rendering

Integralist:
Use `color.Output` instead of `os.Stdout` as it will determine the correct Windows output stream.

Also use `color.Error` in place of `os.Stderr`.

**Screenshot**

The following screenshot demonstrates both printing using `color.Output` which renders as bold, and `colour.Error` for when an error occurs.

Screenshot 2022-01-06 at 11 14 17


phamann (approved):
👍🏻

PR #505, opened 2022-01-07
Implement `--json` output for describe/list operations

### LINKED ISSUE #486: [FEATURE REQUEST] Improve tabular output for scripting
**Is your feature request related to a problem? Please describe.**
A CLI client is useful for automation tasks, but the output format of  the Fastly CLI is not easily parseable.
In particular, the tabular output generated using tabwriter (service lists, etc.) is hard to query on a command line.
The direct use case is querying the configuration of all services I have access to.

There are many possible approaches.
JSON output comes to mind, which opens the tool up for use with the popular tool `jq`, which processes JSON streams and can run complex queries on them.
Another one is to simply support tab delimiters, to allow tools (for example `awk`) to conveniently parse the table.

**Describe alternatives you've considered**
There are ways to work around this now (manual `cut -c` commands or using regular expressions), but the more convenient option is to use the API directly.
Still, I think I would prefer using the CLI for its accessible documentation and convenient interface.

**Additional context**
The direct use case is (read-only) querying the configuration of all services I have access to.

Integralist:
Fixes #486

## Examples

List backends...

```bash
$ fastly backend list --version latest

SERVICE                 VERSION  NAME        ADDRESS    PORT  COMMENT
1S3NQi1yItIGj58gDDmlWx  1        originless  127.0.0.1  80
```

List backends in verbose mode...

```bash
$ fastly backend list --version latest --verbose

Fastly API token provided via FASTLY_API_TOKEN
Service ID (via fastly.toml): 1S3NQi1yItIGj58gDDmlWx

Version: 1
        Backend 1/1
                Name: originless
                Comment:
                Address: 127.0.0.1
                Port: 80
                Override host:
                Connect timeout: 1000
                Max connections: 200
                First byte timeout: 15000
                Between bytes timeout: 10000
                Auto loadbalance: false
                Weight: 100
                Healthcheck:
                Shield:
                Use SSL: false
                SSL check cert: true
                SSL CA cert:
                SSL client cert:
                SSL client key:
                SSL cert hostname:
                SSL SNI hostname:
                Min TLS version:
                Max TLS version:
                SSL ciphers:
```

List backends causes an error when trying to use both verbose and JSON mode at the same time...

```bash
$ fastly backend list --version latest --verbose --json

Fastly API token provided via FASTLY_API_TOKEN

ERROR: invalid flag combination, --verbose and --json.

Use either --verbose or --json, not both.
```

List backends in JSON mode...

```bash
$ fastly backend list --version latest --json | jq

[
  {
    "ServiceID": "1S3NQi1yItIGj58gDDmlWx",
    "ServiceVersion": 1,
    "Name": "originless",
    "Comment": "",
    "Address": "127.0.0.1",
    "Port": 80,
    "OverrideHost": "",
    "ConnectTimeout": 1000,
    "MaxConn": 200,
    "ErrorThreshold": 0,
    "FirstByteTimeout": 15000,
    "BetweenBytesTimeout": 10000,
    "AutoLoadbalance": false,
    "Weight": 100,
    "RequestCondition": "",
    "HealthCheck": "",
    "Hostname": "",
    "Shield": "",
    "UseSSL": false,
    "SSLCheckCert": true,
    "SSLCACert": "",
    "SSLClientCert": "",
    "SSLClientKey": "",
    "SSLHostname": "",
    "SSLCertHostname": "",
    "SSLSNIHostname": "",
    "MinTLSVersion": "",
    "MaxTLSVersion": "",
    "SSLCiphers": "",
    "CreatedAt": "2022-01-07T10:54:11Z",
    "UpdatedAt": "2022-01-07T10:54:11Z",
    "DeletedAt": null
  }
]
```

### DIFF
--- a/.tmpl/create.go
+++ b/.tmpl/create.go
@@ -61,7 +61,6 @@ type CreateCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *CreateCommand) Exec(in io.Reader, out io.Writer) error {
-	// Exit early if no token configured.
 	_, s := c.Globals.Token()
 	if s == config.SourceUndefined {
 		return errors.ErrNoToken
--- a/.tmpl/delete.go
+++ b/.tmpl/delete.go
@@ -61,7 +61,6 @@ type DeleteCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *DeleteCommand) Exec(in io.Reader, out io.Writer) error {
-	// Exit early if no token configured.
 	_, s := c.Globals.Token()
 	if s == config.SourceUndefined {
 		return errors.ErrNoToken
--- a/.tmpl/describe.go
+++ b/.tmpl/describe.go
@@ -56,7 +56,6 @@ type DescribeCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *DescribeCommand) Exec(in io.Reader, out io.Writer) error {
-	// Exit early if no token configured.
 	_, s := c.Globals.Token()
 	if s == config.SourceUndefined {
 		return errors.ErrNoToken
--- a/.tmpl/list.go
+++ b/.tmpl/list.go
@@ -57,7 +57,6 @@ type ListCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *ListCommand) Exec(in io.Reader, out io.Writer) error {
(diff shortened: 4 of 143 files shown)


kailan (approved):
This is fantastic and will be so useful for CI integrations 🎉

PR #507, opened 2022-01-11
Omit unix file perms in messaging

Integralist:
Some Windows users were confused by the messaging otherwise.


doramatadora (approved):
Makes sense 👍🏼

PR #508, opened 2022-01-12
Sentry notifications

Integralist:
Sentry recommend using an [internal integration](https://github.com/marketplace/actions/sentry-release#create-a-sentry-internal-integration) over a standard auth token, but there is some red-tape involved with the former and so we're using an auth token (`SENTRY_TOKEN`) instead to trigger a new deploy.

### DIFF
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -33,3 +33,12 @@ jobs:
       env:
         GOVERSION: ${{ env.GOVERSION }}
         GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
+    - name: Create Sentry release
+      uses: getsentry/action-release@v1
+      env:
+        SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+        SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+      with:
+        environment: 'production'
+        version: '${{ github.ref_name }}'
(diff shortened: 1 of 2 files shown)


triblondon (approved):

.github/workflows/tag_release.yml
```diff
@@ -33,3 +33,12 @@ jobs:
...
         GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
+    - name: Create Sentry release
+      uses: getsentry/action-release@v1
+      env:
+        SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+        SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+      with:
+        environment: 'production'
+        version: '${{ github.ref_name }}'
```
Integralist, line 44: I tested this [here](https://github.com/Integralist/actions-testing/blob/main/.github/workflows/tag_release.yaml) and can confirm that `github.ref_name` will be the tag (e.g. `1.0.0`)

Integralist:
So it seems we _already_ have Sentry being notified of releases via the Sentry initialization in the SDK:
https://github.com/fastly/cli/blob/main/cmd/fastly/main.go#L36

This means I should probably just close this PR. But what's interesting is that we're passing in the version tag, which (for example) should be something like `1.6.0` and yet Sentry reports a completely different version number in its UI. So I've opened a support ticket with Sentry.

scripts/release-commits.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
Integralist, line 1: I tested this logic here: 
https://github.com/Integralist/actions-testing/runs/4984447215?check_suite_focus=true

Issue #512, opened 2022-01-18
Question: Unable to Port Forward "Dockerfile-Rust" fastly compute Serve


byteshiva:
I used https://github.com/fastly/cli/blob/main/Dockerfile-rust to build Docker Fastly CLI for rust

I'm unable to port forward 7676 from guests to the host environment. Am I missing something here?
```
 docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/rust   compute serve
```

Step 1: Compute build
```
➜  compute-starter-kit-rust-default git:(main) ✗ docker run -v $PWD:/app -it fastly/cli/rust   compute  build 
✓ Initializing...
✓ Verifying package manifest...
✓ Verifying local rust toolchain...
✓ Building package using rust toolchain...
✓ Creating package archive...
```

Step2: Serve Locally
```
➜  compute-starter-kit-rust-default git:(main) ✗ docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/rust   compute serve 
✓ Initializing...
✓ Verifying package manifest...
✓ Verifying local rust toolchain...
✓ Building package using rust toolchain...
✓ Creating package archive...

SUCCESS: Built package 'Default-starter-for-Rust' (pkg/Default-starter-for-Rust.tar.gz)

✓ Checking latest Viceroy release...
✓ Fetching latest Viceroy release...
✓ Running local server...

Jan 18 05:43:30.277  WARN no backend definitions found in /app/fastly.toml
Jan 18 05:43:30.277  INFO Listening on http://127.0.0.1:7676
```

Basic Curl
```
curl https://192.168.1.7:7676
curl: (7) Failed to connect to 192.168.1.7 port 7676: Connection refused

```

byteshiva:
I made it work by following the below steps.

```
docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/rust compute serve --addr="0.0.0.0:7676"  
```

```
docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/rust compute serve --addr="0.0.0.0:7676"  
✓ Initializing...
✓ Verifying package manifest...
✓ Verifying local rust toolchain...
✓ Building package using rust toolchain...
✓ Creating package archive...

SUCCESS: Built package 'Default-starter-for-Rust' (pkg/Default-starter-for-Rust.tar.gz)

✓ Checking latest Viceroy release...
✓ Fetching latest Viceroy release...
✓ Running local server...

[...]

PR #517, opened 2022-01-19
Dockerfile updates

Integralist:
- Updated Node version to `latest` as `compute serve` didn't work with older version of GLIBC. Bumping the Node version provided the updated dependency.

- User @byteshiva identified how to correctly [port forward](https://github.com/fastly/cli/issues/512#issuecomment-1015098637) so I've updated the comments in the Dockerfile to reflect this.

- Renamed the Dockerfile from `-assemblyscript` suffix to `-node` because (from my testing) the Dockerfile works with both AssemblyScript and JavaScript.


kailan (approved):
Looks great 👍

PR #519, opened 2022-01-20
Display the requested command in Sentry breadcrumb

Integralist:
Prior to this change we didn't actually know the command executed without going backwards through the stack traces. Now we record the requested command as the first step in the breadcrumb...

Screenshot 2022-01-20 at 14 49 43


kailan (approved):
Useful!

Issue #515, opened 2022-01-19
Docker: Unable to build Javascript starter kit using Docker Instance


byteshiva:
Unable to build Javascript starter kit using Docker Instance

**Steps to reproduce:**

Step 1:
get Docker file

```
wget https://raw.githubusercontent.com/fastly/cli/main/Dockerfile-assemblyscript
```
Refer:https://github.com/fastly/cli/blob/main/Dockerfile-assemblyscript

Step 2:
Build Docker 

```
docker build -t fastly/cli/javascript . -f ./Dockerfile-assemblyscript
docker run -v $PWD:/app -it fastly/cli/javascript
```

Step 3:
Initialize Javascript starter kit
```
sample-proj docker run -v $PWD:/app -it --rm fastly/cli/javascript compute init --from=https://github.com/fastly/compute-starter-kit-javascript-default  

Creating a new Compute@Edge project (using --from to locate package template).

Press ^C at any time to quit.

Name: [app] 
Description: Client
Author: Test

✓ Initializing...
✓ Fetching package template...
✓ Updating package manifest...
✓ Initializing package...

Initialized package app to:
	/app

To publish the package (build and deploy), run:
	fastly compute publish

To learn about deploying Compute@Edge projects using third-party orchestration tools, visit:
	https://developer.fastly.com/learning/integrations/orchestration/


SUCCESS: Initialized package app
```

Build javascript project
Step 4:
```
➜  sample-proj docker run -v $PWD:/app -it --rm fastly/cli/javascript compute build                        
✓ Initializing...
✓ Verifying package manifest...
✓ Verifying local javascript toolchain...
✗ Building package using javascript toolchain...

ERROR: error during execution process:
     
     > compute-starter-kit-javascript-default@0.2.0 prebuild
     > webpack
     
     assets by status 0 bytes [cached] 1 asset
     
     WARNING in configuration
     The 'mode' option has not been set, webpack will fallback to 'production' for this
     value.
     Set 'mode' option to 'development' or 'production' to enable defaults for each
     environment.
     You can also set it to 'none' to disable any default behavior. Learn more:
     https://webpack.js.org/configuration/mode/
     
     ERROR in main
     Module not found: Error: Can't resolve './src/index.js' in '/app'
     resolve './src/index.js' in '/app'
       using description file: /app/package.json (relative path: .)
         Field 'browser' doesn't contain a valid alias configuration
         using description file: /app/package.json (relative path: ./src/index.js)
           no extension
[...]

PR #521, opened 2022-01-20
create and switch to non-root user within Dockerfile

Integralist:
Tested both containers locally using:

```
docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/<rust|node> compute serve --addr="0.0.0.0:7676"
```


phamann (approved):

Integralist:
There were some issues reported in #515 with this approach but as I've been able to successfully build and run the container image using the approach documented here (and as these files are also not officially supported) I think we'll still go ahead and merge this PR.

Issue #289, opened 2021-05-27
"context deadline exceeded" - Continuous delay when running with (broken?) DNS settings


fgsch:
Version:

```
Fastly CLI version 0.30.0 (326882f)
Built with go version go1.16.4 linux/amd64
```

Description:

For some strange reason, when I have ipv6 enabled, updating the configuration fails in my box with a time out.
This causes a 5 seconds delay every time I run the CLI. This does not seem to happen if I build e.g. the v0.30.0 tag natively.
The output when this happens:

```
$ fastly -v

INFO:
Compatibility and versioning information for the Fastly CLI is being updated in the background.  The updated data will be used next time you execute a fastly command.

[..]

ERROR: there was a problem updating the versioning information for the Fastly CLI:

Get "...": context deadline exceeded.

[..]
```

When the update fails `last_checked` is not updated, so the next time the CLI is run the update process is triggered.
In my case, this means that every time I run the CLI I experience the delay as the update process keeps timing out.

Besides trying to find out why the update is timing out (I suspect it's both a network and a cross-compiling issue) it might be desirable to update the `last_checked` timestamp to avoid triggering the update continuously when there is a network issue.

fgsch:
On a related note, we might want to rephrase the error above (`Get "...": context deadline exceeded`) to something more descriptive.

fgsch:
I can reproduce this issue outside the CLI so it looks like a cross compiling problem.

fgsch:
Everything indicates this is because cgo is disabled during cross compilation. 
Quoting from https://golang.org/pkg/net/#hdr-Name_Resolution:

> By default the pure Go resolver is used, because a blocked DNS request consumes only a goroutine, while a blocked C call consumes an operating system thread. When cgo is available, the cgo-based resolver is used instead under a variety of conditions: on systems that do not let programs make direct DNS requests (OS X), when the LOCALDOMAIN environment variable is present (even if empty), when the RES_OPTIONS or HOSTALIASES environment variable is non-empty, when the ASR_CONFIG environment variable is non-empty (OpenBSD only), when /etc/resolv.conf or /etc/nsswitch.conf specify the use of features that the Go resolver does not implement, and when the name being looked up ends in .local or is an mDNS name.

Running the binary built natively under macos and setting `GODEBUG=netdns=go` shows the same (incorrect) behaviour.

peterbourgon:
@fgsch 

[...]

PR #528, opened 2022-01-31
Update config `last_checked` field even on config load error

Integralist:
If the user is suffering from https://github.com/fastly/cli/issues/289, then it means after each CLI invocation they'll have to wait N amount of seconds before the command can timeout and display an error.

The fact this happens after _every_ invocation is frustrating as it's unlikely the issue will magically resolve itself on the next time around. So instead we defer retrying the config until the config TTL has expired (default 5mins but can be extended manually).

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -145,9 +145,18 @@ Compatibility and versioning information for the Fastly CLI is being updated in 
 			// configuration file to determine where to load the config from.
 			err := file.Load(file.CLI.RemoteConfig, httpClient, config.ConfigRequestTimeout, config.FilePath)
 			if err != nil {
+				// If there is an error loading the configuration, then there is no
+				// point trying to retry the operation on the next CLI invocation as we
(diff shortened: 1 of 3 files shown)


fgsch:
Works as expected. The output is a bit verbose:

```
ERROR: there was a problem updating the versioning information for the Fastly CLI (we won't check again until 5m have passed):

Get "https://developer.fastly.com/api/internal/cli-config": context deadline exceeded.

There is a fallback version of the configuration provided with the CLI install (see `fastly configure --display`) which enables the CLI to continue to be usable even though the config couldn't be updated.
```

But maybe it's ok?

Integralist:
Thanks @fgsch 

Yeah it’s definitely verbose, but under the circumstances this issue isn't seen by everyone and it's unclear to me how much of an 'edge case' this is, so being a little bit more verbose probably isn't the worst thing.

fgsch:
@Integralist makes sense. Thank you.

harmony7 (approved):
Yes this makes sense

PR #529, opened 2022-02-03
Make `--name` flag for `service search` command a required flag

Integralist:
The `fastly service search` command has a `--name` flag which is currently not a _required_ flag, meaning go-fastly’s own internal logic will reject the request and return an error to say the required field is not set. This error is then propagated into Sentry (which can be quite noisy). 

We already ignore errors reported to Sentry that have a format of `required flag \-\-[^\s]+ not provided` (i.e. ignore any errors reported where a explicitly required flag has not been provided) and so by making the flag required in the CLI argument parser, we can avoid this error from reaching Sentry.

### DIFF
--- a/pkg/app/run_test.go
+++ b/pkg/app/run_test.go
@@ -265,7 +265,7 @@ SUBCOMMANDS
         --per-page=PER-PAGE  Number of records per page
         --sort="created"     Field on which to sort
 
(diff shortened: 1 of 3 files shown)


harmony7 (approved):
Rationale makes sense to me and the changes look good.

PR #530, opened 2022-02-03
Support Arch User Repositories

Integralist:


.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+    maintainers:
+      - 'oss@fastly.com'
+    license: "Apache license 2.0"
+    skip_upload: auto
+
+    # The SSH private key that should be used to commit to the Git repository.
+    # This can either be a path or the key contents.
+    #
+    # WARNING: do not expose your private key in the config file!
+    private_key: '{{ .Env.AUR_KEY }}'
```
Integralist, line 66: I'm not sure where this private key is supposed to come from.

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+
+    # The AUR Git URL for this package.
+    # Defaults to empty.
+    git_url: 'ssh://aur@aur.archlinux.org/cli-bin.git'
```
Integralist, line 70: Not sure what this should be set to.

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+
+    # List of packages that are not needed for the software to function,
+    # but provide additional features.
+    #
+    # Must be in the format `package: short description of the extra functionality`.
+    #
+    # Defaults to empty.
+    optdepends:
+      - 'viceroy: for running service locally'
```
Integralist, line 79: Don't _think_ this is needed because the CLI installs Viceroy directly from the GitHub repo, and I don't believe it requires anything other than the go net/http package to get the content so we also shouldn't need to define [`depends`](https://goreleaser.com/customization/aur/).

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+        #
+    # Custom package instructions.
+    #
+    # Defaults to `install -Dm755 "./PROJECT_NAME" "${pkgdir}/usr/bin/PROJECT_NAME",
+    # which is not always correct.
+    #
+    # We recommend you override this, installing the binary, license and
+    # everything else your package needs.
+    package: |-
```
Integralist, line 88: I don't know if this is needed, but they suggest it _should_ be updated.

Integralist (commented):
@mccurdyc some comments that hopefully you can help with. Please feel free to take over this PR as I think you are better suited to resolve them.

.goreleaser.yml
mccurdyc: I think we just need to generate a new SSH key using `ssh-keygen` similar to how GitHub works. I actually just recently found out that no two GitHub accounts can share an SSH key because of how ssh auth works with git and that's the same here - https://docs.github.com/en/authentication/troubleshooting-ssh/error-key-already-in-use

mccurdyc: Here are the GitHub docs on generating a key (I never remember how and always have to look at these docs :smile: ) - https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

.goreleaser.yml
mccurdyc: Agreed! I had this in my local one too I think.

.goreleaser.yml
mccurdyc: I think we may want this to be `fastly-bin`. Because I think folks will search AUR for "fastly". And would probably expect to see `fastly`, `fastly-bin` or `fastly-git` is the pattern if you provide all three.

`fastly-bin` would be if you just served a binary. I think `fastly-git` would be if you were going to clone and build from source. And I can't remember what the plain `fastly` would be.

mccurdyc: I always reference https://github.com/Jguer/yay-PKGBUILD because `yay` is currently the most common tool for interacting with the AUR.

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+      # bin
+      install -Dm755 "./mybin" "${pkgdir}/usr/bin/mybin"
```
mccurdyc, line 90: `mybin` being the `fastly` bin that is created somewhere, in CI?

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+
+      # license
+      install -Dm644 "./LICENSE.md" "${pkgdir}/usr/share/licenses/mybin/LICENSE"
```
mccurdyc, line 93: We probably want to replace `mybin` everywhere with `fastly` or `fastly-git` or whatever.

.goreleaser.yml
mccurdyc: So to be clear, I think you are correct using the `-bin` suffix here.

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+      # completions
+      mkdir -p "${pkgdir}/usr/share/bash-completion/completions/"
+      mkdir -p "${pkgdir}/usr/share/zsh/site-functions/"
+      mkdir -p "${pkgdir}/usr/share/fish/vendor_completions.d/"
+      install -Dm644 "./completions/mybin.bash" "${pkgdir}/usr/share/bash-completion/completions/mybin"
+      install -Dm644 "./completions/mybin.zsh" "${pkgdir}/usr/share/zsh/site-functions/_mybin"
+      install -Dm644 "./completions/mybin.fish" "${pkgdir}/usr/share/fish/vendor_completions.d/mybin.fish"
+
+      # man pages
+      install -Dm644 "./manpages/mybin.1.gz" "${pkgdir}/usr/share/man/man1/mybin.1.gz"
```
mccurdyc, line 104: Do we have man pages?

.goreleaser.yml
```diff
@@ -50,14 +50,66 @@ archives:
...
+      # license
+      install -Dm644 "./LICENSE.md" "${pkgdir}/usr/share/licenses/mybin/LICENSE"
+
+      # completions
+      mkdir -p "${pkgdir}/usr/share/bash-completion/completions/"
+      mkdir -p "${pkgdir}/usr/share/zsh/site-functions/"
+      mkdir -p "${pkgdir}/usr/share/fish/vendor_completions.d/"
+      install -Dm644 "./completions/mybin.bash" "${pkgdir}/usr/share/bash-completion/completions/mybin"
+      install -Dm644 "./completions/mybin.zsh" "${pkgdir}/usr/share/zsh/site-functions/_mybin"
+      install -Dm644 "./completions/mybin.fish" "${pkgdir}/usr/share/fish/vendor_completions.d/mybin.fish"
```
mccurdyc, line 101: Do we have completions?

.goreleaser.yml
mccurdyc: I don't think we should set this unless we need to override it. 

Defaults to `install -Dm755 "./PROJECT_NAME" "${pkgdir}/usr/bin/PROJECT_NAME"` seems pretty good to me to start.

mccurdyc (commented):
Referenced the gorelease AUR docs while reviewing this - https://goreleaser.com/customization/aur/

And https://wiki.archlinux.org/title/PKGBUILD

Also quickly browsed - https://github.com/goreleaser/goreleaser/blob/a1447a363579365f489458ad7636fd088a5b66ab/internal/pipe/aur/aur.go

I talked to @Integralist and I'm going to own this branch/PR from here and test it out on my Arch machine.

.goreleaser.yml
Integralist: Nope. We only have help via either `fastly help` or `fastly <command> --help` so I guess this can come out.

.goreleaser.yml
Integralist: We do! Here's how Homebrew sets up the completions (in case it's helpful):
https://github.com/fastly/homebrew-tap/blob/master/Formula/fastly.rb

I also added a test to the CLI for validating the completion output (again, in case it's of interest/help):
https://github.com/fastly/cli/blob/main/pkg/app/run_test.go#L62-L167

.goreleaser.yml
Integralist: Let's go with `fastly` 👍🏻

.goreleaser.yml
Integralist: Yeah I'm not sure where about goreleaser is going to stick the compiled binary 🤔

mccurdyc:
Reading the "Rules of Submission" section of https://wiki.archlinux.org/title/AUR_submission_guidelines

- [x] checked that there is not an existing `fastly` package - https://archlinux.org/packages/?sort=&q=fastly&maintainer=&flagged=
- [x] "Packages that use prebuilt [deliverables](https://en.wikipedia.org/wiki/Deliverable), when the sources are available, must use the `-bin` suffix...The AUR should not contain the binary tarball created by makepkg, nor should it contain the filelist."
    - https://github.com/fastly/cli/pull/530#discussion_r798837269
- [x] "Please add a comment line to the top of the PKGBUILD file which contains information about the current maintainers"
    - https://github.com/fastly/cli/pull/530/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689R57 I'm assuming goreleaser handles this.
        - https://github.com/goreleaser/goreleaser/blob/17aba5ce76cd925f1299df4d1efc99c6c93b9040/internal/pipe/aur/tmpl.go#L27 and https://github.com/goreleaser/goreleaser/blob/17aba5ce76cd925f1299df4d1efc99c6c93b9040/internal/pipe/aur/aur.go#L163. So yep!
- [ ] Going through "Authentication" section of https://wiki.archlinux.org/title/AUR_submission_guidelines 
    ```
    ssh-keygen -f ~/.ssh/aur -C noname
    ```
    And uploaded to a Fastly secret storage system and shared with @Integralist's team.
    
    Added the following to my `~/.ssh/config` file
    
    ```
   Host aur.archlinux.org
     IdentityFile ~/.ssh/aur
     User aur
    ```

mccurdyc:
```
makepkg -p fastly-bin.pkgbuild
==> Making package: fastly-bin 1.6.0-1 (Wed 09 Feb 2022 08:23:41 AM EST)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Downloading fastly_v1.6.0_linux-amd64.tar.gz...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   671  100   671    0     0   3384      0 --:--:-- --:--:-- --:--:--  3388
100 5901k  100 5901k    0     0  5363k      0  0:00:01  0:00:01 --:--:-- 9894k
==> Validating source_x86_64 files with sha256sums...
    fastly_v1.6.0_linux-amd64.tar.gz ... FAILED
==> ERROR: One or more files did not pass the validity check!
```

Need to fix this.

mccurdyc (approved):
Approving because you are the PR author. I did do another pass and things make sense.

.github/workflows/tag_release.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: "Run GoReleaser"
       uses: goreleaser/goreleaser-action@v2
       with:
-        version: v0.183.0 # goreleaser version (NOT goreleaser-action version)
+        version: v1.4.1 # goreleaser version (NOT goreleaser-action version)
         args: release --rm-dist
       env:
         GOVERSION: ${{ env.GOVERSION }}
```
Integralist, line 34:
```suggestion
        AUR_KEY: ${{ secrets.AUR_KEY }}
        GOVERSION: ${{ env.GOVERSION }}
```

.github/workflows/tag_release.yml
```diff
@@ -28,9 +28,10 @@ jobs:
     - name: "Run GoReleaser"
       uses: goreleaser/goreleaser-action@v2
       with:
-        version: v0.183.0 # goreleaser version (NOT goreleaser-action version)
+        version: v1.4.1 # goreleaser version (NOT goreleaser-action version)
```
Integralist, line 31:
```suggestion
        version: v1.5.0 # goreleaser version (NOT goreleaser-action version)
```

PR #531, opened 2022-02-07
Implement Compute@Edge Free Trial Activation

Integralist:
There's two flows when it comes to the C@E free trial activation…

1. Service needs to be created.
2. Service already exists.

For the first scenario, if the service fails to be created, then we attempt to activate the trial and recall the function that creates the service.

For the second scenario I've added a validation step to check if the given service ID is associated with a Wasm service (e.g. I call the service details API endpoint and check the ‘type’ field is set to ‘wasm’). If that fails, then it means the user provided a VCL Service. The error we return indicates we need a Wasm service, not a VCL service and the user should consult fastly.help/cli/ecp-feature.

**NOTE**: This PR includes code changes not directly related to the free trial activation but helped to clean the code up.


pkg/commands/compute/deploy_test.go
```diff
@@ -160,18 +165,20 @@ func TestDeploy(t *testing.T) {
 			name: "service version is active, clone version error",
 			args: args("compute deploy --service-id 123 --token 123 --version 1"),
 			api: mock.API{
-				CloneVersionFn: testutil.CloneVersionError,
-				ListVersionsFn: testutil.ListVersions,
+				CloneVersionFn:      testutil.CloneVersionError,
+				GetServiceDetailsFn: getServiceDetailsWasm,
```
Integralist, line 169: There were quite a few existing tests that needed to be fixed so they mocked `GetServiceDetailsFn`, which was a new API call I added for validating the user's service type.

pkg/commands/compute/deploy_test.go
```diff
@@ -225,6 +232,98 @@ func TestDeploy(t *testing.T) {
 				"Creating service...",
 			},
 		},
+		// The following test mocks the service creation to fail with a specific
```
Integralist, line 235: Here are a bunch of new tests that validate different failure and success scenarios related to the trial activation.

pkg/commands/compute/deploy.go
```diff
@@ -624,6 +684,28 @@ func manageExistingServiceFlow(
 		return serviceVersion, err
 	}
 
+	// Validate that we're dealing with a Compute@Edge 'wasm' service and not a
```
Integralist, line 687: I realised when dealing with a user who provides their own service, that we never actually validate the service is a compatible wasm service type. So I added that validation step here...

Integralist (commented):
Some comments to help the reviewer know where to look...

pkg/api/undocumented/undocumented.go
```diff
@@ -0,0 +1,57 @@
+package undocumented
```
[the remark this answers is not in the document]
Integralist: For whatever reason I'm not overly keen on the name `undocumented`, even though it's an accurate description of the functionality. So I'm open to suggestions.

I tried using `internal` prior to this, but the go compiler recognises that name, and will force certain structural constraints that didn't really work for this use case.

kailan (approved):
This is great! Thanks for the detailed comments, made it really easy to review.

PR #539, opened 2022-03-09
OAuth Support

Integralist:
**Problem**: 
Users of the Fastly CLI currently have to manually log into the Fastly UI to create an access token, once created they need to run fastly configure where they’ll be prompted to enter the access token so it can be persisted to disk and reused for future commands. 

This process is tedious, time consuming, and is not ideal due to users typically creating tokens with no expiration date instead of using short-lived tokens that are regenerated periodically

**Solution**:
When running the Fastly CLI, the first operation should be to look for an existing short-lived access token within the CLI configuration file and, if one exists, check if it has expired. 

If the access token has expired or doesn't exist, then a new authentication flow will be initiated, otherwise the pre-existing access token will be utilised for the current command being executed.

The authentication flow will notify the user of what is about to happen and prompt the user to continue. Once the flow proceeds the CLI will open the user’s default web browser and direct it to a Fastly authentication service.

The full token flow will be opaque to the user, and they should only have to worry about their initial username/password login to the authentication service while the CLI takes care of the rest of the flow, persisting the access token to disk and managing its lifecycle.

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -110,9 +110,9 @@ func Run(opts RunOpts) error {
 		return nil
 	}
 
-	token, source := globals.Token()
+	token, tSource := globals.Token()
 	if globals.Verbose() {
-		switch source {
+		switch tSource {
 		case config.SourceFlag:
 			fmt.Fprintf(opts.Stdout, "Fastly API token provided via --token\n")
 		case config.SourceEnvironment:
@@ -127,7 +127,7 @@ func Run(opts RunOpts) error {
 	// If we are using the token from config file, check the files permissions
 	// to assert if they are not too open or have been altered outside of the
 	// application and warn if so.
-	if source == config.SourceFile && name != "configure" {
+	if tSource == config.SourceFile && name != "configure" {
 		if fi, err := os.Stat(config.FilePath); err == nil {
 			if mode := fi.Mode().Perm(); mode > config.FilePermissions {
 				text.Warning(opts.Stdout, "Unprotected configuration file.")
@@ -138,22 +138,26 @@ func Run(opts RunOpts) error {
 		}
 	}
 
-	if auth.Required(name, token, source, opts.Stdout) {
-		err = auth.Init(opts.Stdout)
-		if err != nil {
-			return err
-		}
-	}
-
-	endpoint, source := globals.Endpoint()
+	endpoint, eSource := globals.Endpoint()
 	if globals.Verbose() {
-		switch source {
+		switch eSource {
 		case config.SourceEnvironment:
 			fmt.Fprintf(opts.Stdout, "Fastly API endpoint (via %s): %s\n", env.Endpoint, endpoint)
 		case config.SourceFile:
 			fmt.Fprintf(opts.Stdout, "Fastly API endpoint (via config file): %s\n", endpoint)
 		default:
 			fmt.Fprintf(opts.Stdout, "Fastly API endpoint: %s\n", endpoint)
+		}
+	}
+
+	if auth.Required(name, token, tSource, opts.Stdout) {
+		token, err = auth.Init(opts.Stdout)
+		if err != nil {
+			return err
+		}
+		err = auth.PersistToken(token, globals.Path, endpoint, &globals.File, opts.APIClient)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -29,7 +29,7 @@ type Versioners struct {
 
 // RunOpts represent arguments to Run()
 type RunOpts struct {
-	APIClient  APIClientFactory
+	APIClient  api.APIClientFactory
 	Args       []string
 	ConfigFile config.File
 	ConfigPath string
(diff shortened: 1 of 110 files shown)


pkg/app/run.go
```diff
@@ -149,7 +151,23 @@ func Run(opts RunOpts) error {
...
+	if auth.Required(name, token, endpoint, tSource, opts.Stdout, opts.ClientFactory) {
+		token, err = auth.Init(opts.Stdin, opts.Stdout, auth.Browser, opts.AuthService)
+		if err != nil {
+			return err
+		}
+		err = auth.PersistToken(token, globals.Path, endpoint, &globals.File, opts.ClientFactory)
+		if err != nil {
+			return err
+		}
+	}
```
Integralist, line 163: Here is where we add calls out to the new authentication logic.

If `auth.Required` returns true, then we'll start a new authentication flow (and persist the newly generated token back to the user's local CLI configuration file).  

Otherwise we'll continue on executing the command using the existing token we have.

pkg/auth/auth.go
```diff
@@ -0,0 +1,81 @@
...
+			return true
+		}
+		if ok := ValidateToken(c, stdout); !ok {
+			required = true
+		}
+	}
+
+	// False (no authentication flow required).
+	return required
+}
```
Integralist, line 42: Here is where we determine if an authentication flow is required.

- If the command doesn't use a token, then _**no**_ an auth flow is not required.
- If no token exists in the CLI config, then _**yes**_ an auth flow is required.
- If a token exists in the CLI config, but is invalid or expired, then _**yes**_ an auth flow is required.

This means, if a user provides a token via `--token` then we skip the authentication flow.  
This allows for greater flexibility, and as a side-effect gives us an easy way to skip auth in our test suite.

pkg/auth/auth.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", err
+	}
+
+	t := <-token
+	if t == "" {
+		return t, errors.New("no token received from authentication service")
+	}
+
+	return t, nil
+}
```
Integralist, line 81: Here is where we initialise the authentication flow.

- We prompt the user to confirm they want to go ahead with opening their web browser.
- We start up a local HTTP server to handle the auth service callback.
- We block the command execution until we've received a token from the auth service.

pkg/auth/server.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := r.URL.Query().Get("auth_error"); err == "" {
+			msg = err
+		}
+		fmt.Fprintf(w, label, msg)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprint(w, "Authentication complete. You may close this browser tab and return to your terminal.")
+}
```
Integralist, line 67: Here is where we handle the auth service callback.

pkg/auth/token.go
```diff
@@ -0,0 +1,62 @@
...
+		return false
+	}
+
+	if (*t.ExpiresAt).Before(time.Now()) {
+		text.Info(stdout, "Your access token has expired.")
+		return false
+	}
+
+	return true
+}
```
Integralist, line 34: Here is where we validate a pre-existing token.

The token is read from the CLI config file and is going to be either a long-lived token (added by the user via `fastly configure` before the CLI supported OAuth) or a short-lived token generated by the new authentication flow.

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+	wantToken := "123"
+	if f.User.Token != wantToken {
+		t.Errorf("want token '%s', have '%s'", wantToken, f.User.Token)
+	}
+	wantEmail := "test@example.com"
+	if f.User.Email != wantEmail {
+		t.Errorf("want email '%s', have '%s'", wantEmail, f.User.Email)
+	}
+}
```
Integralist, line 65: Here is where we validate a successful authentication flow.

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,390 @@
...
+		t.Fatal(err)
+	}
+
+	if f.User.Token != "" {
+		t.Errorf("want no token, have '%s'", f.User.Token)
+	}
+	if f.User.Email != "" {
+		t.Errorf("want no email, have '%s'", f.User.Email)
+	}
+}
```
Integralist, line 107: Here is where we validate the use of `--token` skips the authentication flow but also that the given token doesn't accidentally get persisted to disk (we only persist the token to disk if we've either gone through the authentication flow or if the user has run the `fastly configure` command).

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+	wantToken := "456" // token in config was 123 and should be replaced
+	if f.User.Token != wantToken {
+		t.Errorf("want token '%s', have '%s'", wantToken, f.User.Token)
+	}
+	wantEmail := "test@example.com"
+	if f.User.Email != wantEmail {
+		t.Errorf("want email '%s', have '%s'", wantEmail, f.User.Email)
+	}
+}
```
Integralist, line 160: Here is where we validate a successful authentication flow for a user who had a pre-existing token that was long-lived (i.e. not generated via the authentication service).

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,390 @@
...
```
Integralist, line 213: Here is where we validate a successful authentication flow for a user who had a pre-existing token that was generated via the authentication service, but had since expired.

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,390 @@
...
+	opts.AuthService = <-endpoint
+
+	err := app.Run(opts)
+	if err == nil {
+		t.Log(stdout.String())
+		t.Error("expected error, got nil")
+	}
+
+	testutil.AssertErrorContains(t, err, "no token received from authentication service")
+}
```
Integralist, line 240: Here is where we validate a failed authentication flow.

Integralist (commented):
Here is a breakdown of areas to focus on...

Integralist:
I'm going to close this PR for the time being as internal discussions have meant we might need to revise our implementation.

Issue #540, opened 2022-03-10
SIGSEGV on fastly backend create (v1.7.0)


jfoote:
**Version**

```
% fastly version
Fastly CLI version 1.7.0 (6150a45)
Built with go version go1.17.7 linux/amd64
```

**What happened**

Hello. I got a SIGSEGV when attempting to run `fastly backend create`. See details below.

```
fastly backend create --version=latest --name=<REDACTED> --address=<REDACTED_IPV4_ADDRESS> --port=8080
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13e52df]

goroutine 1 [running]:
github.com/fastly/kingpin.(*uintValue).Set(0xc000206528, {0x20671656d, 0x19acbc0})
	/home/runner/go/pkg/mod/github.com/fastly/kingpin@v2.1.12-0.20191105091915-95d230a53780+incompatible/values_generated.go:679 +0x3f
github.com/fastly/kingpin.(*Application).setValues(0xc0003c0ea0, 0xc0004cc160)
	/home/runner/go/pkg/mod/github.com/fastly/kingpin@v2.1.12-0.20191105091915-95d230a53780+incompatible/app.go:479 +0x354
github.com/fastly/kingpin.(*Application).Parse(0xc0003c0ea0, {0xc000030080, 0x19ac978, 0x4})
	/home/runner/go/pkg/mod/github.com/fastly/kingpin@v2.1.12-0.20191105091915-95d230a53780+incompatible/app.go:240 +0xd0
github.com/fastly/cli/pkg/app.processCommandInput({0x19f1b48, {0xc000030080, 0x6, 0x6}, {{{0xc00024a280, 0x34}, {0xc00020a298, 0x2}, {0xc0002221e0, 0x19}, ...}, ...}, ...}, ...)
	/home/runner/work/cli/cli/pkg/app/usage.go:319 +0x6b4
github.com/fastly/cli/pkg/app.Run({0x19f1b48, {0xc000030080, 0x6, 0x6}, {{{0xc00024a280, 0x34}, {0xc00020a298, 0x2}, {0xc0002221e0, 0x19}, ...}, ...}, ...})
	/home/runner/work/cli/cli/pkg/app/run.go:95 +0x952
main.main()
	/home/runner/work/cli/cli/cmd/fastly/main.go:183 +0xa58
```

OS details follow

```
% sw_vers
ProductName:	macOS
[...]

PR #542, opened 2022-03-14
Fix runtime panic in arg parser

Integralist:
Fixes https://github.com/fastly/cli/issues/540

This was caused by a nil pointer dereference. 

The solution is to pass a non-pointer type to the arg parser. We are then able to (after the CLI arg parser has run) inspect whether the value was provided, and if so, set the corresponding Go-Fastly struct field.

> **NOTE**: I've checked all other Go-Fastly CREATE input types and there doesn't appear to be any other instances that need to be corrected within the CLI.


doramatadora (approved):
LGTM!

Issue #142, opened 2020-07-09
Multiple switchable profiles


jregistr:
👋  Hellos!

I created this issue https://github.com/fastly/cli/issues/141 earlier sharing some thoughts. I wanted to ask if y'all have anything in mind or planned for folks using multiple API Keys and Service IDs.
I'm currently taking care of this with some functions in my bash profile, like so:
```bash
dev() {
  export FASTLY_API_KEY = '...'
  export FASTLY_SVC_ID = '...'
}
stg () {......}
```

In a way, this is similar to "profiles" on the AWS CLI.

So I could see something like a `$HOME/.config/fastly.toml` config containing these.

That being said, I understand this tool is quite new, and this probably something way at the back of the roadmap or even thoughts.

StarpTech:
Hi, what's the status? I also have multiple accounts and it would be great to deal with them as with multiple was accounts. This is very basic functionality.

Integralist:
Hi @jregistr thanks for raising this issue, and thanks @StarpTech for popping in to check for a status update!

So we're currently in the process of evolving the authentication model used by the CLI. Once that work is complete (worst case in a couple of weeks), then we can start looking into supporting multiple profiles. 

We already have an internal ticket related to this work, but it has not yet been prioritised. That said, there's been interest both internally and externally for this feature so I could see this getting scheduled maybe in the next quarter. 

Unfortunately I can't guarantee that will happen at this stage, but we are aware of this feature as being an important one for our users so it's a likely next candidate following the CLI's authentication model update.

Integralist:
👋🏻 @StarpTech 

I've managed to get the ball rolling on multiple profiles: https://github.com/fastly/cli/pull/546

I've more work to do on this, and it's very much a WIP, but in case you were interested in an update 🙂

Integralist:
The PR to add profiles (https://github.com/fastly/cli/pull/546) has been approved, but isn't yet merged. 

@StarpTech if you had any opinions on the change of interface, now would be the time to comment in the PR comments 🙂

PR #546, opened 2022-03-24
Support multiple profiles via `[profiles]` configuration

Integralist:
Fixes https://github.com/fastly/cli/issues/142

**This PR introduces breaking changes, which will result in a new major version release (`v2.0.0`).**

This PR replaces the `fastly configure` command with:

- `fastly config`
- `fastly profile`

> **NOTE**: This PR also introduces a new global `--profile` flag for temporarily switching the user's profile for the duration of the command execution.

## Migrating existing user data

The configuration file was previously defined as:

```toml
[user]
  email = ""
  token = ""
```

The data in these fields will be auto-migrated to a generic 'user' profile:

```toml
[profile]

  [profile.user]
    default = true
    email = "<REDACTED>"
    token = "<REDACTED>"
```

## Screenshots of new commands

Screenshot 2022-04-04 at 15 26 02

Screenshot 2022-04-04 at 15 28 11

## Screenshots of usage

**The following screenshots do not capture all scenarios or use cases, for that refer to the code and test suite.**

List profiles so we can see there are none defined yet...

Screenshot 2022-04-04 at 15 29 14

Create a profile for the first time, it will automatically be set to being the 'default' profile...

Screenshot 2022-04-04 at 15 29 52

Create a new profile, and set it to be the default...

Screenshot 2022-04-04 at 15 30 28

List profiles so we can be sure the new 'personal' profile created has been set to be the default...

Screenshot 2022-04-04 at 15 31

Switching profiles, and then checking the profile was switched...

Screenshot 2022-04-04 at 15 34 36

Delete the profile that is currently the 'default', notice the warning provided...

Screenshot 2022-04-04 at 15 39 22

Update a profile but give an unknown profile name...

Screenshot 2022-04-04 at 15 40 36

Update a profile, and also set it to be the default...

Screenshot 2022-04-04 at 15 41 23

Use the global `--profile` flag to temporarily switch profiles, but notice we provide an unrecognised profile name...

Screenshot 2022-04-04 at 15 43 50


.gitignore
```diff
@@ -1,5 +1,8 @@
+# ignore fastly binary
+#
+# NOTE: This can cause fuzzy finders to fail to locate cmd/fastly/main.go
+# because gitignore cannot negate files inside an ignore directory.
 fastly
-!cmd/fastly
```
Integralist, line 2: Just to clarify: I'm removing this line as per the note above (i.e. it didn't actually negate the ignored item).

I also Googled for ways to ignore binaries and unfortunately there's no easy way to do that.

Makefile
```diff
@@ -26,14 +26,14 @@ fastly:
...
-	@go build -gcflags="all=-N -l" -o "fastly" ./cmd/fastly
+	@go build -gcflags="all=-N -l" $(LDFLAGS) -o "fastly" ./cmd/fastly
 
 .PHONY: all
 all: config tidy fmt vet staticcheck gosec test build install
 
 .PHONY: dependencies
 dependencies:
-	go install github.com/securego/gosec/cmd/gosec@latest
+	go install github.com/securego/gosec/v2/cmd/gosec@latest
```
Integralist, line 36: Bumped so we can support the latest Go 1.18 release that supports Generics.

Makefile
```diff
@@ -26,14 +26,14 @@ fastly:
 
 # useful for attaching a debugger such as https://github.com/go-delve/delve
 debug:
-	@go build -gcflags="all=-N -l" -o "fastly" ./cmd/fastly
+	@go build -gcflags="all=-N -l" $(LDFLAGS) -o "fastly" ./cmd/fastly
```
Integralist, line 29: Added in the `LDFLAGS` throughout this file because otherwise when running `fastly version` with a local dev build of the CLI you would just get empty output.

Here's the difference between me running `make install` on the current `main` branch vs building on this PR's branch...

Screenshot 2022-04-01 at 09 15 54

pkg/app/commands.go
```diff
@@ -97,7 +97,7 @@ func defineCommands(
 	computeServe := compute.NewServeCommand(computeCmdRoot.CmdClause, globals, computeBuild, opts.Versioners.Viceroy, data)
 	computeUpdate := compute.NewUpdateCommand(computeCmdRoot.CmdClause, globals, data)
 	computeValidate := compute.NewValidateCommand(computeCmdRoot.CmdClause, globals)
-	configureCmdRoot := configure.NewRootCommand(app, opts.ConfigPath, configure.APIClientFactory(opts.APIClient), globals)
+	configureCmdRoot := configure.NewRootCommand(app, configure.APIClientFactory(opts.APIClient), globals)
```
Integralist, line 100: Removed this argument because the value is already exposed as part of the `globals` argument.

pkg/app/run.go
```diff
@@ -12,8 +12,9 @@ import (
 	"github.com/fastly/cli/pkg/commands/version"
 	"github.com/fastly/cli/pkg/config"
 	"github.com/fastly/cli/pkg/env"
-	"github.com/fastly/cli/pkg/errors"
+	fsterr "github.com/fastly/cli/pkg/errors"
```
Integralist, line 15: We should never have imported our own `errors` package without a custom namespace because it gets very confusing when you actually want to use the go standard library's own `errors` package.

So whenever I stumble across code that doesn't use the `fsterr` custom namespace I make sure to update it.

pkg/app/run.go
```diff
@@ -123,6 +131,48 @@ func Run(opts RunOpts) error {
...
+			if err != nil {
+				return err
+			}
+			if !cont {
+				return nil
+			}
+			token = p.Token
+		}
+	}
+
```
Integralist, line 175: **TODO**: Not essential for the PR review. 

I'm thinking of refactoring this chunk of code into the new `profile` package.

pkg/config/data.go
```diff
@@ -181,13 +191,13 @@ type File struct {
...
 	Fastly        Fastly              `toml:"fastly"`
 	Language      Language            `toml:"language"`
+	Profiles      Profiles            `toml:"profile"`
 	StarterKits   StarterKitLanguages `toml:"starter-kits"`
-	User          User                `toml:"user"`
 	Viceroy       Viceroy             `toml:"viceroy"`
 
 	// We store off a possible legacy configuration so that we can later extract
 	// the relevant email and token values that may pre-exist.
-	Legacy LegacyFile `toml:"legacy"`
```
Integralist, line 190: The `user` field in the config.toml is now considered 'legacy' and will be auto-migrated to the new `[profiles]` section. 

A long time ago the `user` field was considered _new_ and the _very original_ config.toml format had a user's email and token in the top level file. So when we first switched to `user` we introduced a `legacy` field into the toml. 

That was a long time ago and we don't need to support `legacy` any more.

pkg/profile/profile.go
```diff
@@ -0,0 +1,85 @@
+package profile
```
Integralist, line 1: For reference, and quick scanning, this is the new `profile` package that contains the logic for handling profiles.

Integralist (commented):
@doramatadora There are a lot of scenarios to cover and so I've tried to use screenshots in the PR description to help visualise the features this PR is introducing. Otherwise hopefully the following comments help to digest the relevant code changes.

pkg/app/run.go
Integralist: This refactor is now pushed: https://github.com/fastly/cli/pull/546/files/#diff-7e180686f4acd352dbcfd7ac75355f83218e23e9dd38699bb0ac8f9029903175R95

pkg/commands/profile/create.go
```diff
@@ -0,0 +1,229 @@
...
+	displayCfgPath(c.Globals.Path, out)
+	text.Success(out, "Profile '%s' created", c.profile)
+	return nil
+}
+
+// tokenFlow initialises the token flow.
+func (c *CreateCommand) tokenFlow(profile string, def bool, in io.Reader, out io.Writer) error {
+	var err error
+
+	// If user provides --token flag, then don't prompt them for input.
```
doramatadora, line 72:
```suggestion
	// If user provides a --token flag, then don't prompt them for input.
```

doramatadora (approved):
This is very exciting to see ❤️ 🚀

pkg/commands/compute/init.go
```diff
@@ -121,13 +123,20 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	}
 	c.dir = dst
 
-	name, desc, authors, err := promptOrReturn(c.manifest, c.dir, c.Globals.File.User.Email, in, out)
+	// Assign the default profile email if available.
```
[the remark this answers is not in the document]
Integralist: There may well be no profiles created if it's a new CLI user and they go straight to wanting to use the C@E features. 

The 'email' variable here is actually just used as the 'authors' value in the fastly.toml and isn't required to be filled in. 

The user is always prompted to enter an author value, but if we're able to find the email from their token then we display that as the 'default' value, in case they enter nothing at the prompt that asks for the package author. Otherwise if there's no profile data to get an email from the is no default displayed and it just means no value is assigned to the author fields in the fastly.toml file.

PR #550, opened 2022-04-13
Fix `compute deploy --package` 'no manifest' bug.

Integralist:
**Before**
Trying to deploy a project via the `--package` flag would cause an error because the code logic would try to update a manifest file in the user's current directory to have a `service_id` field. But as the user has provided the `--package` flag, this indicates the user isn't in a project directory and is trying to deploy a prior build from some other location.

Screenshot 2022-04-13 at 14 10 16

**After**
The deploy will now succeed because we skip the manifest update when dealing with a manifest from within a package archive.

Screenshot 2022-04-13 at 14 10 32


kailan (approved):
Looks good 👍

PR #551, opened 2022-04-13
Fix `compute serve --watch` to keep watching after build error.

Integralist:
Currently the CLI will stop completely when an error is encountered during a build, but the compilation error could well be because a user made a typo error in their code, and so we should keep watching the source code even when the compilation fails.

> **NOTE**: This only relates to the build compilation that occurs when file changes are detected. If the initial build fails then the CLI will still fail outright.

## Screenshots

Notice there was a compilation error during the development cycle:

Screenshot 2022-04-13 at 17 57 39

But the Fastly CLI will now continue to watch for file changes:

Screenshot 2022-04-13 at 17 57 46


kailan (approved):
Big win for developer experience, thanks!

PR #554, opened 2022-04-17
Skip checking for updates for completion commands

thommahoney:
I have my shell (ZSH) configured to support completion of `fastly` CLI commands. To do this, I have placed the following into my ~/.zshrc:

```shell
eval "$(fastly --completion-script-zsh)"
```

When on high-latency internet connections, the asynchronous update check adds 3 seconds to the loading of my shell. Furthermore, because the check fails with such spotty internet, I see the following message before my shell prompt following the 3 second timeout:

```
ERROR: there was a problem updating the versioning information for the Fastly CLI (we won't check again until 5m have passed):

Get "https://developer.fastly.com/api/internal/cli-config": context deadline exceeded (Client.Timeout exceeded while awaiting headers).

There is a fallback version of the configuration provided with the CLI install (run `fastly config` to view the config) which enables the CLI to continue to be usable even though the config couldn't be updated.
```

Instead of filing a bug, I thought I'd try to fix it myself. ~I'm not sure if this actually works because I can't build the source code given my spotty airplane wi-fi.~

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/fastly/cli/pkg/app"
 	"github.com/fastly/cli/pkg/check"
+	"github.com/fastly/cli/pkg/cmd"
 	"github.com/fastly/cli/pkg/commands/update"
 	"github.com/fastly/cli/pkg/config"
 	fsterr "github.com/fastly/cli/pkg/errors"
@@ -141,7 +142,10 @@ func main() {
 	var errLoadConfig error
 
 	// Validate if configuration is older than its TTL
-	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) {
+	// NOTE: We don't want to trigger a config check when the user is making an
+	// autocomplete request because this can add additional latency to the user's
+	// shell loading completely.
+	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) && !cmd.IsCompletion(args) && !cmd.IsCompletionScript(args) {
 		if verboseOutput {
 			text.Info(out, `
 Compatibility and versioning information for the Fastly CLI is being updated in the background.  The updated data will be used next time you execute a fastly command.


pkg/app/run.go
```diff
@@ -174,7 +175,7 @@ func Run(opts RunOpts) error {
 		return fmt.Errorf("error constructing Fastly realtime stats client: %w", err)
 	}
 
-	if opts.Versioners.CLI != nil && name != "update" && !version.IsPreRelease(revision.AppVersion) {
+	if opts.Versioners.CLI != nil && name != "update" && !version.IsPreRelease(revision.AppVersion) && !cmd.IsCompletion(opts.Args) && !cmd.IsCompletionScript(opts.Args) {
```
Integralist, line 178:
```suggestion
	// NOTE: We don't want to trigger a config check when the user is making an 
	// autocomplete request because this can add additional latency to the user's 
	// shell loading completely.
	if opts.Versioners.CLI != nil && name != "update" && !version.IsPreRelease(revision.AppVersion) && !cmd.IsCompletion(opts.Args) && !cmd.IsCompletionScript(opts.Args) {
```

Integralist (changes requested):
This looks fine to me but I don't think it will fix your issue with the error:

> 'there was a problem updating the versioning information...'

Because that error looks to come from a separate call to load the config here:

https://github.com/fastly/cli/blob/23068722b1f802c09c9043c52484ffb21a1598ce/cmd/fastly/main.go#L133-L165

thommahoney:
Weird I didn't see the contents of `checkAgain` in my output but perhaps I was running an outdated version of the CLI. Is the CLI config useful in the context of shell completion? If not, perhaps we should wrap that in a similar conditional.

Integralist:
@thommahoney 

>  Is the CLI config useful in the context of shell completion?

Nope. I don't think having the config would be useful so we should try to avoid it when handling the shell autocomplete command. 

That said, the logic for checking whether the config is stale is happening in `cmd/fastly/main.go` and the logic for identifying the current command is happening in `pkg/app/run.go`. 

So you could go big and refactor the main func so that the stale check is moved into the run func, or maybe because we're only checking for a single flag, `--completion-script-<T>`, maybe just tack on an extra conditional check to keep things simple.

thommahoney:
I was able to reproduce the latency and error output by installing the Network Link Conditioner tooling from Apple. I made an "Extremely Bad Network" configuration to test out this functionality: 

Screen Shot 2022-05-27 at 11 08 49 AM

In order to properly test this, one must override the `cli.last_checked` value in the config.toml to something greater than 5 minutes in the past.

Using this technique, I arrived at the latest commit: f7be6c3e099f0faf139f601b69fd0aa45b28c23f

Integralist (approved):

PR #555, opened 2022-04-19
`pkg/commands/compute/serve.go`: `--watch` to support ignore files

Integralist:
This PR adds the ability for the CLI to respect the following ignore files (in order of precedence)...

- `.ignore` (local)
- `.gitignore` (local)
- `~/.gitconfig`'s `core.excludesfile` (global)

> **NOTE**: Although `git` also consults `$GIT_DIR/info/exclude` I've not included this as (in my experience) it's a rarely used location for configuring ignore patterns.

If there are no ignore files present, the CLI will default to watching all files inside the project's `./src` directory.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,10 @@ require (
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 )
 
-require github.com/otiai10/copy v1.7.0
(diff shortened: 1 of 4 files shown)


kailan (approved):
This is a really helpful feature, and the code is well commented. Great work!

PR #557, opened 2022-05-12
Prevent profile data from being unexpectedly deleted

Integralist:
**PROBLEM**: 
Users have reported that their profile data is getting removed from the CLI config file.

**SOLUTION**:
As the root cause of the bug is unclear, I've opted to add extra error handling to hopefully surface the cause, as well as utilise a mutex to prevent overlapping calls to functions that would change the in-memory data representation.

**NOTES**: 
This appears to only affect a select number of users, and isn't isolated to any one command.

We already have a mutex around writing to the config file so there should be no data race with regards to that segment of the logic flow. 

My initial thoughts were that it's likely either the `update.CheckAsync` function, which calls `config.File.Read`, OR the asynchronous updating of the config (when the data is considered stale) when calling `config.File.Load`.

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -2,14 +2,13 @@ package main
 
 import (
 	_ "embed"
-	"log"
-	"time"
-
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/fastly/cli/pkg/app"
 	"github.com/fastly/cli/pkg/check"
@@ -88,7 +87,7 @@ func main() {
 	// Extract a subset of configuration options from the local application directory.
 	var file config.File
 	file.SetStatic(cfg)
-	err = file.Read(config.FilePath, in, out)
+	err = file.Read(config.FilePath, in, out, fsterr.Log)
 
 	if err != nil {
 		if err == config.ErrLegacyConfig {
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -127,6 +127,7 @@ func main() {
 	waitForWrite := make(chan bool)
 	wait := false
 
+	// errLoadConfig should be a RemediationError.
 	var errLoadConfig error
 
 	// Validate if configuration is older than its TTL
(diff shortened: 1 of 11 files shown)


cmd/fastly/main.go
```diff
@@ -88,7 +87,7 @@ func main() {
 	// Extract a subset of configuration options from the local application directory.
 	var file config.File
 	file.SetStatic(cfg)
-	err = file.Read(config.FilePath, in, out)
+	err = file.Read(config.FilePath, in, out, fsterr.Log)
```
Integralist, line 90: Updated the method signature so that we can internally record any errors to our `errors.log` file on disk.

pkg/commands/update/check_test.go
```diff
@@ -136,7 +137,16 @@ func TestCheckAsync(t *testing.T) {
...
-			f := update.CheckAsync(ctx, testcase.file, configFilePath, testcase.currentVersion, testcase.cliVersioner, in, &out)
+			f := update.CheckAsync(
+				ctx,
+				testcase.file,
+				configFilePath,
+				testcase.currentVersion,
+				testcase.cliVersioner,
+				in,
+				&out,
+				fsterr.MockLog{},
```
Integralist, line 148: Passing through a mock error log instance and multi-lining this method call for readability.

pkg/config/data.go
```diff
@@ -167,13 +169,16 @@ func (d *Data) Endpoint() (string, Source) {
...
+
 // FilePath is the location of the fastly CLI application config file.
 var FilePath = func() string {
 	if dir, err := os.UserConfigDir(); err == nil {
-		return filepath.Join(dir, "fastly", "config.toml")
+		return filepath.Join(dir, "fastly", FileName)
 	}
 	if dir, err := os.UserHomeDir(); err == nil {
-		return filepath.Join(dir, ".fastly", "config.toml")
+		return filepath.Join(dir, ".fastly", FileName)
```
Integralist, line 181: There's a few places in the code where I need to reference the string `"config.toml"` so it made sense to move it to a package level constant.

pkg/config/data.go
```diff
@@ -342,7 +347,18 @@ func (f *File) Load(endpoint, path string, c api.HTTPClient) error {
 		return fmt.Errorf("fetching remote configuration: expected '200 OK', received '%s'", resp.Status)
 	}
 
+	// NOTE: In an attempt to prevent unexpected changes to the in-memory data
```
Integralist, line 350: There are quite a few places throughout the CLI code where we call functions that ultimately end up mutating the in-memory data representation of the config (and in a couple of cases those calls happen asynchronously). 

Because the root cause of the bug causing profile data to be deleted is unclear, I wanted to take a 'belt and braces' approach and to synchronise any access to the in-memory data.

Integralist (commented):
Some comments to help whoever is reviewing this code...

triblondon (approved):
LGTM.  Thanks for inline comments :-)

PR #567, opened 2022-05-27
Support Viceroy "inline-toml" `format` and new `contents` field.

fgsch:
Make `file` optional as well.  This is an interim solution while a
more robust one is investigated that is `format` aware, that's it,
so either `file` or `contents` are added even if empty, depending
on whether the `format` is `json` or `inline-toml`, respectively.


fgsch:
Tests are missing. Happy to have a chat about a better solution.

PR #574, opened 2022-06-01
Hash main.wasm and not the package

### LINKED ISSUE #243: Compute Edge hashsum should be the hash of main.wasm
Currently we hash the tarball, which might contain a newly generated `main.wasm` if `compute build` was issued before.
Furthermore, the tarball does not preserve the dates and times of e.g. `Cargo.toml` or `fastly.toml`, resulting on different hashes every time the tarball is created. 

To avoid this situation, and only create new revisions when the binary changes, we want to change the code to only compute and use the hash of `main.wasm`.

CC @phamann

fgsch:
This moves the hash calculation into validate(), now validateAndHash(),
to avoid opening the package twice.

Fixes #243.

### DIFF
--- a/pkg/commands/compute/validate.go
+++ b/pkg/commands/compute/validate.go
@@ -31,7 +31,7 @@ func (c *ValidateCommand) Exec(in io.Reader, out io.Writer) error {
 		return fmt.Errorf("error reading file path: %w", err)
 	}
 
-	if err := validate(p); err != nil {
+	if err := validate(p, nil); err != nil {
 		c.Globals.ErrLog.AddWithContext(err, map[string]interface{}{
 			"Path": c.path,
 		})
@@ -48,14 +48,19 @@ type ValidateCommand struct {
 	path string
 }
 
+// FileValidator validates a file.
+type FileValidator func(archiver.File) error
+
 // validate is a utility function to determine whether a package is valid.
 // It attemptes to unarchive and read a tar.gz file from a specfic path,
 // if successful, it then iterates through (streams) each file in the archive
 // checking the filename against a list of required files. If one of the files
 // doesn't exist it returns an error.
+// validate also call fileValidator, if not nil, passing the file obtained from
+// tar.Read().
 //
 // NOTE: This function is also called by the `deploy` command.
-func validate(path string) error {
+func validate(path string, fileValidator FileValidator) error {
 	file, err := os.Open(filepath.Clean(path))
 	if err != nil {
 		return fmt.Errorf("error reading package: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/commands/compute/validate.go
```diff
@@ -31,7 +31,7 @@ func (c *ValidateCommand) Exec(in io.Reader, out io.Writer) error {
 		return fmt.Errorf("error reading file path: %w", err)
 	}
 
-	if err := validate(p); err != nil {
+	if err := validate(p, nil); err != nil {
```
Integralist, line 34: I won't consider this a breaking change, but I'm personally not a fan of having to pass `nil`. This was why I had suggested a variadic function signature. 

But I appreciate this makes the code's intention clearer as far as we only ever expect there to be a _single_ validation step, so I can live with it.

pkg/commands/compute/validate.go
```diff
@@ -89,9 +94,17 @@ func validate(path string) error {
...
+		}
+
+		closeErr := f.Close()
+
 		if err != nil {
-			return fmt.Errorf("error closing package: %w", err)
+			return err
+		}
+		if closeErr != nil {
+			return fmt.Errorf("error closing file: %w", closeErr)
```
Integralist, line 107: I think the `if err != nil` check can be moved inside of the `if fileValidator != nil` statement,  
and `closeErr` renamed to just `err`.

So something like:

```go
if fileValidator != nil {
	if err := fileValidator(f); err != nil {
		return err
	}
}

err = f.Close()
if err != nil {
	return fmt.Errorf("error closing file: %w", err)
}
```

fgsch: The reason for this is to avoid having to call `f.Close()` in two different places.
Are you saying that you want to omit the call, or that you are happy with the call being within the `fileValidator` error path?

fgsch: To be clear, the alternative I wanted to avoid was:

```go
if fileValidator != nil {
  if err := fileValidator(f); err != nil {
    f.Close()
    return err
  }
}
```

Integralist: OK that's understandable. I actually missed the fact that you would have had to close in two places.

So what would be the _practical_ outcome of not closing the file when we're in a CLI environment? My understanding is the moment the process stops the associated file descriptors will be automatically cleaned up.

If this was a long running server process, then I'd agree to always explicitly calling `f.Close()` but for a CLI I don't think it matters as much.

Also from a user experience, I personally don't feel it's a useful error to report back as a user can't do anything about it if it did fail so maybe if we do want to close the file we instead use `defer` and at just report it into the error log using `c.Globals.ErrLog`.

PR #583, opened 2022-06-24
Fix `compute publish --non-interactive`

### LINKED ISSUE #582: `fastly compute publish --non-interactive` still requires user input
**Is your feature request related to a problem? Please describe.**
I was wanting to setup a temporary fastly service for automated testing using `fastly compute publish -i` however when running this with a fastly.toml that has no service_id defined, the command is wanting user input even though the `-i` flag is set, which I think is meant to indicate "accept default values"

When running `fastly compute publish -i` with a fastly.toml that has ho service_id defined, I would like the command to accept the default values - which would be a randomly assigned domain and either no defined backends or backends defined from the setup field of fastly.toml

**Describe alternatives you've considered**
I could write a program which spawns a child-process of `fastly compute publish -i` and send "enter" inputs to the child-process, which would make the default values become accepted but this feels like a workaround.

**Additional context**
I like the idea of this feature as it means a test suite could do these steps:
- Create temporary service -- `fastly compute publish -i`
- Get the domain assigned to the service -- `fastly domain list --version latest --json | jq '.[0].Name'`
- Run your tests here
- Teardown the temporary service -- `fastly service delete --force`

Integralist:
Fixes #582

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -154,6 +154,7 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	domains := &setup.Domains{
 		APIClient:      apiClient,
 		AcceptDefaults: c.Globals.Flag.AcceptDefaults,
+		NonInteractive: c.Globals.Flag.NonInteractive,
 		PackageDomain:  c.Domain,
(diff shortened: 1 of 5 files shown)


JakeChampion (approved):
💃

PR #588, opened 2022-06-29
Handle build info more accurately

06kellyjac:
`runtime.Version()` will always give an accurate build host version

https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/runtime/extern.go;l=251-266

For the host's GOOS and host's GOARCH they're not available so they must still be passed in via ldflags

matches the `go version` cli output

https://github.com/golang/go/blob/master/src/cmd/go/internal/version/version.go#L76

`make fastly` now uses goreleaser. I removed the cleanup comment and left the ldflags for the "dev" builds
Maybe the whole LDFLAGS entry can be removed and the dev builds just dont have version info?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,40 +5,47 @@ SHELL := /bin/bash -o pipefail
 # the output is consistent across environments.
 VERSION ?= $(shell git describe --tags 2>/dev/null || git rev-parse --short HEAD)
 
+# allows for passing additional args to the build e.g.
+# make build GO_ARGS='--ldflags "-s -w"'
+GO_ARGS ?= ""
+
 # Enables support for tools such as https://github.com/rakyll/gotest
 TEST_COMMAND ?= go test
 
 # The compute tests can sometimes exceed the default 10m limit.
-TESTARGS ?= -timeout 15m ./{cmd,pkg}/...
+TEST_ARGS ?= -timeout 15m ./{cmd,pkg}/...
 
 CLI_ENV ?= "development"
 
-# TODO: This is duplicated in .goreleaser and we should figure out how to clean
-# this up so we have one source of truth.
-LDFLAGS = -ldflags "\
- -X 'github.com/fastly/cli/pkg/revision.AppVersion=${VERSION}' \
- -X 'github.com/fastly/cli/pkg/revision.GitCommit=$(shell git rev-parse --short HEAD || echo unknown)' \
- -X 'github.com/fastly/cli/pkg/revision.GoVersion=$(shell go version)' \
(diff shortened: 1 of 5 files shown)


pkg/revision/revision.go
```diff
@@ -32,8 +45,17 @@ func init() {
...
+	if GoHostOS == "" {
+		GoHostOS = "unknown"
+	}
+	if GoHostArch == "" {
+		GoHostArch = "unknown"
+	}
 	if GoVersion == "" {
-		GoVersion = "go version unknown"
+		// runtime.Version() provides the Go tree's version string at build time
+		// the other values like OS and Arch aren't accessable and are passed in
```
Integralist, line 56:
```suggestion
		// the other values like OS and Arch aren't accessible and are passed in
```

Makefile
```diff
@@ -13,19 +13,22 @@ TESTARGS ?= -timeout 15m ./{cmd,pkg}/...
...
  -X 'github.com/fastly/cli/pkg/revision.GitCommit=$(shell git rev-parse --short HEAD || echo unknown)' \
- -X 'github.com/fastly/cli/pkg/revision.GoVersion=$(shell go version)' \
+ -X 'github.com/fastly/cli/pkg/revision.GoHostOS=${GOHOSTOS}' \
+ -X 'github.com/fastly/cli/pkg/revision.GoHostArch=${GOHOSTARCH}' \
  -X 'github.com/fastly/cli/pkg/revision.Environment=${CLI_ENV}' \
  "
 
  GO_FILES = $(shell find cmd pkg -type f -name '*.go')
 
 fastly: $(GO_FILES)
```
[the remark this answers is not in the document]
06kellyjac: sure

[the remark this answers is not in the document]
06kellyjac: I tweaked the example slightly

pkg/revision/revision.go
06kellyjac: Addressed spelling mistake & updated LDFLAG info at the top of the file

Makefile
```diff
@@ -13,19 +13,28 @@ TESTARGS ?= -timeout 15m ./{cmd,pkg}/...
 
 CLI_ENV ?= "development"
 
-# TODO: This is duplicated in .goreleaser and we should figure out how to clean
```
06kellyjac, line 16: @Integralist is this TODO addressed or in the interest of de-duplicating the LDFLAGs should I completely remove LDFLAGS used for the dev builds (using `go build ..` directly)

PR #589, opened 2022-06-29
Fix version check

06kellyjac:
Fixing a failure I introduced

Not sure if this is the best way to have this test without exactly re-implementing the revision.go stuff

cc @Integralist


Integralist (approved):
Thanks! This looks OK to me.

PR #594, opened 2022-07-12
Compute@Edge TinyGo Support

Integralist:


.fastly/config.toml
```diff
@@ -8,6 +8,10 @@ remote_config = "https://developer.fastly.com/api/internal/cli-config"
 ttl = "5m"
 
 [language]
+  [language.go]
+    tinygo_constraint = ">= 0.24.0"
+    toolchain_constraint = ">= 1.17"
```
fgsch, line 13: Do we want to also include an upper limit here?

Integralist: We don't for other languages so I'd be hesitant to introduce one here unless we feel there is likely to be a breaking change result from using (for example go 1.19)

fgsch: YMMV but I think we should play it safe for the first release.

cmd/fastly/static/config.toml
```diff
@@ -21,42 +21,67 @@ rustup_constraint = ">= 1.23.0"
...
-[[starter-kits.javascript]]
-name = "Default starter for JavaScript"
-description = "A basic starter kit that demonstrates routing, simple synthetic responses and overriding caching rules."
-path = "https://github.com/fastly/compute-starter-kit-javascript-default"
-
-[[starter-kits.javascript]]
-name = "Empty starter for JavaScript"
-description = "An empty application template for the Fastly Compute@Edge environment which simply returns a 200 OK response."
-path = "https://github.com/fastly/compute-starter-kit-javascript-empty"
+[starter-kits]
```
triblondon, line 24: What's the benefit in having this section heading?  It looks like all TOML sections have to be fully qualified, so I'm wondering what this does.

triblondon (approved):
Quite a lot of JS related changes here too?

Integralist:
@triblondon 

> Quite a lot of JS related changes here too?

Just some general refactoring, no logic/behaviour changes.

cmd/fastly/static/config.toml
Integralist: @triblondon So to clarify, this change was me manually updating the static config for testing purposes. When I come to merge this PR the `static/config.toml` will instead be updated via the `config` Makefile target, which pulls the latest version of the config from the Developer Hub (DevHub). 

This means we need the (internal) DevHub content PR merged first before merging this PR so that the go starter kit will be included in the generated remote config endpoint. Once that's available, then the CLI's CI process will check that the static config in this PR matches it. 

The top level `[starter-kits]` section isn't present in the DevHub's remote CLI config endpoint (simply because we didn't need to be so verbose, it instead only adds the fully qualified version like `[[starter-kits.<LANGUAGE>]]`) but it's worth realising that the TOML parser generates an AST with `starter-kits` as a top-level section and so when the CLI makes changes to the user's _local_ CLI config, the format of the local config matches this more verbose version you see here (that's why you see the verbose formatting here because when developing this branch I manually copied my local config into the static/backup config -- that will disappear once I switch to the `make config` version).

PR #595, opened 2022-07-12
update config

Integralist:
In preparation for a new Compute@Edge language, I'm updating the CLI configuration.

Once this PR is merged we can initiate an internal build process to ensure the CLI config endpoint starts to return the updated config.

This means the PR for integrating the new C@E language will not have a conflict with the 'static' config baked in (which gets automatically updated via a `make config` target that overrides the static config with whatever happens to be exposed via the remote config endpoint). As we need that PR to have the static config contain the new language fields.

### DIFF
--- a/.fastly/config.toml
+++ b/.fastly/config.toml
@@ -10,7 +10,7 @@ ttl = "5m"
 [language]
   [language.go]
     tinygo_constraint = ">= 0.24.0"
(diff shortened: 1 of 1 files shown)


fgsch (approved):
My comment regarding adding an upper limit still applies here but LGTM 👍

Issue #597, opened 2022-07-12
😵‍💫 Starter kit `Cargo.toml` loses its formatting


acfoltzer:
**Version**

```
❯ fastly version
Fastly CLI version v3.1.1 (59efd38)
Built with go version go1.18.3 linux/amd64
Viceroy version: viceroy 0.2.14
```

**What happened**

Ran `fastly compute init` and chose the empty Rust starter kit. The resulting `Cargo.toml` looked like:

```toml

[dependencies]
  fastly = "^0.8.2"

[package]
  edition = "2021"
  name = "my-new-project"
  publish = false
  version = "0.1.0"

[profile]

  [profile.release]
    debug = 1
```

While semantically correct, the formatting does not match the [starter kit](https://github.com/fastly/compute-starter-kit-rust-empty/blob/8644490363da58f50f448988f2cfb6fe79b6a48a/Cargo.toml#L1-L13)'s formatting and comments. The resulting formatting, alphabetized with indentation, is extremely surprising to a Rust programmer.

fgsch:
I think we can close this now.

PR #599, opened 2022-07-13
Fix Cargo.toml formatting

fgsch:
Set the package name in the Cargo.toml without parsing the file to
avoid any side effects.

Should address #597.

### DIFF
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -59,15 +59,16 @@ func (m *CargoManifest) Read(path string) error {
 }
 
 // SetPackageName into Cargo.toml manifest.
-//
-// NOTE: We can't presume to know the structure of the Cargo manifest, and so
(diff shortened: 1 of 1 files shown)


pkg/commands/compute/language_rust.go
```diff
@@ -65,17 +65,14 @@ func (m *CargoManifest) Read(path string) error {
...
-	tree.Set("package.name", name)
-
-	data, err := tree.Marshal()
+	data, err := os.ReadFile(path)
 	if err != nil {
-		return err
+		return fmt.Errorf("error reading Cargo.toml file: %w", err)
 	}
-
+	data = bytes.ReplaceAll(data, []byte(m.Package.Name), []byte(name))
```
Integralist, line 75: You'll want to remove the NOTE from the function documentation.

fgsch: Indeed. Thank you. Updated.

Integralist (approved):

PR #609, opened 2022-07-27
Don't include manifest in package

Integralist:
We shouldn't include the language manifests (e.g. `package.json`, `Cargo.toml`, `go.mod`) in the built package as in the case of golang there might not always be one (as we've discovered with some internal projects).


kailan (approved):
Perhaps `IncludeFiles` can be removed in the future, but happy to approve this for now!

Integralist:
Agreed, might be worth removing in future but for now I'll leave it in 👍🏻

fgsch:
Thanks. I agree with this change but perhaps we want to make sure other folks are aware? 
I doubt anyone is using this but won't be the first time I'm wrong :)

fgsch (approved):
👍

PR #612, opened 2022-07-28
ignore TTL & update config if app version doesn't match config version

joeshaw:
Fixes an issue where the local config did not have the Go starter kit
after updating until at least 5 minutes afterward.

The issue went something like this:
1. You run `fastly update` to get the latest software
2. The config gets updated to the latest version, but unsupported features (like the new Go support) are dropped when the config gets reserialized to disk
3. The software gets updated
4. The config file gets updated again (from #610), but it's still the old software version so there's no change
5. The config file has its last checked time updated to now
6. You run the new software and its check as to whether to update the config file sees that it's been updated within the TTL (5m) and doesn't update it
7. The Go starter kit is sill missing

This PR fixes the issue by having the runtime check for whether the config needs to be updated ignore the TTL when the version in the config file does not match the currently-running CLI version.  That way, the first time you run the CLI after updating, it will always update the config file first.

The change from #610 should probably be reverted?  It doesn't seem to have any effect since the same update happens immediately before downloading the software.  I wasn't 100% sure on that and didn't include it in this PR but I can add it if it's the right thing to do.

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -145,7 +145,8 @@ func main() {
 	// NOTE: We don't want to trigger a config check when the user is making an
 	// autocomplete request because this can add additional latency to the user's
 	// shell loading completely.
-	if check.Stale(file.CLI.LastChecked, file.CLI.TTL) && !cmd.IsCompletion(args) && !cmd.IsCompletionScript(args) {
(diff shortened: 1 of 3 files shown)


joeshaw:
Looks like a spurious CI failure but I don't have the perms to restart them.

fgsch (approved):
Thank you!

I think we should revert #610 but it's your call. We can always do it in a separate PR.

fgsch (approved):
Re-approving 👍

PR #615, opened 2022-07-28
when updating config due to version change, block command on it

joeshaw:
Normally the CLI updates the config in the background for subsequent
executions.  However, if the version changes we need to get the updated
config in place before we execute the command.

### DIFF
--- a/cmd/fastly/main.go
+++ b/cmd/fastly/main.go
@@ -141,16 +141,28 @@ func main() {
 	// errLoadConfig should be a RemediationError.
 	var errLoadConfig error
 
-	// Validate if configuration is older than its TTL
+	// If the version stored in the config doesn't match our current version,
+	// block running of the command until the new config is downloaded.
+	waitBeforeRun := file.CLI.Version != revision.AppVersion
+
+	// Validate if configuration is older than its TTL or this is the
+	// first run after an update.
 	// NOTE: We don't want to trigger a config check when the user is making an
 	// autocomplete request because this can add additional latency to the user's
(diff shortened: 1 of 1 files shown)


joeshaw:
An alternative approach I explored was having the `fastly update` command execute the new binary as a test, and rely on it writing out the config because the versions don't match.

```diff
diff --git pkg/commands/update/root.go pkg/commands/update/root.go
index d832861..a434f82 100644
--- pkg/commands/update/root.go
+++ pkg/commands/update/root.go
@@ -5,6 +5,7 @@ import (
        "fmt"
        "io"
        "os"
+       "os/exec"
        "path/filepath"

        "github.com/fastly/cli/pkg/cmd"
@@ -122,6 +123,15 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
                }
        }

+       progress.Step("Testing binary...")
+       if err := exec.Command(currentPath, "version").Run(); err != nil {
+               c.Globals.ErrLog.AddWithContext(err, map[string]interface{}{
+                       "Executable": currentPath,
+               })
+               progress.Fail()
+               return fmt.Errorf("error running new binary: %w", err)
+       }
+
        progress.Done()

        text.Success(out, "Updated %s to %s.", currentPath, latest)
```

The problem with this is that there's a race because the old code could write out the config file after the new code does because it's done in the background.

Integralist (approved):
LGTM. I think for these occasions (i.e. when updating to a new CLI version) it's OK to block for the config 👍🏻

PR #620, opened 2022-08-01
Remove dynamic configuration

Integralist:
## Summary

We're removing the dynamic configuration feature/logic from the CLI.

We will continue to have a static configuration file embedded into the CLI binary that will be used as a failover in case reading from the user's modified configuration file on disk fails.

For every CLI release the build process will update the static config that is embedded into the CLI using the existing DevHub remote config endpoint. The reason for continuing to use the DevHub remote config endpoint is that the DevHub has an existing build process that generates the config with the latest list of starter kits that are available. This means we don’t need to manually curate that list of starter kits ourselves.

## Context
The CLI introduced a dynamic configuration loader in Jan 2021. This was intended to ease the number of releases required while Fastly's Rust SDK (ABI) stabilised. It also enabled Fastly to dynamically generate configurations such as the language Starter Kits (rather than having to remember to update them manually) and to add anything else we might want to control remotely in future.

## Actual usage
We initially made a few changes to the config dynamically, but then the Rust SDK ABI stabilized and we never updated the config dynamically for anything else, and we never added anything new to the config because adding anything to the config meant a CLI code change to support that new configuration setting.

## Problems
What has become apparent is a constant list of issues caused by the dynamic configuration. 

The CLI is dependent on its configuration, but fetching the config on every command invocation is not performant for users and is a waste of resources. So we introduced a TTL to the configuration so that the CLI could check if the config was stale before making a request to fetch the config from the remote endpoint. This use of a TTL in practice has caused lots of user complaints regarding invalid configuration where new CLI features were added and released but the user didn’t get the updated configuration because their version of the config was not considered stale. 

There have been various permutations of this issue in the CLI since its inception and subsequently, many different fixes have been put in place to try and resolve these issues, but we continue to have problems with the dynamic config. 

We’ve recently (as a hotfix for the TinyGo release) put a blocking call for the configuration into the CLI. This isn’t ideal because if the fetching of the config fails, then we’ll not update the ‘version’ in the configuration and subsequently each invocation will block until the configuration request is successful. 

We’ve also historically run into many customer reports of the config request failing constantly due to their DNS resolver being configured incorrectly (this is still a problem to this day as it’s not something the CLI can fix, being a user problem, but the dynamic config highlights this problem and thus causes reports to continue to come in that need to be investigated, managed and explained each time).

There’s also the added complexity in the CLI logic to handle all these cases, which makes it much harder to reason about a new release and what a user might experience when we have runtime issues with the dynamic configuration.


pkg/config/data.go
```diff
@@ -463,63 +390,43 @@ func (f *File) ValidConfig(verbose bool, out io.Writer) bool {
...
-				return readErr
-			}
-			return retry.RetryableError(readErr)
-		}
+		msg := "unable to load your configuration data"
+		label := fmt.Sprintf("We were %s. %s", msg, replacement)
 
-		return nil
-	})
+		cont, err := text.AskYesNo(out, label, in)
```
Integralist, line 417: @fgsch I might need to double check that this doesn't affect CIs that use `--non-interactive` or related flags.

kailan (approved):
Looks good to me, and I'm glad to see so much unpredictability being removed. Ship it 🚀✨

pkg/config/data.go
```diff
@@ -463,63 +390,43 @@ func (f *File) ValidConfig(verbose bool, out io.Writer) bool {
...
+func (f *File) Read(
+	path string,
+	cfg []byte,
+	in io.Reader,
+	out io.Writer,
+	errLog fsterr.LogInterface,
+	verbose bool,
+) error {
+	f.static = cfg
+	replacement := "Replace it with a valid version? (any existing email/token data will be lost) [y/N] "
```
fgsch, line 402: Not a blocker: should this be a const?

pkg/config/data.go
```diff
@@ -529,7 +436,8 @@ func (f *File) Read(path string, in io.Reader, out io.Writer, errLog fsterr.LogI
 	if unmarshalErr != nil {
 		errLog.Add(unmarshalErr)
 
-		// The embedded config is unexpectedly invalid.
+		// If the static embedded config failed to be unmarshalled, then that's
+		// unexpected and we can't recover from that.
 		if readErr != nil {
```
fgsch, line 441: Not a blocker: while I understand the logic here, it might be better to use something else to signal that the configuration we failed to unmarshal was the static one.

pkg/config/data.go
```diff
@@ -539,34 +447,34 @@ func (f *File) Read(path string, in io.Reader, out io.Writer, errLog fsterr.LogI
...
 			}
 			errLog.Add(err)
 			return err
 		}
+
+		data = f.static
+
+		Mutex.Lock()
+		err = toml.Unmarshal(data, f)
+		Mutex.Unlock()
```
fgsch, line 469: Do we still need the mutexes?

pkg/config/data.go
```diff
@@ -463,63 +390,43 @@ func (f *File) ValidConfig(verbose bool, out io.Writer) bool {
...
-			errLog.Add(readErr)
+func (f *File) Read(
+	path string,
+	cfg []byte,
+	in io.Reader,
+	out io.Writer,
+	errLog fsterr.LogInterface,
+	verbose bool,
+) error {
+	f.static = cfg
```
fgsch, line 401: Not a blocker: it feels to me this the static configuration is no longer something that should be known outside here, so maybe internalise it and drop the parameter from Read?

Integralist: Done.

pkg/config/data.go
Integralist: Done.

pkg/config/data.go
Integralist: Done.

pkg/config/data.go
Integralist: Removed, with the exception of the Write method (as documented in its comments).

pkg/config/config.go
```diff
@@ -344,193 +277,121 @@ type StarterKit struct {
...
 
-	err = createConfigDir(path)
-	if err != nil {
-		return err
-	}
+	// Store a copy of the static application configuration that has been embedded
+	// into the compiled CLI binary.
+	//
+	// NOTE: This field is private to prevent it being written back to disk.
+	static []byte `toml:",omitempty"`
```
fgsch, line 315: Not a blocker: we can remove this now.

dgryski:
:+1: :chainsaw:

PR #621, opened 2022-08-01
clean up more issues found by static analysis tooling

dgryski:


dgryski:
CI failed: `syscall.Stdin` is not an `int` on Windows.  Removing that commit.

dgryski:
`file.Join()` needed with a single argument on Windows due to path separators.  We could probably fix this by calling `filepath.Clean()` on the expected output, but dropping this commit for now.

PR #623, opened 2022-08-05
Fix `fastly help` tests to work with go 1.19

Integralist:
We noticed that go 1.19 would cause the `fastly help` tests to fail due to a difference in white space rendering. This wasn't happening with go 1.18.x but we became aware of it due to the Windows CI accidentally installing go 1.19 once it was released (which happened very recently).

We've bumped the CLI to use 1.19 and fixed the tests to work with 1.19.

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.18.x
+        go-version: 1.19.x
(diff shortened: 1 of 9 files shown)


fgsch (approved):
👍

Integralist:
With regards to "why", that was something I added because in the CLI output we were finding there were long lines trailing off the user's screen and we preferred these to be wrapped instead.

dgryski:
No, I meant why it didn't show up until Go 1.19.

PR #624, opened 2022-08-05
CI to use TinyGo action

Integralist:
Replace manual installation of TinyGo with a third-party GitHub Action (https://github.com/acifani/setup-tinygo).

**UPDATE**: Because the `setup-tinygo` repo author was unresponsive to an issue I raised that was blocking our use of their action, I was forced to fork it (https://github.com/integralist/setup-tinygo/) and publish my own GitHub Action that installs the necessary dependency.

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -66,6 +66,9 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go-version }}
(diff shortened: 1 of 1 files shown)


Integralist:
The error in the CI tests is happening on both macOS and Windows, so I've raised an issue in the action repo:
https://github.com/acifani/setup-tinygo/issues/2

Journie-Fastly (approved):
Looks good to me!

PR #626, opened 2022-08-10
Refactor

fgsch:
Some refactor after removing the dynamic config.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -43,30 +43,22 @@ const (
 
 	// FilePermissions is the default file permissions for the config file.
 	FilePermissions = 0o600
-
-	// RemoteEndpoint represents the API endpoint where we'll pull the dynamic
-	// configuration file from.
-	//
-	// NOTE: once the configuration is stored locally, it will allow for
-	// overriding this default endpoint.
-	RemoteEndpoint = "https://developer.fastly.com/api/internal/cli-config"
-
-	// UpdateSuccessful represents the message shown to a user when their
-	// application configuration has been updated successfully.
-	UpdateSuccessful = "Successfully updated platform compatibility and versioning information."
 )
 
-// ErrLegacyConfig indicates that the local configuration file is using the
-// legacy format.
-var ErrLegacyConfig = errors.New("the configuration file is in the legacy format")
+var (
 
-// ErrInvalidConfig indicates that the configuration file used was invalid.
-var ErrInvalidConfig = errors.New("the configuration file is invalid")
+	// ErrLegacyConfig indicates that the local configuration file is using the
+	// legacy format.
+	ErrLegacyConfig = errors.New("the configuration file is in the legacy format")
 
-// RemediationManualFix indicates that the configuration file used was invalid
-// and that the user rejected the use of the static config embedded into the
-// compiled CLI binary and so the user must resolve their invalid config.
-var RemediationManualFix = "You'll need to manually fix any invalid configuration syntax."
+	// ErrInvalidConfig indicates that the configuration file used was invalid.
+	ErrInvalidConfig = errors.New("the configuration file is invalid")
+
+	// RemediationManualFix indicates that the configuration file used was invalid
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -543,7 +509,9 @@ func (f *File) Write(path string) (err error) {
 	if err != nil {
 		return fmt.Errorf("error creating config file: %w", err)
 	}
-	if err := toml.NewEncoder(fp).Encode(f); err != nil {
+	encoder := toml.NewEncoder(fp)
+	encoder.Indentation("")
```
Integralist, line 513: What problem is `.Indentation` solving? 

I'd like us to add a `// NOTE: ...` to clarify why we use it.

Integralist: UPDATE: I think I see why it's used in the `config_test.go` below.

Integralist (commented):
LGTM. Couple of small comments.

pkg/config/config.go
fgsch: Sure, I will add a comment.

fgsch: Done.

pkg/config/config.go
```diff
@@ -484,20 +427,35 @@ func (f *File) MigrateLegacy() {
...
+
+		if verbose {
+			text.Output(out, `
+				Found your local configuration file (required to use the CLI) was using a legacy format.
+				File will be updated to the latest format.
+			`)
+			text.Break(out)
+		}
+		return true
+	case f.ConfigVersion != currentConfigVersion:
```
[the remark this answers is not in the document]
fgsch: That's fair but we currently don't have such a test. 

Do you want to cross that bridge when we get there or you would like to see a test for it?

pkg/config/config.go
fgsch: To clarify the original question, this is removing the leading spaces. 
I'm not sure it qualifies as a problem here but I don't think it's needed either.

pkg/config/config.go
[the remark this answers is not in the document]
fgsch: Test added.

pkg/config/config.go
```diff
@@ -347,7 +340,16 @@ func (f *File) Read(
...
 	verbose bool,
 ) error {
-	var useStatic bool
+	// Ensure the static config is sound. This should never happen (tm).
+	// We are checking this earlier to simplify the code later on.
+	var staticConfig File
+	err := toml.Unmarshal(Static, &staticConfig)
+	if err != nil {
+		errLog.Add(err)
+		return invalidConfigErr(err)
```
Integralist, line 349: The error message `ErrInvalidConfig` is a bit vague for a user who doesn't know anything about the CLI's internal config or the static embed etc, we should update the error and make it clearer to the user what we're referring to and in this scenario it's actually quite unexpected.

pkg/config/config.go
```diff
@@ -43,30 +43,23 @@ const (
...
-	// UpdateSuccessful represents the message shown to a user when their
-	// application configuration has been updated successfully.
-	UpdateSuccessful = "Successfully updated platform compatibility and versioning information."
 )
 
-// ErrLegacyConfig indicates that the local configuration file is using the
-// legacy format.
-var ErrLegacyConfig = errors.New("the configuration file is in the legacy format")
+var (
+	CurrentConfigVersion int
```
Integralist, line 49: Let's get a doc string comment on this now it's public.

Integralist (approved):
Couple of minor comments, but otherwise approved.

pkg/config/config.go
fgsch: Sure. Do you have any suggestions?

pkg/config/config.go
fgsch: Added.

pkg/config/config.go
fgsch: Updated.

Integralist (approved):
LGTM

PR #629, opened 2022-08-16
Change default Makefile target

### LINKED ISSUE #605: Error building from source
**Version**
Built with 4fdd620

**What happened**

When running `make` I get the following error

```
   • building...
   • loading config file       file=.goreleaser.yml
   • loading environment variables
   • getting and validating git state
      • building...               commit=4fdd620de7f8ae6a749bff970902702bb7c65a0a latest tag=v3.1.1
   ⨯ build failed after 0.21s error=git tag v3.1.1 was not made against commit 4fdd620de7f8ae6a749bff970902702bb7c65a0a
make: *** [Makefile:30: fastly] Error 1
```

Integralist:
**Problem**: Users expect `make` to compile a project successfully, but the current default target `fastly` would cause an error because it was setup to build with [goreleaser](https://goreleaser.com), which will report an error if the current commit of the `main` branch doesn't align with the latest git tag.

**Solution**: Set `build` to be the default target as it simply executes `go build` and will produce an executable binary for the user's OS/Arch.

Fixes #605

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+.PHONY: default
+default: build ;
+
 SHELL := /bin/bash -o pipefail
 
 # the rationale for using both `git describe` and `git rev-parse` is because


pnorman (approved):

chrispoole643 (approved):
LGTM

PR #632, opened 2022-09-02
Refactor

Integralist:
Fixing issues after running `revive` across the code base.

### DIFF
--- a/pkg/app/usage.go
+++ b/pkg/app/usage.go
@@ -215,12 +215,12 @@ func processCommandInput(
 	// existence of `help --format json`, if present we print usage JSON and
 	// exit early.
 	if cmd.ArgsIsHelpJSON(opts.Args) {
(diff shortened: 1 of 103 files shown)


pkg/app/usage.go
```diff
@@ -215,12 +215,12 @@ func processCommandInput(
 	// existence of `help --format json`, if present we print usage JSON and
 	// exit early.
 	if cmd.ArgsIsHelpJSON(opts.Args) {
-		json, err := UsageJSON(app)
+		j, err := UsageJSON(app)
```
Integralist, line 218: NOTE: These types of changes at first glance look not very useful but they're actually avoiding shadowing errors (e.g. `json` matches the standard library's json package name and so there's an overlap which could cause errors). 

So if you see a variable being renamed it's like to avoid package import name shadowing.

triblondon (approved):

Issue #603, opened 2022-07-22
Infinite loop when having `fastly compute build` defined in npm run build command


JakeChampion:
**Version**

Fastly CLI version v3.1.1 (59efd38)
Built with go version go1.18.3 linux/amd64
Viceroy version: viceroy 0.2.14

**What happened**

I created a new empty js compute project:
`fastly compute init --language javascript --from https://github.com/fastly/compute-starter-kit-javascript-empty`

I then changed the npm build run-script to "fastly compute build":
`npm set-script build "fastly compute build"`

I then ran `npm run build`:
`npm run build`

At this point it gets into an infinite loop.

The log gets this far and doesn't make it look like an infinite loop, which could confuse people:
```
> inf@0.3.0 prebuild
> webpack

asset index.js 219 bytes [emitted] [minimized] (name: main)
./src/index.js 212 bytes [built] [code generated]

WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value.
Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/

webpack 5.73.0 compiled with 1 warning in 110 ms

> inf@0.3.0 build
> fastly compute build

✓ Initializing...
✓ Verifying package manifest...
✓ Verifying local javascript toolchain...
/ Building package using javascript toolchain...
```

I think we should possibly add a flag to `fastly compute build` which enables it to *not* run `npm run build` within itself, something like `fastly compute build --ignore-npm-build` - that way javascript projects can keep the convention of having their build script be fully defined by `npm run build` and not get into this infinite loop

Integralist:
I propose the CLI no longer runs the `package.json` file's `build` step, basically in the CLI instead of calling `npm run build` we'll call `js-compute-runtime bin/index.js bin/main.wasm`. 

We don't do it for the `deploy` step (e.g. in all the starter kits they are set to `npm run build && fastly compute deploy` and the CLI's internal deploy logic doesn't rely on any package.json configuration).

I'll open a PR to fix this. Then I'll open a follow-up PR to update each JS starter kit so that its build step is just `fastly compute build`, then that way user's can modify it however they see fit.

PR #640, opened 2022-09-14
Move the internal build scripts to the fastly.toml manifest

Integralist:
Fixes https://github.com/fastly/cli/issues/603

This PR can be merged when approved, but the next CLI release will need to be closely coordinated with corresponding changes to the [Fastly Starter Kits](https://developer.fastly.com/solutions/starters/) and to the [Fastly Developer Hub documentation on the fastly.toml manifest](https://developer.fastly.com/reference/compute/fastly-toml/).

Due to the size of the refactoring and the fundamental change in how the build behaviour works, I'm going to treat this as a major version change when we come to do the next CLI release, so `v4.0.0`.

The majority of changes are refactoring the existing logic to use a new 'toolchain validation' abstraction layer. So for anyone reviewing the PR I would recommend looking at the `ToolchainValidator` struct (and its associated methods) in `pkg/commands/compute/language_toolchain.go`.


.fastly/config.toml
```diff
@@ -13,11 +13,8 @@ ttl = "5m"
     toolchain_constraint = ">= 1.17 < 1.19" # upper limit should be temporary (just while we figure out any rough edges)
 
   [language.rust]
-  toolchain_version = "1.56.1"
   toolchain_constraint = ">= 1.56.1"
   wasm_wasi_target = "wasm32-wasi"
-  fastly_sys_constraint = ">= 0.3.3"
-  rustup_constraint = ">= 1.23.0"
```
Integralist, line 20: The validation steps have been simplified and we no longer need to check the internal fastly-sys crate version, nor is the deprecated `toolchain_version` needed to be kept. We also don't need to check for all three binaries (rustc, cargo _and_ rustup) as the official recommended installation is to use rustup and that comes with cargo (which we check for as the primary toolchain tool).

.github/workflows/pr_test.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     strategy:
       matrix:
         tinygo-version: [0.24.0]
-        go-version: [1.19.x]
+        go-version: [1.18.x]
```
Integralist, line 63: Previously I bumped this to use the latest go1.19 but realised that this was causing issues running the tests as there is a constraint on using go < 1.19.

pkg/commands/compute/build.go
```diff
@@ -23,21 +23,10 @@ import (
 // IgnoreFilePath is the filepath name of the Fastly ignore file.
 const IgnoreFilePath = ".fastlyignore"
 
-// CustomBuildScriptMessage is the message displayed to a user when there is a
-// custom build script.
-const CustomBuildScriptMessage = "This project has a custom build script defined in the fastly.toml manifest"
```
Integralist, line 28: The `scripts.build` field is no longer 'custom' it's mandatory.

pkg/commands/compute/build.go
```diff
@@ -23,21 +23,10 @@ import (
...
 // CustomPostBuildScriptMessage is the message displayed to a user when there is a
 // custom post build script.
 const CustomPostBuildScriptMessage = "This project has a custom post build script defined in the fastly.toml manifest"
 
-// Toolchain abstracts a Compute@Edge source language toolchain.
-type Toolchain interface {
-	Initialize(out io.Writer) error
-	Verify(out io.Writer) error
-	Build(out io.Writer, progress text.Progress, verbose bool, callback func() error) error
-}
```
Integralist, line 39: This isn't deleted, it has just been moved to another file in the same package.

pkg/commands/compute/build_test.go
```diff
@@ -16,20 +16,30 @@ import (
...
+// buildScriptAssemblyScript is interpolated with each AssemblyScript test's
+// inlined fastly.toml
+const buildScriptAssemblyScript = "$(npm bin)/asc assembly/index.ts --outFile bin/main.wasm --optimize --noAssert"
+
+// buildScriptJavaScript is interpolated with each JavaScript test's inlined
+// fastly.toml
+const buildScriptJavaScript = "$(npm bin)/webpack && $(npm bin)/js-compute-runtime ./bin/index.js ./bin/main.wasm"
+
+// buildScriptGo is interpolated with each Go test's inlined fastly.toml
+const buildScriptGo = "tinygo build -target=wasi -wasm-abi=generic -gc=conservative -o bin/main.wasm ./"
```
Integralist, line 31: Since moving the build logic into the fastly.toml we needed to update a bunch of tests to have properly defined build steps.

pkg/commands/compute/compute_mocks_test.go
```diff
@@ -85,32 +80,3 @@ func getServiceDetailsWasm(i *fastly.GetServiceInput) (*fastly.ServiceDetail, er
 		Type: "wasm",
 	}, nil
 }
-
-type versionClient struct {
```
Integralist, line 89: As part of the consistency improvements and reduction in validation complexity, these test types are no longer used.

pkg/commands/compute/language_toolchain.go
```diff
@@ -1,10 +1,534 @@
...
+	"github.com/Masterminds/semver/v3"
+	fsterr "github.com/fastly/cli/pkg/errors"
+	fstexec "github.com/fastly/cli/pkg/exec"
+	"github.com/fastly/cli/pkg/filesystem"
+	"github.com/fastly/cli/pkg/manifest"
+	"github.com/fastly/cli/pkg/text"
+	toml "github.com/pelletier/go-toml"
 )
 
+// SDKErrMessageFormat is a format string that can be used by the
```
Integralist, line 23: Here's where the majority of the new functionality (i.e the new abstraction layer) exists.

pkg/commands/compute/language_toolchain.go
```diff
@@ -18,12 +542,77 @@ func getJsToolchainBinPath(bin string) (string, error) {
...
+	buildScript string
+	buildFn     func(string) (string, []string)
+	errlog      fsterr.LogInterface
+	pkgName     string
+	postBuild   string
+	timeout     int
+}
+
+// build compiles the user's source code into a Wasm binary.
+func build(
```
Integralist, line 579: Previously this logic was duplicated _four times_ across the code base. Now the logic is shared across all supported languages.

pkg/commands/compute/language_toolchain.go
```diff
@@ -18,12 +542,77 @@ func getJsToolchainBinPath(bin string) (string, error) {
...
-func checkJsPackageDependencyExists(bin, name string) bool {
-	// gosec flagged this:
-	// G204 (CWE-78): Subprocess launched with variable
-	// Disabling as the variables come from trusted sources:
-	// The CLI parser enforces supported values via EnumVar.
-	/* #nosec */
-	err := exec.Command(bin, "list", "--json", "--depth", "0", name).Run()
-	return err == nil
+// execCommand opens a sub shell to execute the language build script.
+func execCommand(cmd string, args []string, out, progress io.Writer, verbose bool, timeout int, errlog fsterr.LogInterface) error {
```
Integralist, line 546: Previously this logic was duplicated four times across the code base. Now the logic is shared across all supported languages.

Integralist (commented):
Some comments to help the PR reviewer.

pkg/commands/compute/build_test.go
```diff
@@ -394,16 +289,24 @@ func TestBuildAssemblyScript(t *testing.T) {
...
+			language = "assemblyscript"
+
+      [scripts]
+      build = "%s"`, buildScriptAssemblyScript),
 			wantOutputContains: "Built package 'test'",
+			skipWindows:        true,
 		},
 	} {
 		t.Run(testcase.name, func(t *testing.T) {
+			if fstruntime.Windows && testcase.skipWindows {
```
Integralist, line 306: There are two tests I've added a "Skip Windows" logic block to, and that's because we use `$()` command substitution to lookup the user's node_modules directory so we can safely execute the npm compiler binaries necessary. The `$()` syntax doesn't exist for `cmd.exe` on Windows. To support Windows would require the tests to be rewritten to use a proprietary `FOR` loop syntax which isn't worth implementing here.

**Side note:** The use of `$()` in the `script.build` of each starter kit is likely going to be an issue for these specific Window users and we'll need to document this somewhere either in DevHub or in each starter kit. That said I believe most Window developers nowadays use Powershell, Cygwin, WSL or some other posix compatible shell and so it shouldn't be a big issue).

pkg/commands/compute/language_assemblyscript.go
```diff
@@ -3,126 +3,105 @@ package compute
...
-	"github.com/fastly/cli/pkg/filesystem"
 	"github.com/fastly/cli/pkg/manifest"
 	"github.com/fastly/cli/pkg/text"
 )
 
-// ASSourceDirectory represents the source code directory.
-const ASSourceDirectory = "assembly"
+// AsCompilation is a language specific compilation target that converts the
+// language code into a Wasm binary.
+const AsCompilation = "asc"
```
kailan, line 15: Is this still needed? I would have thought that the CLI no longer needs to worry about the tool being used for compilation.

pkg/commands/compute/language_rust.go
```diff
@@ -4,710 +4,314 @@ import (
...
-// CargoPackage models the package configuration properties of a Rust Cargo
-// package which we are interested in and is embedded within CargoManifest and
-// CargoLock.
-type CargoPackage struct {
-	Name    string `toml:"name" json:"name"`
-	Version string `toml:"version" json:"version"`
-}
+// RustCompilationCommandRemediation is the command to execute to fix the
+// missing compilation target.
+const RustCompilationCommandRemediation = "rustup target add %s --toolchain $ACTIVE_TOOLCHAIN"
```
kailan, line 32: Again, not sure it's necessary to have all of this language-specific validation logic if we are just expecting the user to set the build script.

kailan (dismissed):
Awesome to see so much code removed!

The code changes look good and work as expected, however I think the fact that `fastly compute build` suddenly stops working with this update is very jarring and will be difficult for our users even with updated documentation.

Screenshot 2022-09-16 at 17 59 34

I feel that we should set a default build script based on the configured language if there is not one there already.

Integralist:
@kailan 

re: 

> I feel that we should set a default build script based on the configured language if there is not one there already.

Agreed. That's a good shout and will definitely smooth out the developer experience when upgrading. So I'll get that implemented 👍🏻

Integralist:
@kailan 

re:

> Is this still needed? I would have thought that the CLI no longer needs to worry about the tool being used for compilation.

and

> Again, not sure it's necessary to have all of this language-specific validation logic if we are just expecting the user to set the build script.

So, what was decided between @triblondon and myself was that the CLI would move the build command to fastly.toml but there would still fundamentally need to be some form of validation logic in the CLI to help guide developers who might not know what 'tools' need to be installed for the starter kit to work.

Now originally I had the idea that the validation logic in the CLI would be very 'generic' and that the fastly.toml would contain fields for indicating to the CLI what 'compiler' it should check for, and what 'SDK' the CLI should check was installed, etc. 

The downside of this approach was that it made the fastly.toml very complex, and for the most part the customer didn't need to have that level of detail exposed to them (as most customers won't end up messing around with the build command).

As a compromise we landed on having the validation logic stay within the CLI and be a 'known' set of tools (toolchains/compilers/sdks etc). If the user decides to change the fastly.toml build command then it will be documented (I think I'll also have a concise/single-line comment added above the build field in each starter kit too) that they need to use the `--skip-validation` flag. Because if they don't use that flag, then once they modify the build command, the CLI will continue to do validation and expect certain toolchains/compilers/sdks etc etc to be used, and depending on what they change the build command to, those internal validation steps could break because one of the tools the CLI expects to be installed is no longer necessary because of the change in build command.

An example might be that a customer decides that for a go starter kit they want to change the build command to no longer use tinygo and to instead just use Go proper (which does have Wasm support albeit very basic -- in some future time period where its basic support matures let's say).

Integralist:
@kailan just for visibility I've attached a screenshot of the latest behaviour where the CLI now injects a 'default' build script for the chosen language to help users with an existing project migrate more seamlessly to the CLI v4.0.0.

Screenshot 2022-09-20 at 13 50 15

See the line `Found default [scripts.build]`. If the user had started a fresh project with the updated starter kits that have the [scripts.build] field already in-place, then it would instead read `Found [scripts.build]` (no 'default' indicator).

pkg/commands/compute/language_toolchain.go
```diff
@@ -1,10 +1,562 @@
...
+
+	err = tv.FastlyManifestFile.Load(data)
+	if err != nil {
+		return fsterr.RemediationError{
+			Inner:       err,
+			Remediation: fmt.Sprintf(fsterr.ComputeBuildRemediation, tv.DefaultBuildCommand),
+		}
+	}
+
+	fmt.Fprintf(tv.Output, "Found default [scripts.build] '%s'\n", tv.DefaultBuildCommand)
```
triblondon, line 515: > No build command found.  Patching fastly.toml with the default build command for {language}: {command}

Maybe output this even when not in verbose mode.

triblondon: Also instead of "Building package uisng {language} toolchain" consider:

```
Running [scripts.build]:
> {command}
```

Integralist: @triblondon I've modified the above line 515 as per your suggestion but as for your second comment I've only changed the output to `Running [scripts.build]` and not to include the command output because when not using the `--verbose` flag we need the title to be one line. As demonstrated below...

Screenshot 2022-09-21 at 12 42 58

Integralist:
@triblondon I've addressed all your comments now.

Integralist:
@triblondon your comments (from an internal DM) have been addressed in this PR now.

**NOTES**:

- **_Strip the name and description fields from the fastly.toml on init._**  
  We still prompt for the 'name' and 'description' but the latter will be overwritten in the fastly.toml with an empty string if no value is provided at the prompt, while 'name' will default to the user's current directory before being written back to the fastly.toml manifest.

- **_Don’t populate any name or description fields in language-specific manifests._**  
  Done. But a side effect of this for the Rust language is that our build script will use `--bin compute-starter-kit-rust` and so each Rust starter kit will need their `Cargo.toml` 'package name' changed to that same value. Currently, each starter kit has a specific suffix depending on the starter kit, e.g. [`compute-starter-kit-rust-{type}`](https://github.com/fastly/compute-starter-kit-rust-default/blob/main/Cargo.toml#L2).

- **_Drop support for `--name` flag everywhere._**  
  Done.

- **_Drop support for `{name}` in build commands._**  
  Done (see above comment regarding Rust starter kits).

- **_On a service-creating deploy, prompt interactively for service name._**  
  Done.

- **_Add `--service-name` flag to the deploy command to allow the above to be done non-interactively._**  
  The flag already exists and can't be overloaded to do two things. We do provide a default when prompting a user for a service name and that default is the 'package' name we prompt users for as part of `fastly compute init` flow (which is persisted to the fastly.toml manifest).

triblondon (approved):

Issue #639, opened 2022-09-08
[FEATURE REQUEST] Allow configuration of what files and folder are watched via the `watch` flag


JakeChampion:
**Is your feature request related to a problem? Please describe.**
I'm not able to use the `--watch` flag as my folder structure is different from the assumption made in the fastly cli that the folder to watch is named `src` and causes the cli to error:

```bash
❯ fastly compute serve -i --watch
✓ Initializing...
✓ Verifying package manifest...
✓ Building package using custom toolchain...
✓ Creating package archive...

SUCCESS: Built package 'dynamic-backend' (pkg/dynamic-backend.tar.gz)

✓ Running local server...

INFO: The --watch flag defaults to watching file modifications in a ./src
directory.

2022/09/08 13:42:05 error walking directory tree 'src': lstat src: no such file or directory
```

**Describe the solution you'd like**
I would like to be able to supply a list of target files and folder (or perhaps a GLOB?) that I want the fastly cli to watch.

Perhaps something like `fastly compute serve --watch folder1 --watch file.js --watch folder2

**Describe alternatives you've considered**
I can avoid using the watch feature of the fastly cli and use another program to handle watching the folders/files and rerunning the fastly cli command

**Additional context**
output from `fastly version`:
```
Fastly CLI version v3.3.0 (96b3aa7)
Built with go version go1.19 linux/amd64
Viceroy version: viceroy 0.2.15
```

Integralist:
This would be a breaking change, requiring a major version update (which is fine if changing `--watch` is the best solution) but an alternative might be to add a specific flag like `--watch-pattern` and have it accept a regex pattern.

JakeChampion:
> This would be a breaking change, requiring a major version update (which is fine if changing `--watch` is the best solution) but an alternative might be to add a specific flag like `--watch-pattern` and have it accept a regex pattern.

A way to make this non-breaking could be to have a default value for `--watch` of `src`, I'm not sure how difficult that would be to implement though

Integralist:
> A way to make this non-breaking could be to have a default value for `--watch` of src.

That's not actually a big deal, we just need to change the flag type from a 'boolean' type to a 'string' type and call `.Default("src")`, so 👍🏻

Integralist:
[...]

PR #642, opened 2022-09-16
Watch all project files

Integralist:
Fixes https://github.com/fastly/cli/issues/639

When setting the `--watch` flag on the `compute serve` command, the current implementation attempts to identify the language so it can watch files within the language-specific directory.

This PR changes that behaviour so that all files within the project root directory is watched. This is a more flexible solution, and also means that customers with a directory structure that doesnt' match our language defaults will still work with the `--watch` flag (see linked issue).


chrispoole643 (approved):
LGTM

PR #649, opened 2022-09-28
Implement `compute hashsum`

Integralist:
**NOTE**: It has been agreed that we should remove the `pkg/app/run_test.go` `TestApplication` as it's not a useful test and is more noise than signal


fgsch:
Looks good to me. 

I'd expect to see the build output when using -v, however. What do you think?

Integralist:
Added support for showing build output with --verbose flag...

Screenshot 2022-09-29 at 11 59 29

fgsch:
🚢 🇮🇹

PR #653, opened 2022-10-04
Reduce duplication of string in logging package

grantstephens:
https://fastly.atlassian.net/browse/LOG-1354


pkg/commands/logging/common/flags.go
```diff
@@ -0,0 +1,81 @@
+package common
+
+import (
+	"github.com/fastly/cli/pkg/cmd"
+	"github.com/fastly/kingpin"
+)
+
+// Add format flag to the CmdClause
```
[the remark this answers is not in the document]
Integralist: I would recommend making the same change for all functions below.

pkg/commands/logging/datadog/create.go
```diff
@@ -63,9 +64,9 @@ func NewCreateCommand(parent cmd.Registerer, globals *config.Data, data manifest
 	})
 	c.CmdClause.Flag("region", "The region that log data will be sent to. One of US, US3, US5, or EU. Defaults to US if undefined").Action(c.Region.Set).StringVar(&c.Region.Value)
 	c.CmdClause.Flag("format", "Apache style log formatting. For details on the default value refer to the documentation (https://developer.fastly.com/reference/api/logging/datadog/)").Action(c.Format.Set).StringVar(&c.Format.Value)
```
Integralist, line 66: I would say for flags like this where the description is very specific, we either...

A.) modify the function to accept a string that can be 'appended' if provided.

or

B.) Just remove the difference (e.g. remove "For details on the default value...").

The latter might be better as some flag differences might not be as simple as simply _appending_ a value to the flag description. Also, it's not that useful providing such a long link in the description (IMHO).

Integralist (commented):
Thanks @grantstephens for opening this PR.

I have a couple of comments for you.

pkg/commands/logging/common/flags.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/fastly/cli/pkg/cmd"
+	"github.com/fastly/kingpin"
+)
+
+// Format defines the format flag
+func Format(cmd *kingpin.CmdClause, c *cmd.OptionalString) {
+	cmd.Flag("format", "Apache style log formatting. Your log must produce valid JSON").Action(c.Set).StringVar(&c.Value)
+}
+
+// Gziplevel defines the gzip flag
```
Integralist, line 13:
```suggestion
// GzipLevel defines the gzip flag
```

PR #657, opened 2022-10-05
Fix Rust 'existing project' bug

Integralist:
The v4 release didn't take into account a user who upgrades to the latest CLI version but has a pre-existing project.

The pre-existing project will have in their Cargo.toml a `[package.name]` set to something like `compute-starter-kit-rust-default`, which doesn't match the expected package name that the build script is looking for: `--bin fastly-compute-project`.

### DIFF
--- a/pkg/commands/compute/build_test.go
+++ b/pkg/commands/compute/build_test.go
@@ -125,7 +125,7 @@ func TestBuildRust(t *testing.T) {
 			language = "rust"
 
       [scripts]
-      build = "%s"`, compute.RustDefaultBuildCommand),
+      build = "%s"`, fmt.Sprintf(compute.RustDefaultBuildCommand, compute.RustPackageName)),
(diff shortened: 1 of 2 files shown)


triblondon (commented):
OK so AIUI, this problem arises when the user installs the updated CLI, and is in this (common) situation:

1. Has no `[scripts.build]` in their fastly.toml
2. Does have a package name in their cargo.toml which is not `fastly-compute-project`

In that case, right now we error because we populate a default build command into fastly.toml, which doesn't match their package name.

So we have two possible solutions:

1. Generate a customised build command to insert into the fastly.toml, referencing the user's chosen cargo.toml package name
2. Update the cargo.toml package name to match the default build command that we insert into fastly.toml

This PR proposes option 2, I think - but option 1 seems more obvious to me :-)

triblondon (approved):

PR #662, opened 2022-10-11
Fix `compute pack` to produce expected `package.tar.gz` file name

Integralist:
We updated the `compute build` and `compute deploy` commands to generate, and lookup, a package inside `/pkg/package.tar.gz` but we then neglected to update the `compute pack` command (which still had the old behaviour of _dynamically_ generating the package filename, e.g. `/pkg/<dynamic_name>.tar.gz`).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,9 +16,7 @@ require (
 	github.com/getsentry/sentry-go v0.12.0
 	github.com/google/go-cmp v0.5.6
 	github.com/google/go-github/v38 v38.1.0
-	github.com/kennygrant/sanitize v1.2.4
 	github.com/mattn/go-isatty v0.0.14
-	github.com/mholt/archiver v3.1.1+incompatible
 	github.com/mholt/archiver/v3 v3.5.0
 	github.com/mitchellh/go-wordwrap v1.0.1
 	github.com/mitchellh/mapstructure v1.4.3
@@ -33,6 +31,7 @@ require (
 )
 
 require (
+	github.com/mholt/archiver v3.1.1+incompatible
 	github.com/otiai10/copy v1.7.0
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/tcnksm/go-gitconfig v0.1.2
@@ -55,6 +54,5 @@ require (
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
-	golang.org/x/net v0.0.0-20211015210444-4f30a5c0130f // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -16,9 +16,7 @@ require (
 	github.com/getsentry/sentry-go v0.12.0
 	github.com/google/go-cmp v0.5.6
 	github.com/google/go-github/v38 v38.1.0
-	github.com/kennygrant/sanitize v1.2.4
```
Integralist, line 19: ℹ️ Now we no longer generate a package name from information in the user's `fastly.toml` manifest file, it means we don't need this dependency.

pkg/commands/compute/pack.go
```diff
@@ -39,6 +39,8 @@ func (c *PackCommand) Exec(_ io.Reader, out io.Writer) (err error) {
 	progress := text.NewProgress(out, c.Globals.Verbose())
 
 	defer func(errLog fsterr.LogInterface) {
+		os.RemoveAll("pkg/package")
```
Integralist, line 42: ℹ️ In the `compute build` command we create a 'temp' directory to store files that are placed into the final `package.tar.gz`, and then clean up the temp directory. So we are now doing the same thing here, and are just making sure we clean-up after ourselves.

pkg/commands/compute/pack.go
```diff
@@ -48,13 +50,12 @@ func (c *PackCommand) Exec(_ io.Reader, out io.Writer) (err error) {
 	if err = c.manifest.File.ReadError(); err != nil {
 		return err
 	}
-	name := sanitize.BaseName(c.manifest.File.Name)
-	pkg := fmt.Sprintf("pkg/%s/bin/main.wasm", name)
-	dir := filepath.Dir(pkg)
-	err = filesystem.MakeDirectoryIfNotExists(dir)
+	bin := "pkg/package/bin/main.wasm"
+	bindir := filepath.Dir(bin)
```
Integralist, line 54: ℹ️ I'm just renaming variables for greater clarity, as the previous names were a bit too ambiguous.

pkg/commands/compute/pack.go
```diff
@@ -91,7 +92,7 @@ func (c *PackCommand) Exec(_ io.Reader, out io.Writer) (err error) {
 
 	progress.Step("Copying manifest...")
 	src = manifest.Filename
-	dst = fmt.Sprintf("pkg/%s/%s", name, manifest.Filename)
+	dst = fmt.Sprintf("pkg/package/%s", manifest.Filename)
```
Integralist, line 95: ℹ️ Here is the crux of the PR, this is where we're fixing the destination path.

pkg/commands/compute/pack_test.go
```diff
@@ -36,41 +34,17 @@ func TestPack(t *testing.T) {
 				"Creating .tar.gz file...",
 			},
 			expectedFiles: [][]string{
-				{"pkg", "mypackagename", "bin", "main.wasm"},
```
Integralist, line 39: ℹ️ Now we're cleaning up after ourselves we need to fix the tests to not check for that directory/files any more.

Integralist (commented):
Here are some comments to help guide the reviewer...

PR #665, opened 2022-10-11
Ensure the `./bin` directory is present even with `--skip-verification`

Integralist:
**Problem**: Currently, using `--skip-verification` will remove a critical side-effect step (i.e. it will also skip creating a `./bin` directory necessary for compilers to place the `main.wasm` file into).

**Solution**: We move the code for creating the bin directory out of the validation code section, to _before_ we attempt validating the user's toolchain.


fgsch (approved):
LGTM. Thank you 👍

PR #667, opened 2022-10-11
Fix changing first byte timeout in backend update command

cs278:
Using the `--first-byte-timeout` option would always update the timeout to zero regardless of the value supplied.

Example:

```sh
$ fastly backend update --name test --version latest --first-byte-timeout=10000
...
$ fastly backend describe --name test --version latest

Service ID: ...
Service Version: 16

Name: ...
Comment: 
Address: ...
Port: 443
Override host: 
Connect timeout: 10000
Max connections: 200
First byte timeout: 0
Between bytes timeout: 5000
...
```


la5942 (approved):

Integralist (approved):
Ah, nice! Good spot. Thank you for the PR ❤️

PR #672, opened 2022-10-13
handle --help with no arguments. fixes #154

### LINKED ISSUE #154: fastly --help returns argument parsing error
**Version**

Fastly CLI version 0.16.1 (b12ad86)
Built with go version go1.14.6 linux/amd64

**What happened**

Ran the command `fastly --help`

got
```
ERROR: error parsing arguments: command not specified.
```

expected: this command should be complete and not continue with argument parsing.

shawnps:
Fixes #154

### DIFF
--- a/pkg/app/usage.go
+++ b/pkg/app/usage.go
@@ -246,6 +246,10 @@ func processCommandInput(
 	// But it's useful to have it implemented so it's ready to roll when we do.
 	var vars map[string]any
 
+	if cmd.IsHelpFlagOnly(opts.Args) && len(opts.Args) == 1 {
+		return command, cmdName, help(vars, nil)
+	}
+
 	// NOTE: We call two similar methods below: ParseContext() and Parse().
 	//
 	// We call Parse() because we want the high-level side effect of processing


Integralist (approved):
Thanks @shawnps for opening this PR.

I will approve because it fixes the issue reported, but it would be nice if the command also didn't return an exit code of 1 (in case you wanted to try and address that in this PR).

I have noticed that the other help commands, like `fastly help` and `fastly help <command>`, all return an exit code 1 (while `fastly --help` and `fastly <command> --help` all exit with zero).

shawnps:
Thanks @Integralist 

I think [this](https://github.com/fastly/cli/blob/main/cmd/fastly/main.go#L144-L153) is where the exit 1 happens, so maybe we would need to return `(error, int)` from the `app.Run` function (called [here](https://github.com/fastly/cli/blob/main/cmd/fastly/main.go#L132)) where `int` is the exit code.

benzvan:
Hey! Thanks for addressing this. It was a minor inconvenience but really good to get fixed.

Integralist:
@benzvan I'm just sorry it took so long to get around to it! Thanks for your patience 🙂

There is an additional improvement I'd like to see made (see [my earlier comment](https://github.com/fastly/cli/pull/672#pullrequestreview-1140261519)) and so hopefully it doesn't take as long to resolve that.

PR #676, opened 2022-10-17
Workaround Window's `cmd.exe`'s lack of support for command substitution

### LINKED ISSUE #675: Windows support needs more consideration for JavaScript projects
In order for the CLI (v4.0.0+) to build a C@E project, it needs to read a build script defined in the fastly.toml manifest. 

On Windows the shell used to execute the build script is hard-coded to [`cmd.exe`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd).

This introduces a problem because although `cmd.exe` is the default shell on Windows, and so we know it's always available, it doesn't recognise the 'command substitution' syntax `$()` which is necessary for some starter kits.

For example, the JavaScript starter kits need to look up the `node_modules` directory so it can utilise the installed binaries like `webpack` and `js-compute-runtime`...

```shell
$(npm bin)/webpack && $(npm bin)/js-compute-runtime ./bin/index.js ./bin/main.wasm
```

> **NOTE**: Earlier versions of the CLI, which had the build script hard-coded into it, didn't have an issue because the Go code would open a subshell to execute `npm build` and then it would concatenate the result with another string and pass the single string through to another subshell which would just execute the inlined command (i.e. no command substitution was being used).

I'm using an M1 Mac and so running a Windows VM is not possible currently and so I've had to set-up a development environment on a spare Windows laptop I had lying around at home. From there I've tried to validate some alternative approaches (and just Windows testing in general).

I initially wondered if the Fastly CLI should just be more explicit about its shell requirements (e.g. we require a shell that is posix compliant), but this meant validating the code against various Windows shells to see what _actually_ works. Once we know what works and what caveats we have, then we can decide if we want to switch the CLI's hard-coded Windows shell from `cmd.exe` to one of the alternatives.

There are a few different shells available for Windows (this is not an exhaustive list)...

1. [Command Prompt](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd) (Cmd.exe)
2. [Cygwin](https://www.cygwin.com)
3. [Git for Windows](https://gitforwindows.org) (e.g. provides a BASH emulation used to run Git from the command line)
4. [PowerShell](https://learn.microsoft.com/en-us/powershell/scripting/overview?view=powershell-7.2)
5. [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/install) (WSL)
6. [Windows Terminal](https://learn.microsoft.com/en-us/windows/terminal/)

## PowerShell

I tried to switch to using `powershell.exe` (which although not installed by default, is the next simplest shell to get installed for Windows) and I discovered it doesn't support the 'logical AND' operator `&&`. This means you have to use a more arcane `-and` syntax, but this would make the fastly.toml `[scripts.build]` incompatible with _other_ shells that are POSIX compatible. 

We _could_ have configured the CLI to pattern match `&&` and then, when on a Windows OS, replaced it with `-and` but I then discovered PowerShell fails to execute a binary using the syntax `$(npm bin)/webpack` (I would hit this issue again later when testing the new Windows Terminal). The following works in Bash `bin=$(npm bin); $bin/webpack` and avoids `&&` but in PowerShell you can't assign the result of a command substitution to a variable. So we can't workaround it that way either.

## Cmd.exe

Interestingly, `cmd.exe` _does_ support `&&` but as we know, it fails to support `$()`. There is an alternative syntax but I couldn't get it to work (even with a simple example), let alone something more complex like: 

```shell
for /F "tokens=*" %n IN ('npm bin') DO @(%n\webpack)
```

But this is a very confusing syntax. Also, I'm not sure how to 'chain' the command so it also executes the `js-compute-runtime` binary (as we do currently in the fastly.toml `[scripts.build]`). Having to logic hop around this sort of code would be a project maintenance/sustainability concern (if we were to go to these lengths to support building on Windows), it also means to ensure the fastly.toml stays compatible with the majority of users on actual POSIX/Bash shells we'd have to have the CLI parse the build script and identify any `$()` usage and replace it with the above more complex syntax (which would be a very brittle solution).

## Cygwin

I've not tested Cygwin here. I've used it in the past for testing but it is a specialised tool that provides an isolated environment to work within and so that could cause problems if there are expectations for the CLI to be able to access files outside of that environment. Also, from my understanding Cygwin is less commonly used nowadays and so it's unlikely we would want to require users to install it just to use the CLI. Ideally a tool that doesn't require an isolated environment would be better.

## Git for Windows

I installed Git for Windows but found it also created its own isolated environment, and I also couldn't figure out how to share files from my host Windows machine to it.

## Windows Subsystem for Linux (WSL)

Installing WSL requires Windows 10+, which might exclude some Window developers (looks like Windows 10 was released in 2015 so maybe users on an earlier version is not a concern). WSL also sets up an isolated environment but what is interesting is that it also (_partially_ at least) integrates with `cmd.exe` and PowerShell (e.g. you can pass commands to be executed to the isolated environment from the host environment). 

This integration is not a panacea, as although a simple execution like `wsl ls -a` (which prints all files from my host Windows machine) works fine, when I try to execute something like `wsl ls $(npm bin)` it fails to access the directory. This means a more complex mounting of files into the WSL is necessary and again highlights the 'isolated environment' concern I had with Cygwin and Git for Windows.

## Windows Terminal

When installing the new Windows Terminal, it appears to be an enhanced PowerShell (it actually shows as 'PowerShell' in the terminal), as it supports `&&` as well as other posix/Bash flavoured syntax. Although executing  `ls $(npm bin)` worked (unlike WSL), the moment I tried `$(npm bin)/webpack` to execute the binary within the directory, the command execution failed because of some unsupported syntax (i.e. the forward slash was not recognised). Meaning if we have to modify the script in the `fastly.toml` then it isn't going to be compatible with other shells. This error also occurred with the standard PowerShell.

# Conclusion

As I mentioned at the start of this issue. The Windows problem with `$()` didn't exist prior to v4.0.0 because the CLI was opening a separate subshell to get the `npm bin` results and concatenating it with the binary that needed to be executed, and passing a single string to a new subshell to execute.

With v4.0.0+ I'm not sure what the best approach is. This currently looks to only affect the JavaScript starter kits as they are the only language that needs to resolve installed binaries to the `node_modules` directory using command substitution. We also can't have the CLI assume the user's `node_modules` directory is in the same directory as where the CLI is running (as we know there are customers who work within a mono repo and have nested project structures).

We could have the CLI try and flatten out the user's defined `[scripts.build]`. This would mean parsing the script for any use of `$(npm bin)` and have the CLI do that lookup via a subshell. Then we replace `$(npm bin)` in the user's fastly.toml with the fully qualified path that `npm bin` would have returned. We could even limit this workaround to just Window users.

Integralist:
Fixes #675 

Refer to the linked issue for context.

## Notes

This PR could introduce behaviour that is unexpected in two ways:

1. We're replacing shell command substitution at runtime.
2. We're only doing that for instances of `$(npm bin)`.

If the fastly.toml `[scripts.build]` uses command substitution (e.g. `$(npm bin)`) and the CLI internally evaluates that on behalf of the user (i.e. it doesn't actually allow the command substitution to happen, but makes a separate subshell call instead to work around the fact that command substitution isn't supported well in Windows), then that could confuse a seasoned Windows developer who might look at the `[scripts.build]` and wonder how it is working (if they happen to know that command substitution isn't supported by the default Command Prompt, `cmd.exe`, on Windows).

That said, the user doesn't actually know we're using `cmd.exe` (on Windows) unless they look at the CLI source code, as we don't display the shell used during a `compute build`. 

To avoid confusion I've added a note that will be printed for Windows users if a command substitution for `npm bin` has been evaluated by the CLI.

We might want to expand this behaviour to command substitution in general. But I'll await an initial review to gauge feedback.

### DIFF
--- a/pkg/commands/compute/language.go
+++ b/pkg/commands/compute/language.go
@@ -1,8 +1,10 @@
 package compute
 
 import (
+	"bytes"
 	"fmt"
 	"io"
+	"os/exec"
 	"runtime"
 	"sort"
 	"strings"
@@ -151,9 +153,36 @@ func (s Shell) Build(command string) (cmd string, args []string) {
 	if runtime.GOOS == "windows" {
 		cmd = "cmd.exe"
 		args = []string{"/C"}
+		command = flattenNPMBinSubshell(command)
 	}
 
 	args = append(args, command)
 
 	return cmd, args
+}
+
+// flattenNPMBinSubshell parses the command for the string $(npm bin) or the
+// older backtick command substitution `npm bin`, and executes a subshell to
+// acquire the value (rather than letting the user's shell identify the value).
+//
+// NOTE: This function should only be called when running on Windows.
+// The default Windows 'Command Prompt' (cmd.exe) can't parse the POSIX command
+// substitution syntax. Refer to https://github.com/fastly/cli/issues/675 for
+// the full details.
+func flattenNPMBinSubshell(command string) string {
+	var buf bytes.Buffer
+
+	cmd := "cmd.exe"
+	args := []string{"/C", "npm bin"}
+
+	c := exec.Command(cmd, args...)
+	c.Stdout = &buf
+
+	if err := c.Run(); err == nil {
+		result := strings.TrimSpace(buf.String())
+		command = strings.ReplaceAll(command, "$(npm bin)", result)
+		command = strings.ReplaceAll(command, "`npm bin`", result)
+	}
+
+	return command
 }
--- a/pkg/commands/compute/language.go
+++ b/pkg/commands/compute/language.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/fastly/cli/pkg/config"
 	"github.com/fastly/cli/pkg/manifest"
+	"github.com/fastly/cli/pkg/text"
 )
 
 // NewLanguages returns a list of supported programming languages.
@@ -146,14 +147,14 @@ type Shell struct{}
 //
 // Should be converted into a command such as (on unix):
 // sh -c "yarn install && yarn build"
-func (s Shell) Build(command string) (cmd string, args []string) {
+func (s Shell) Build(command string, out io.Writer) (cmd string, args []string) {
 	cmd = "sh"
 	args = []string{"-c"}
 
 	if runtime.GOOS == "windows" {
 		cmd = "cmd.exe"
 		args = []string{"/C"}
-		command = flattenNPMBinSubshell(command)
+		command = flattenNPMBinSubshell(command, out)
 	}
 
 	args = append(args, command)
@@ -169,7 +170,7 @@ func (s Shell) Build(command string) (cmd string, args []string) {
 // The default Windows 'Command Prompt' (cmd.exe) can't parse the POSIX command
 // substitution syntax. Refer to https://github.com/fastly/cli/issues/675 for
 // the full details.
-func flattenNPMBinSubshell(command string) string {
+func flattenNPMBinSubshell(command string, out io.Writer) string {
 	var buf bytes.Buffer
 
 	cmd := "cmd.exe"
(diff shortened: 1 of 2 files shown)


Integralist:
I've tested the use of `npm exec` on my Windows machine and that works fine. So I will close off this PR now.

PR #679, opened 2022-10-18
Fix Makefile for Windows

Integralist:


Makefile
```diff
@@ -15,13 +15,21 @@ GO_BIN ?= go
...
 
 GOHOSTOS ?= $(shell go env GOHOSTOS || echo unknown)
 GOHOSTARCH ?= $(shell go env GOHOSTARCH || echo unknown)
-GO_FILES = $(shell find cmd pkg -type f -name '*.go')
+
+ifeq ($(OS), Windows_NT)
+	SHELL = cmd.exe
+	.SHELLFLAGS = /c
+	GO_FILES = $(shell where /r pkg *.go)
+	GO_FILES += $(shell where /r cmd *.go)
```
Integralist, line 29: ℹ️ Recursively searching for files is actually quite tricky in PowerShell (the default CI Windows environment), and so I switched to using `cmd.exe` which is the default Terminal on Windows typically (PowerShell is an enhanced shell environment but it's something you have to manually install on Windows). Additionally, the CLI (because of this fact) uses `cmd.exe` when opening a subshell to execute build scripts on Windows, and so it's better for our CI to be consistent with that.

Makefile
```diff
@@ -15,13 +15,21 @@ GO_BIN ?= go
 TEST_COMMAND ?= $(GO_BIN) test
 
 # The compute tests can sometimes exceed the default 10m limit.
-TEST_ARGS ?= -timeout 15m ./{cmd,pkg}/...
+TEST_ARGS ?= -timeout 15m ./...
```
Integralist, line 18: ℹ️ Following my comment about switching to `cmd.exe` the shell expansion `{x, y}` doesn't work in `cmd.exe` and so we have to remove it. I'm not sure why we did this, maybe for performance reasons (e.g. narrow down the number of files the `go` toolchain has to lookup).

PR #683, opened 2022-10-18
Fix runtime panic

Integralist:
Running `fastly` with no arguments now triggers a runtime panic.

This wasn't visible until https://github.com/fastly/cli/pull/672 was landed. The logic bug was already there, but the recent merge meant the control flow was able to reach that portion of the code that was attempting to index an empty list.

### DIFF
--- a/.fastly/config.toml
+++ b/.fastly/config.toml
@@ -3,10 +3,6 @@ config_version = 2
 [fastly]
 api_endpoint = "https://api.fastly.com"
 
-[cli]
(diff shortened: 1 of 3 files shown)


fgsch (approved):

fgsch:
I feel the [remove redundant fields/objects](https://github.com/fastly/cli/pull/683/commits/5c2c343e5299ba00e08c7d7592de4d23d9bb6f3a) change should be done separately but either way looks good to me.

shawnps (approved):

PR #684, opened 2022-10-19
Fix help to not always exit 1

Integralist:


fgsch:
My first impression from a rather quick glance is that perhaps we could encode the exit code in a new error type.
That might simplify the code and reduce the number of changes.

Integralist:
Closing in favour of https://github.com/fastly/cli/pull/689.

PR #688, opened 2022-10-19
Fix release process

Integralist:


pkg/config/config.toml
```diff
@@ -0,0 +1,96 @@
...
+  [language.rust]
+  toolchain_constraint = ">= 1.56.1"
+  wasm_wasi_target = "wasm32-wasi"
+
+[viceroy]
+ttl = "24h"
+
+[starter-kits]
+
+[[starter-kits.assemblyscript]]
```
Integralist, line 20: ❓ @triblondon which starter kits would you like to have excluded from the list.

Integralist (commented):
Just some comments for the reviewer...

triblondon (commented):
When we chatted about this on zoom, the approach we discussed was to list SK repo URLs in a file in the CLI repo, and then pull the title and description of the SKs from the SK repo's fastly.toml file during the CLI build.  It seems like the solution here is to bake the title and description data into the CLI repo directly, which duplicates and disconnects it from the SK's canonical data.

My concern is that this creates a maintenance overhead, and an unexpected gotcha if you update the SK and then expect the output from the CLI to change.

These are the repos I think we should offer in the CLI, in this order:

* Rust:
    * https://github.com/fastly/compute-starter-kit-rust-default
    * https://github.com/fastly/compute-starter-kit-rust-empty
    * https://github.com/fastly/compute-starter-kit-rust-static-content
    * https://github.com/fastly/compute-starter-kit-rust-websockets
* JavaScript
    * https://github.com/fastly/compute-starter-kit-javascript-default
    * https://github.com/fastly/compute-starter-kit-javascript-empty
* AssemblyScript
    * https://github.com/fastly/compute-starter-kit-assemblyscript-default
* Go
    * https://github.com/fastly/compute-starter-kit-go-default

scripts/config.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
Integralist, line 1: @triblondon here's the script for generating the config.

DEVELOP.md
```diff
@@ -55,18 +55,12 @@ CLI_CATEGORY=logging CLI_CATEGORY_COMMAND=logging CLI_PACKAGE=foobar CLI_COMMAND
...
 
-```
-https://developer.fastly.com/api/internal/cli-config
-```
-
-The config served from that endpoint is the result of an internal Fastly build process that uses the `.fastly/config.toml` file in the CLI repo as a template, and then dynamically adds in any available Starter Kits.
+The file is added to `.gitignore` to avoid it being added to the git repository.
 
-The configuration is then saved to disk (at the following location) and embedded into the CLI when compiled.
+When compiling the CLI for a new release, it will pull execute [`./scripts/config.sh`](./scripts/config.sh). The script uses [`./.fastly/config.toml`](./.fastly/config.toml) as a template file to then dynamically inject a list of starter kits (pulling their data from their public repositories).
```
Integralist, line 64:
```suggestion
When compiling the CLI for a new release, it will execute [`./scripts/config.sh`](./scripts/config.sh). The script uses [`./.fastly/config.toml`](./.fastly/config.toml) as a template file to then dynamically inject a list of starter kits (pulling their data from their public repositories).
```

fgsch (approved):
LGTM, windows CI failure aside :)

PR #689, opened 2022-10-20
Skip exit code 1 for 'help' output

Integralist:


pkg/commands/compute/build_test.go
```diff
@@ -567,7 +567,7 @@ func TestBuildGo(t *testing.T) {
 			// NOTE: The following constraints should be kept in-sync with
 			// ./pkg/config/config.toml
 			opts.ConfigFile.Language.Go.TinyGoConstraint = ">= 0.24.0-0" // NOTE: -0 is to allow prereleases.
-			opts.ConfigFile.Language.Go.ToolchainConstraint = ">= 1.17 < 1.19"
+			opts.ConfigFile.Language.Go.ToolchainConstraint = ">= 1.17"
```
fgsch, line 570: Sneaked in?

Integralist: This was something that would cause the tests to fail when running them against go1.19 

It should have been updated/kept in-sync with the config.toml constraint that was updated a while ago, but I forgot about it because at the time when I ran the tests before I was using go1.18.5 so this test didn't fail at the time.

PR #694, opened 2022-10-26
Make a build quiet

Integralist:
The `fastly compute build` doesn't prompt for information, and although validation of the user's environment might fail, if the user says they want output to be `--quiet` we should respect that.


fgsch (approved):
👍

PR #695, opened 2022-10-26
Add missing --help flag to globals

Integralist:
`--help` is a Kingpin global flag but it was missing from the list of global flags to check for.


fgsch (approved):
🚢 🇮🇹

PR #698, opened 2022-10-27
Add PrintLines to text package and use in logging

grantstephens:
As per LOG-1355


pkg/commands/logging/azureblob/describe.go
```diff
@@ -103,27 +104,29 @@ func (c *DescribeCommand) Exec(_ io.Reader, out io.Writer) error {
 		}
 		return nil
 	}
-
+	lines := text.Lines{
```
Integralist, line 107: Sorry to be a pain, but can we ensure all the fields are also listed in the code in order (again, it makes it easier to find and add new fields in future). If you could do that across each file please.

Integralist (approved):
Approving, but have one requested change...

pkg/commands/logging/azureblob/describe.go
grantstephens: Done in https://github.com/fastly/cli/pull/698/commits/d79b71ec0b32c3121d76331fef143a8231e21671

PR #707, opened 2022-11-14
Bump go-fastly to v7

grantstephens:
This got a bit complicated as Kingpin doesn't like pointers, which is what go-fastly v7 introduced.

> **NOTE**: An important change within this PR is not presuming we know how the API works. The go-fastly API client has taken this approach by not trying to validate certain fields that it _thinks_ the API will need. The CLI will now take a similar approach where it will only provide the fields that go-fastly says are 'required' because the code base shouldn't introduce code logic that could become stale if the API ever changes its behaviours.


pkg/commands/dictionary/dictionary_test.go
```diff
@@ -324,12 +324,16 @@ func describeDictionaryOKDeleted(i *fastly.GetDictionaryInput) (*fastly.Dictiona
 }
 
 func createDictionaryOK(i *fastly.CreateDictionaryInput) (*fastly.Dictionary, error) {
+	// This is a bit nastly, but just an artifact of testing- nil is false.
```
Integralist, line 327: I think it's fine not to have the comment here.
```suggestion
```

pkg/commands/domain/create.go
```diff
@@ -8,17 +8,20 @@ import (
...
-	Input          fastly.CreateDomainInput
-	serviceName    cmd.OptionalServiceNameID
-	serviceVersion cmd.OptionalServiceVersion
-	autoClone      cmd.OptionalAutoClone
+	manifest manifest.Data
+	// required
+	name           string
+	ServiceName    cmd.OptionalServiceNameID
+	ServiceVersion cmd.OptionalServiceVersion
+	// optional
```
Integralist, line 22: nit: I noticed we're a bit inconsistent in spacing in between the required and optional flags, for example elsewhere I've see you include a space:

```
// required
...

// optional
...
```

pkg/commands/domain/create.go
```diff
@@ -8,17 +8,20 @@ import (
...
 	cmd.Base
-	manifest       manifest.Data
-	Input          fastly.CreateDomainInput
-	serviceName    cmd.OptionalServiceNameID
-	serviceVersion cmd.OptionalServiceVersion
-	autoClone      cmd.OptionalAutoClone
+	manifest manifest.Data
+	// required
+	name           string
+	ServiceName    cmd.OptionalServiceNameID
```
Integralist, line 20: Is there a reason this is a publicly exported field and not `serviceName`?

pkg/commands/logging/bigquery/bigquery_test.go
```diff
@@ -126,7 +126,7 @@ func TestUpdateBigQueryInput(t *testing.T) {
 			want: &fastly.UpdateBigQueryInput{
 				ServiceID:         "123",
 				ServiceVersion:    4,
-				Name:              "log",
+				Name:              *fastly.String("log"),
```
Integralist, line 129: I get the sense there was some automation script running here 🙂 

I don't think we need to wrap a string in a pointer and then dereference the pointer 😄 

If you could look out and correct these overlaps.

pkg/commands/logging/bigquery/create.go
```diff
@@ -84,31 +84,31 @@ func (c *CreateCommand) ConstructInput(serviceID string, serviceVersion int) (*f
 
 	input.ServiceID = serviceID
 	input.ServiceVersion = serviceVersion
-	input.Name = c.EndpointName
-	input.ProjectID = c.ProjectID
-	input.Dataset = c.Dataset
-	input.User = c.User
-	input.Table = c.Table
-	input.SecretKey = c.SecretKey
+	input.Name = &c.EndpointName
```
Integralist, line 87: Hmm, we're being inconsistent here but maybe there is a reason that I'm not aware of...

e.g. `&someVariable` vs `fastly.String(someVariable)`.

Integralist (changes requested):
This is great! Thanks @grantstephens for taking the time to work on this. There's a lot of nice but subtle clean-up improvements made in here that I really like.

I have a couple of comments/requests. Take a look through and either amend or let me know your thoughts. Thanks!

pkg/commands/backend/create.go
```diff
@@ -67,30 +77,30 @@ func NewCreateCommand(parent cmd.Registerer, globals *config.Data, data manifest
 		Action: c.autoClone.Set,
 		Dst:    &c.autoClone.Value,
 	})
-	c.CmdClause.Flag("name", "Backend name").Short('n').Required().StringVar(&c.input.Name)
-	c.CmdClause.Flag("address", "A hostname, IPv4, or IPv6 address for the backend").Required().StringVar(&c.input.Address)
-	c.CmdClause.Flag("comment", "A descriptive note").StringVar(&c.input.Comment)
-	c.CmdClause.Flag("port", "Port number of the address").Action(c.port.Set).UintVar(&c.port.Value)
+	c.CmdClause.Flag("name", "Backend name").Short('n').Required().StringVar(&c.name)
+	c.CmdClause.Flag("comment", "A descriptive note").Action(c.comment.Set).StringVar(&c.comment.Value)
+	c.CmdClause.Flag("address", "A hostname, IPv4, or IPv6 address for the backend").Required().StringVar(&c.address)
```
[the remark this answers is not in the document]
grantstephens: That all makes sense. In essence I think this means that anything that isn't a pointer in the `Create*Input` should be required?

pkg/commands/domain/create.go
grantstephens: Ah oops- I think most of them have spaces so I'll add the spaces back.

pkg/commands/backend/create.go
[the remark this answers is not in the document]
Integralist: > In essence I think this means that anything that isn't a pointer in the Create*Input should be required?

Yeah, I that sounds right to me. I should have made anything that wasn't required a pointer so 👍🏻

pkg/commands/logging/bigquery/create.go
grantstephens: No- this is another tricky one- some packages used the fastly.X() a lot, and others didn't. I think the right way to do it would be to use fastly.X only when needing to create the pointer on the fly, i.e. not putting a variable into the fastly.X. Or in other words: `fastly.String("hello")` is ok, but `&myStringVar` should be used elsewhere? Does that make sense?

pkg/commands/logging/bigquery/bigquery_test.go
grantstephens: Oh dear! Fixed in https://github.com/fastly/cli/pull/707/commits/a8210d4bb2b90b9427b452f229817faa14326835

pkg/commands/domain/create.go
grantstephens: Done in https://github.com/fastly/cli/pull/707/commits/a8210d4bb2b90b9427b452f229817faa14326835

pkg/commands/dictionary/dictionary_test.go
grantstephens: Cool: https://github.com/fastly/cli/pull/707/commits/a8210d4bb2b90b9427b452f229817faa14326835

pkg/commands/logging/bigquery/create.go
Integralist: Ah, I see. Yup, that makes sense. Thanks for clarifying 👍🏻

pkg/commands/backend/create.go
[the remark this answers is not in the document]
grantstephens: Ok, this should not be done in https://github.com/fastly/cli/pull/707/commits/fb308e91a48872c490c6401fffab15ebf8d0dd27 and https://github.com/fastly/cli/pull/707/commits/0fb629bd3f759ce064c6ad4cba1ab4a7d8add832

pkg/commands/backend/create.go
```diff
@@ -67,30 +77,30 @@ func NewCreateCommand(parent cmd.Registerer, globals *config.Data, data manifest
 		Action: c.autoClone.Set,
 		Dst:    &c.autoClone.Value,
 	})
-	c.CmdClause.Flag("name", "Backend name").Short('n').Required().StringVar(&c.input.Name)
-	c.CmdClause.Flag("address", "A hostname, IPv4, or IPv6 address for the backend").Required().StringVar(&c.input.Address)
-	c.CmdClause.Flag("comment", "A descriptive note").StringVar(&c.input.Comment)
-	c.CmdClause.Flag("port", "Port number of the address").Action(c.port.Set).UintVar(&c.port.Value)
+	c.CmdClause.Flag("name", "Backend name").Short('n').Action(c.name.Set).StringVar(&c.name.Value)
```
Integralist, line 80: Not essential, but I'd appreciate it if it was done as part of this PR:
Could we make sure the flags in each command are sorted in ascending order (just makes it easier when looking at the --help output to find flags).

Integralist (approved):
One comment, but otherwise approving.

pkg/commands/backend/create.go
grantstephens: Ok, all sorted in https://github.com/fastly/cli/pull/707/commits/cbe12eebe989695bad94f383abb85726e0433f89

Integralist (approved):
Fantastic work @grantstephens ❤️

PR #715, opened 2022-11-15
fix object store TOML definitions, add test data

joeshaw:
This fixes the object store configs to match the Viceroy implementation, and adds object store definitions to the test data.

I think it remains an open question as to whether we should change the Viceroy implementation to be more consistent with the existing local server definitions (backends, dictionaries).

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -249,9 +249,9 @@ type SetupLogger struct {
 
 // LocalServer represents a list of mocked Viceroy resources.
 type LocalServer struct {
-	Backends     map[string]LocalBackend     `toml:"backends"`
-	Dictionaries map[string]LocalDictionary  `toml:"dictionaries,omitempty"`
-	ObjectStore  map[string]LocalObjectStore `toml:"object_stores,omitempty"`
+	Backends     map[string]LocalBackend       `toml:"backends"`
+	Dictionaries map[string]LocalDictionary    `toml:"dictionaries,omitempty"`
+	ObjectStore  map[string][]LocalObjectStore `toml:"object_store,omitempty"`
 }
 
 // LocalBackend represents a backend to be mocked by the local testing server.
@@ -272,8 +272,8 @@ type LocalDictionary struct {
 // LocalObjectStore represents an object_store to be mocked by the local testing server.
 type LocalObjectStore struct {
 	Key  string `toml:"key"`
-	Path string `toml:"path"`
-	Data string `toml:"data"`
+	Path string `toml:"path,omitempty"`
+	Data string `toml:"data,omitempty"`
 }
 
 // Exists yields whether the manifest exists.
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -307,6 +307,12 @@ func (f *File) SetOutput(output io.Writer) {
 	f.output = output
 }
 
+func (f *File) logErr(err error) {
+	if f.errLog != nil {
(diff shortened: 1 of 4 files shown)


joeshaw:
The inconsistencies I see:

1. The name of the object store section is `object_store`, singular.  `backends` and `dictionaries` are plural.
2. Dictionaries are populated with a struct containing a key-value map, whereas object stores are populated with a list of structs.  Personally I think the object store implementation is nicer to use, but they are different.
3. For loading data from files, `dictionaries` uses `file` and `object_store` uses `path`.

Go structure for dictionaries:

```go
type LocalServer struct {
	Backends     map[string]LocalBackend       `toml:"backends"`
	Dictionaries map[string]LocalDictionary    `toml:"dictionaries,omitempty"`
	ObjectStore  map[string][]LocalObjectStore `toml:"object_store,omitempty"`
}

type LocalDictionary struct {
	File     string            `toml:"file,omitempty"`
	Format   string            `toml:"format"`
	Contents map[string]string `toml:"contents,omitempty"`
}
```

which maps to a toml like:

```toml
    [local_server.dictionaries.toml]
      format = "inline-toml"

    [local_server.dictionaries.toml.contents]
      foo = "bar"
      baz = """
qux"""
```

Go structure for object stores:

```go
type LocalServer struct {
	Backends     map[string]LocalBackend       `toml:"backends"`
	Dictionaries map[string]LocalDictionary    `toml:"dictionaries,omitempty"`
	ObjectStore  map[string][]LocalObjectStore `toml:"object_store,omitempty"`
}

type LocalObjectStore struct {
	Key  string `toml:"key"`
	Path string `toml:"path,omitempty"`
	Data string `toml:"data,omitempty"`
}
```

which maps to a toml like:

```toml
  [local_server.object_store]

    [[local_server.object_store.store_two]]
      key = "first"
      data = "This is some data"

    [[local_server.object_store.store_two]]
      key = "second"
      path = "strings.json"
```

or, IMO even nicer:

```toml
  [local_server.object_store]
    store_one = [{key = "first", data = "This is some data"}, {key = "second", path = "strings.json"}]
```

Integralist:
Thanks @joeshaw PRs are definitely welcomed, especially for the consistency issues which are quite frustrating from a devex perspective 👍🏻

pkg/manifest/manifest.go
```diff
@@ -249,9 +249,9 @@ type SetupLogger struct {
 
 // LocalServer represents a list of mocked Viceroy resources.
 type LocalServer struct {
-	Backends     map[string]LocalBackend     `toml:"backends"`
-	Dictionaries map[string]LocalDictionary  `toml:"dictionaries,omitempty"`
-	ObjectStore  map[string]LocalObjectStore `toml:"object_stores,omitempty"`
+	Backends     map[string]LocalBackend       `toml:"backends"`
+	Dictionaries map[string]LocalDictionary    `toml:"dictionaries,omitempty"`
+	ObjectStore  map[string][]LocalObjectStore `toml:"object_store,omitempty"`
```
Integralist, line 254: @chrispoole643 were you documenting the ObjectStore in https://developer.fastly.com/reference/compute/fastly-toml/ ? I thought I recalled you doing so but couldn't see it in there at the moment. At any rate, we'll want to ensure that's kept consistent with what's here ☝🏻

Integralist (approved):

joeshaw:
I opened https://github.com/fastly/Viceroy/pull/206 to address it.  If it's merged I'll adjust things here too.

PR #722, opened 2022-12-02
CI/Dockerfiles: minor dockerfiles improvements

dcaba:
* CI: suggesting to execute the PR test flow also on push events (*useful while developing? At least it was for me*)
* CI: executing a dumb docker-build for each Dockerfile to ensure we can generate images / there are no syntax issues or other base-image changes (we are pointing to `latest`!) that make break `RUN` statements.
* Dockerfiles: consolidating `RUN` stages to reduce the number of images
* Dockerfiles: cleaning apt cache after installing software, to save some space / make resulting image a bit slimmer, but..
* for the rust Dockerfile: caching the crates index to save time when building C@E apps (*so bye bye disk savings*)
* Dockerfiles: Consistent usage of `curl --silent` to avoid verbose docker-build output
* Dockerfiles: Removing bash-specific syntax that was generating errors in the output and probably breaking checksum validations
* Dockerfiles: Replacing `adduser` by `useradd` to avoid errors in the output (as `adduser` is expected to be interactive)

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -144,3 +144,12 @@ jobs:
           TEST_COMPUTE_INIT: true
           TEST_COMPUTE_BUILD: true
           TEST_COMPUTE_DEPLOY: true
+  docker-builds:
+    runs-on: ubuntu-latest
+    steps:
+    - name: "Checkout code"
+      uses: actions/checkout@v3
+    - name: Build docker images
+        run: |
+          for dockerFile in Dockerfile*; do docker build -f $dockerFile . ; done
+          
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -150,6 +150,6 @@ jobs:
     - name: "Checkout code"
       uses: actions/checkout@v3
     - name: Build docker images
-        run: |
-          for dockerFile in Dockerfile*; do docker build -f $dockerFile . ; done
+      run: |
+        for dockerFile in Dockerfile*; do docker build -f $dockerFile . ; done
           
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: pull_request
 name: Test
 
 # Stop any in-flight CI jobs when a new commit is pushed.
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -1,4 +1,4 @@
-on: pull_request
+on: [push, pull_request]
 name: Test
 
 # Stop any in-flight CI jobs when a new commit is pushed.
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_test.yml
```diff
@@ -1,4 +1,4 @@
-on: pull_request
+on: [push, pull_request]
```
Integralist, line 1: So currently (due to `pull_request`), once a PR is open, every push to the PR causes the pr_test.yml jobs to be triggered but what happens when `push` is added as well? I presume that a push to a PR will cause two separate jobs to spin up (which would be an unnecessary extra workload).

I typically run the full test suite locally while developing and so I don't use CI to validate my work while I'm developing. Instead, I use the local `make test` (see [`docs/TESTING.md`](https://github.com/fastly/cli/blob/main/TESTING.md)) and when ready to open a PR, this will cause the full test suite to be run again (just to ensure there are no odd things with the CI environment).

Dockerfile-rust
```diff
@@ -2,21 +2,21 @@ FROM rust:latest
...
-
-RUN tar -xzf fastly.tar.gz --directory /usr/bin && rm fastly.tar.gz
+  && if [ "$dlsha" != "$expected" ]; then echo "shasums don't match" && exit 1; fi \
+  && tar -xzf fastly.tar.gz --directory /usr/bin && rm -f sha256sums fastly.tar.gz \
+  && useradd -ms /bin/bash fastly
 
-RUN adduser fastly
 USER fastly
+# this forces the download of the crates index; it will save build time if the image was recently built
+RUN cargo search --limit 0
```
Integralist, line 19: Outside the concern of this PR but just thinking we should really have a `Dockerfile-go` as well.

dcaba: +1!! Our cli depends on the specific-programming-language build tooling, and I dont think that a common image bringing all languages tooling is going to scale well (will make the image even more huge). What we may want to do is to consolidate all Dockerfiles in a single one: a common stage can just install the fastly-cli, and then we can have dedicated stages per language we are supporting, each stage picking the common fastly-cli and adding the language-specific tooling (or inheriting it from the right base image). Then, rather than doing a docker-build against a specific Dockerfile, CI and users will need to be explicit on the stage we want to build (the common fastly-cli layers will be reused if you build more than one image/language, which is good for caching)

.github/workflows/pr_test.yml
dcaba: got it, and no probl.. I missed that doc / instructions of how to run tests locally, but it is fine pointing to them

dcaba: reverted, marking as resolved

Dockerfile-rust
Integralist: @dcaba Nice. I bow to your knowledge in this area 🙇🏻  (ps, PRs welcome 😛)

PR #727, opened 2022-12-06
Update tinygo default build command

Integralist:
The flag `-wasm-abi=generic` doesn't need to be explicitly set as it is the default behaviour.


Integralist:
cc @dkegel-fastly

Integralist:
I manually tested and deployed an application using the default starter kit for Go and it all worked fine.

PR #728, opened 2022-12-06
Switch JavaScript build script based on webpack in package.json prebuild

Integralist:
**Problem**: Not all JS starter-kits require web pack, but the default build script in the CLI will try to execute webpack regardless.
**Context**: The default build script in the CLI is there to support CLI versions <4.0.0 that see the `[scripts.build]` field as optional (i.e. we want starter kits to continue to work with older CLI versions).
**Solution**: Check for `prebuild: webpack` in the package.json and if there, then we know our default build script should call webpack.


harmony7 (approved):
Great! thanks for this.

PR #730, opened 2022-12-08
Link to Starter Kits during `compute init`

Integralist:
There are two related changes made in this PR:

1. Relax the git regex to allow a GitHub URL without a `.git` suffix. This means when a user looks up a Starter-Kit on the Fastly Developer Hub, they'll be able to just copy/paste the standard HTTP URL.

2. Add a link to the Fastly Developer Hub 'Starter-Kits' page during the `compute init` flow so that users know there are more kits than are listed in the terminal output (screenshot below).

Screenshot 2022-12-08 at 17 10 37

### DIFF
--- a/pkg/commands/compute/init.go
+++ b/pkg/commands/compute/init.go
@@ -28,7 +28,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


chrispoole643 (approved):
LGTM

PR #732, opened 2022-12-12
build(dependency): bump google/go-github

Integralist:


Integralist:
Going to close this as we remove this dependency in https://github.com/fastly/cli/pull/739

PR #738, opened 2022-12-15
WIP: Migrate from Kingpin to urfave/cli

### LINKED ISSUE #673: Migrate from Kingpin to https://github.com/urfave/cli
Kingpin is unmaintained and Fastly has forked a version of it that uses an unstable branch because of the necessity to access a v3 feature set that was never actually finished. The CLI has lots of hacks in it to work around missing features.

We should migrate to a maintained parser, and I propose https://github.com/urfave/cli

Here are some docs for getting started:
https://cli.urfave.org/v2/getting-started/

If I get time I'll open a PR to start experimenting with what's needed/missing and whether we need a different CLI parser.

grantstephens:
This is to close #673 

Still a WIP.
Done:
- Basics wired up
- Pattern for adding subcommands
- Exit codes (Untested)

To do:
- Actions
- json output
- Help formatting
- Testing
- Code gen for all commands as rewriting doesn't seem worth it.


dgryski:
It's likely not "too hard" to make a semgrep rule to handle the conversion.  Let me know if you want me to give it a shot.

grantstephens:
> It's likely not "too hard" to make a semgrep rule to handle the conversion. Let me know if you want me to give it a shot.

Oh wow- thank you- that would be amazing. I'm still torn between trying to generate it from go-fastly as it feels like that would be less maintenance in the long term. However if the semgrep rule to convert them isn't "too hard" that would probably result in this being done a lot faster.

Integralist:
Discussed internally and am going to close this out until we have official development time allotted to complete the work.

PR #753, opened 2023-01-06
fix(pkg/commands/compute/serve): prevent 386 arch users executing `compute serve` command

Integralist:
Fixes https://github.com/fastly/Viceroy/issues/200


kailan (approved):
Looks great; thanks!

PR #758, opened 2023-01-16
feat(compute/serve): implement --watch-dir flag

### LINKED ISSUE #646: [FEATURE REQUEST] fastly compute watch does not watch dependant packages
**Feature request**
We have a Rust-based Compute@Edge workspace that has a structure that looks something like
```
/dir
  ./wasm-api
  ./other-crate
```

Where wasm-api is the fastly package and it depends on other-crate in the same directory.

When I `cd` into `wasm-api` and run `fastly compute serve --watch` it will accurately reload when I change files inside 
`wasm-api` (Yay #551 ), but it does not seem to reload when changing files in `other-crate`

I would like to have all files in the root directory be watched. Optimally only things that are directly in the dependency graph, but I'll take "watch all the files" as a "good-enough" solution.

I currently have a script I run to get this behaviour (which doesn't analyze the dependency graph, but just rebuilds on any file changes)
```bash
find . -name "*.rs" -not -path "*./target*" | entr -r -s '(cd ./wasm-api && fastly compute serve)'
```

Integralist:
Fixes #646 

## Context

The `--watch` flag (for the `compute serve` command) indicates to the Fastly CLI that all files within the Compute@Edge project directory need to be watched for changes.

The Fastly CLI will respect a `.ignore` or `.gitignore` file within the project directory (it also respects the user's _global_ git ignore configuration as well).

## New Feature

I've added a new `--watch-dir` flag which informs the Fastly CLI which directory it should start watching files from (defaults to the current directory, which is expected to be the associated Compute@Edge project directory). 

## Example

I've created a Compute@Edge project nested within another directory:

```
.
├── compute
│   ├── .gitignore
│   ├── Cargo.toml
│   ├── README.md
│   ├── fastly.toml
│   ├── rust-toolchain.toml
│   ├── src
└── foo

```

> **NOTE:** `foo` (in the above tree output) is just a text file

The Fastly CLI must be run inside the `compute` directory as that is the Compute@Edge project, but now a user can provide the `--watch-dir` flag to indicate that changes to files outside of the directory need to be watched:

```bash
fastly compute serve --watch --watch-dir ../ --verbose
```

The above example uses a relative path from the `compute` directory up to the parent directory, but an absolute path works as well:

```bash
fastly compute serve --watch --watch-dir /Users/integralist/Code/test-projects/testing-fastly-cli/ --verbose
```

> **CAVEAT:** The Fastly CLI only considers one set of ignore files. For example, with the above command executed, although I have `.gitignore` defined inside the `compute` directory, those patterns will not be used. Once the `-watch-dir` flag is set, it will look inside its specified directory for any ignore configuration.

### DIFF
--- a/pkg/commands/compute/compute_test.go
+++ b/pkg/commands/compute/compute_test.go
@@ -99,6 +99,7 @@ func TestServeFlagDivergence(t *testing.T) {
 		"file",
 		"skip-build",
 		"watch",
+		"watch-dir",
 	}
 
 	iter = serveFlags.MapRange()
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -52,6 +52,7 @@ type ServeCommand struct {
 	file      string
 	skipBuild bool
 	watch     bool
+	watchDir  cmd.OptionalString
 }
 
 // NewServeCommand returns a usable command registered under the parent.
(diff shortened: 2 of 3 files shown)


kailan (approved):
Looks good!

PR #717, opened 2022-11-21
Add Secret Store support

awilliams-fastly:
Adds the following commands:

```
fastly secret-store
  create    Create a new secret store
  describe  Retrieve a single secret store
  delete    Delete a secret store
  list      List secret stores

fastly secret-store-entry
  create    Create a new secret within specified store
  describe  Retrieve a single secret
  delete    Delete a secret
  list      List secrets within a specified store
```

PR #769, opened 2023-01-20
feat(compute/deploy): add Secret Store to manifest `[setup]`

awilliams-fastly:
Adds ability to define Secret Stores in the `[setup]` section of a `fastly.toml` config file. The user will be prompted to create the defined stores and secrets on creation of a service when using `fastly compute deploy`.

Related: #717, #764

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -126,16 +126,15 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return nil
 	}
 
-	domains, backends, dictionaries, loggers, kvStores, err := constructSetupKVs(
+	so, err := constructSetupKVs(
 		newService, serviceID, serviceVersion.Number, c, in, out,
 	)
 	if err != nil {
 		return err
 	}
 
-	if err := processSetupConfig(
-		newService, domains, backends, dictionaries, loggers, kvStores,
-		serviceID, serviceVersion.Number, c,
+	if err = processSetupConfig(
+		newService, so, serviceID, serviceVersion.Number, c,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/commands/compute/deploy.go
```diff
@@ -839,37 +837,43 @@ func pkgUpload(spinner text.Spinner, client api.Interface, serviceID string, ver
...
+type setupObjects struct {
+	domains      *setup.Domains
+	backends     *setup.Backends
+	dictionaries *setup.Dictionaries
+	loggers      *setup.Loggers
+	kvStores     *setup.KVStores
+	secretStores *setup.SecretStores
+}
+
 func constructSetupKVs(
```
awilliams-fastly, line 851: See note: https://github.com/fastly/cli/pull/904/files#r1164377390

Integralist: Agreed this should be `constructSetupObjects`

pkg/commands/compute/setup/secret_store.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return nil
+}
+
+// Create calls the relevant API to create the service resource(s).
+func (s *SecretStores) Create() error {
+	if s.Spinner == nil {
+		return fsterrors.RemediationError{
+			Inner:       fmt.Errorf("internal logic error: no text.Progress configured for setup.SecretStores"),
```
Integralist, line 112:
```suggestion
			Inner:       fmt.Errorf("internal logic error: no spinner configured for setup.SecretStores"),
```

Integralist (approved):
One suggested edit otherwise LGTM. Thanks for refactoring the multiple returned objects into a single object, much cleaner and future proof 👍🏻

Integralist:
Oh I forgot about `constructSetupObjects`, yes please if you could update that. Thanks!

PR #776, opened 2023-01-25
fix(compute/deploy): clean-up new service creation on-error

Integralist:


fgsch (approved):
Untested but code wise it LGTM 👍

PR #780, opened 2023-01-27
v5.1.0

Integralist:


mike-carey:
@Integralist Looks like the commit failed to publish the release for some reason.  Would you like me to open an issue on this repo as well?  I have already [created one up on the tap](https://github.com/fastly/homebrew-tap/issues/6) since the formula was updated, but the release does not exist.

PR #789, opened 2023-02-03
update local_server object and secret store formats

joeshaw:
`object_store` becomes `object_stores`
`secret_store` becomes `secret_stores`
`path` becomes `file`

This makes these more consistent with other parts of the manifest like
dictionaries, backends, geolocation, etc.

Companion to https://github.com/fastly/Viceroy/pull/206

### DIFF
--- a/pkg/commands/compute/testdata/init/fastly-viceroy-update.toml
+++ b/pkg/commands/compute/testdata/init/fastly-viceroy-update.toml
@@ -35,24 +35,24 @@ name = "Default Rust template"
       baz = """
 qux"""
 
-  [local_server.object_store]
+  [local_server.object_stores]
     store_one = [{key = "first", data = "This is some data"}, {key = "second", path = "strings.json"}]
 
-    [[local_server.object_store.store_two]]
+    [[local_server.object_stores.store_two]]
       key = "first"
       data = "This is some data"
 
-    [[local_server.object_store.store_two]]
+    [[local_server.object_stores.store_two]]
       key = "second"
-      path = "strings.json"
+      file = "strings.json"
 
-  [local_server.secret_store]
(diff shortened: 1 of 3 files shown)


awilliams-fastly (approved):

joeshaw:
Lint CI error is caused by an old Go version.  I filed https://github.com/fastly/cli/pull/790 for that.

I think this PR is probably not enough currently as-is.  We might need to migrate the old `object_store` key to the new `object_stores` key.  I'm not sure the best way to do that.

Integralist:
👋🏻 

This might be a breaking interface change for the `fastly.toml` manifest which would require the `manifest_version` field to be bumped from `2` to `3`. But after reviewing its documentation page https://developer.fastly.com/reference/compute/fastly-toml/ I don't see either `object_store` or `secret_store` documented.

So I'm not sure how we proceed because normally this would be considered a 'new addition' (rather than a breaking interface change, because nothing has been documented) but I had a feeling there were people already using these local_server fields (maybe as part of beta access) so by changing them here we're going to break projects/expectations (even if it's currently undocumented).

Also, I know that @chrispoole643 has an internal PR open to document `object_store` so you might need to co-ordinate with him regarding the pluralisation `object_stores` as well as documentation for `secret_stores`.

triblondon:
I'm doing the docs right now and I'd like to call it `object_stores`.  Because these products are all beta and not previously documented, breaking change should be fine.

Integralist:
**UPDATE**: 
I spoke with @triblondon who is working on documenting these features and he is going to be using the plurals, and that as users would have been testing these features under a 'beta' PLC it's fine for us to change the interface (and as it's not documented officially, yet, we can treat this as the final implementation and so leave the `manifest_version` at version `2`). 

**So in summary**: This PR is fine to go ahead. I've also approved its [dependant PR](https://github.com/fastly/cli/pull/790).

PR #800, opened 2023-02-07
feat(resource-link): Add Service Resource commands

awilliams-fastly:
https://developer.fastly.com/reference/api/services/resource/

This adds a `resource-link` sub-command which has associated create, delete, list, and update commands.

```
USAGE
  fastly resource-link <command> [<args> ...]

Manipulate Fastly service resource links

COMMANDS
  resource-link
  create    Create a Fastly service resource link
  delete    Delete a resource link for a Fastly service version
  describe  Show detailed information about a Fastly service resource link
  list      List all resource links for a Fastly service version
  update    Update a resource link for a Fastly service version

SEE ALSO
  https://developer.fastly.com/reference/cli/resource-link/
```

Upgraded github.com/fastly/go-fastly/v7 v7.1.0 => v7.2.0

Some of the names in the API may be confusing. I've done my best to add clarification whenever possible. Specifically:

- A resource, as named by the API, is truly a _link_ between a service and a resource (e.g. an Object Store), not the resource itself. I've tried to use "resource **link**" wherever appropriate, as also used by the API docs:
   > A resource represents a link between a resource type and a service version.
  
  Update https://github.com/fastly/cli/pull/800/commits/1d2eb20e521edf0e4afcc91404b8d5b16d58301d: uses a different base command name `resource-link`.
- ~In some places, the API mixes `id` and `resource_id` which both refer to the ID of the resource _link_, and not the resource itself. I've changed this so that `id` always refers the resource link, and `resource_id` to the linked to object.~ Fixed by https://github.com/fastly/go-fastly/pull/394
- It's possible to provide a resource link name. If set, this is an alias for the resource. Otherwise, whatever name was used when creating the resource itself will be the name to reference from the service for the resource. I've added to the description that this is an _alias_. See this code comment: https://github.com/fastly/go-fastly/blob/ef4694a59779f202d8ea630b41333eb2b76969e8/fastly/resource.go#L86-L88

Keeping this as a draft until a new `go-fastly` release is created that incorporates https://github.com/fastly/go-fastly/pull/394.


pkg/commands/serviceresource/create.go
```diff
@@ -0,0 +1,120 @@
...
+			"Service Version": c.input.ServiceVersion,
+		})
+		return err
+	}
+
+	if ok, err := c.WriteJSON(out, resource); ok {
+		return err
+	}
+
+	text.Success(out, "Created service resource link %s on service %s version %s", resource.ID, resource.ServiceID, resource.ServiceVersion)
```
joeshaw, line 118: It would be good to include the name here, I think, because that's what folks will be using in their C@E applications to open these resources.

pkg/commands/serviceresource/create.go
```diff
@@ -0,0 +1,120 @@
...
+	jsonOutput
+
+	autoClone      cmd.OptionalAutoClone
+	input          fastly.CreateResourceInput
+	manifest       manifest.Data
+	serviceVersion cmd.OptionalServiceVersion
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent cmd.Registerer, globals *global.Data, data manifest.Data) *CreateCommand {
```
Integralist, line 26: FYI for consistency across the code base I've switched the argument names to `g` and `m`, for example:

```suggestion
func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *CreateCommand {
```

Integralist (changes requested):

joeshaw:
Bikeshedding on the command name, I wonder if `service-link` or `resource-link` are clearer?  The _resource_ itself (object store, secret store, etc.) is already created.  This is linking it to a particular service.

pkg/commands/serviceresource/create.go
awilliams-fastly: Updated by [7b67275](https://github.com/fastly/cli/pull/800/commits/7b67275bf290b5eb4cfe0a7d7d8ea6a15165b441)

pkg/commands/serviceresource/create.go
awilliams-fastly: Added in 0882eb33e128433a2ca943473cbab73b270e9383.

Integralist:
@joeshaw to be honest for me either the current `service-resource` or your `service-link` works in my mind, and that's because it all depends on how I think about the command 'in my head' at the time of using the command 😬

| COMMAND | MY INTERNAL UNDERSTANDING WHEN SEEING THE COMMAND |
| :--- | :--- |
| `service-resource` | I have a resource I want to attach to my service |
| `service-link` | I want to link something up to my service |

I'm not sure which 'description' I'm more likely to say in my head when looking at the command.

At a push my gut says `service-link` but then I recall an internal document from Patrick saying why "link" was a bad name and that's why they changed from "Links API" to "Resource API".

awilliams-fastly:
The UI will eventually be updated with a top-level `Resources` tab. That page will contain a list of resources (e.g. Object Stores), with a `Link to services` button alongside each resource.

For this reason, I think that `resource-link` may best align with the UI's verbiage.

pkg/commands/resourcelink/serviceresource_test.go
```diff
@@ -0,0 +1,610 @@
+package resourcelink_test
```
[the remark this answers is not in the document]
awilliams-fastly: Yes, good catch.

Updated in a3cbc6fe942f02a849d555860aca133eb679ac32

pkg/commands/resourcelink/create.go
```diff
@@ -0,0 +1,120 @@
...
+	if c.Globals.Verbose() && c.jsonOutput.enabled {
+		return fsterr.ErrInvalidVerboseJSONCombo
+	}
+
+	serviceID, serviceVersion, err := cmd.ServiceDetails(cmd.ServiceDetailsOpts{
+		AutoCloneFlag:      c.autoClone,
+		APIClient:          c.Globals.APIClient,
+		Manifest:           c.manifest,
+		Out:                out,
+		ServiceNameFlag:    cmd.OptionalServiceNameID{}, // ServiceID flag is required, no need to lookup service by name.
```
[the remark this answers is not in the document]
awilliams-fastly: I'm not sure why I omitted this flag.

Added to all appropriate commands in 3f19765bd8a3f0c63d58c9e6996de0954252f27b.

pkg/commands/resourcelink/json.go
```diff
@@ -0,0 +1,36 @@
+package resourcelink
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/fastly/cli/pkg/cmd"
+)
+
+// jsonOutput is a helper for adding a `--json` flag and encoding
```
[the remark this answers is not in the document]
awilliams-fastly: Nice!

Updated in fc57b4ff17d49505d1fb0b65274fc51c82dca4e1.

Integralist:
CI is failing with some weird error 🤔 

Screenshot 2023-02-17 at 09 58 33

Integralist (approved):
LGTM 👍🏻

awilliams-fastly:
> CI is failing with some weird error 🤔
> 
> Screenshot 2023-02-17 at 09 58 33

Not sure what that was from. But a re-run of the failing CI job passed, and `staticcheck` also passes for me locally.

PR #805, opened 2023-02-08
use secret store client keys when creating secret store entries

joeshaw:


pkg/commands/secretstoreentry/create.go
```diff
@@ -21,6 +22,24 @@ const (
...
 )
 
+// The signing key is a public key that is used to sign client keys.
+// It's meant to be a long-lived key and infrequently (if ever) rotated.
+// Hardcoding it in the CLI gives us the benefit of distributing it via
+// a different channel from the client keys its signing.
+//
+// When we do rotate it, we will need to update this value and release a
+// new version of the CLI.  However, users can also override this with
+// the FASTLY_USE_API_SIGNING_KEY environment variable.
```
Integralist, line 32: @joeshaw where does `FASTLY_USE_API_SIGNING_KEY` get read via the CLI?

Integralist: I see that we check if the value is empty (below) but if the value is set then where/who is consuming that provided value?

joeshaw: The value doesn't matter, that it's set to anything causes the check to be loosened.  This is mostly an escape hatch in case we rotate the signing key at some point in the future and a user has an old CLI version they can't upgrade for some reason, and are willing to take the very small risk of trusting the signing key provided via the API.

joeshaw: Given the test failure, maybe this should be a command-line flag instead?

pkg/commands/secretstoreentry/create.go
```diff
@@ -21,6 +22,24 @@ const (
 	maxSecretLen = maxSecretKiB * 1024
 )
 
+// The signing key is a public key that is used to sign client keys.
+// It's meant to be a long-lived key and infrequently (if ever) rotated.
+// Hardcoding it in the CLI gives us the benefit of distributing it via
+// a different channel from the client keys its signing.
```
awilliams-fastly, line 28:
```suggestion
// a different channel from the client keys it's signing.
```

pkg/commands/secretstoreentry/secretstoreentry_test.go
```diff
@@ -138,6 +181,11 @@ func TestCreateSecretCommand(t *testing.T) {
 
 			opts.APIClient = mock.APIClient(testcase.api)
 
+			// Tests generate their own signing keys, which won't match
+			// the hardcoded value.  Loosen the validation checks.
+			os.Setenv("FASTLY_USE_API_SIGNING_KEY", "1")
+			defer os.Unsetenv("FASTLY_USE_API_SIGNING_KEY")
```
awilliams-fastly, line 187: Could [`testing.Setenv`](https://pkg.go.dev/testing#T.Setenv) be used here?

awilliams-fastly (approved):

pkg/commands/secretstoreentry/secretstoreentry_test.go
joeshaw: I did not know about `testing.Setenv`!  TIL, will definitely do that.

Integralist (approved):

PR #834, opened 2023-02-21
fix(compute/serve): replace separate ignore files with .fastlyignore

Integralist:
**Problem:** supporting multiple ignore files introduced layered complexity such as when the `.ignore` file couldn't override something defined in the `.gitignore` or global ignore file. This ended up causing problems for customers using our Compute Starter Kits.

**Solution:** We already utilise a `.fastlyignore` file for compiling projects (as a way to indicate files to ignore) so we've replaced the existing ignore files with that.


fgsch (approved):
LGTM

PR #842, opened 2023-02-23
Fix `compute build` rendered output

Integralist:
Noticed some issues with how content was being rendered under specific scenarios.

### DIFF
--- a/pkg/commands/backend/create.go
+++ b/pkg/commands/backend/create.go
@@ -213,7 +213,7 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
 		input.Port = fastly.Int(443)
 	}
 
-	if !c.overrideHost.WasSet && !c.sslCertHostname.WasSet && !c.sslSNIHostname.WasSet {
+	if input.Address != nil && !c.overrideHost.WasSet && !c.sslCertHostname.WasSet && !c.sslSNIHostname.WasSet {
 		overrideHost, sslSNIHostname, sslCertHostname := SetBackendHostDefaults(*input.Address)
 		input.OverrideHost = &overrideHost
 		input.SSLSNIHostname = &sslSNIHostname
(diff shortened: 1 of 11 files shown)


pkg/commands/backend/create.go
```diff
@@ -213,7 +213,7 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
 		input.Port = fastly.Int(443)
 	}
 
-	if !c.overrideHost.WasSet && !c.sslCertHostname.WasSet && !c.sslSNIHostname.WasSet {
+	if input.Address != nil && !c.overrideHost.WasSet && !c.sslCertHostname.WasSet && !c.sslSNIHostname.WasSet {
```
Integralist, line 216: Fixing a nil pointer dereference bug

pkg/commands/compute/build.go
```diff
@@ -157,7 +157,7 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return err
 	}
 
-	language, err := language(toolchain, c, out)
+	language, err := language(toolchain, c, in, out, spinner)
```
Integralist, line 160: This change was related to simplifying the Toolchain interface's `Build` method (as we already pass a bunch of things into each language constructor, and some of the data was overlapping).

pkg/commands/compute/build.go
```diff
@@ -167,17 +167,7 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return err
 	}
 
-	postBuildCallback := func() error {
```
Integralist, line 170: To fix the order of rendering I needed to move this logic (and associated prompt functionality) inside the toolchain implementation.

Integralist (commented):
Just some minor comments for future travellers to understand the changes in this PR.

PR #860, opened 2023-03-07
feat(compute/deploy): check service availability

Integralist:
## Problem
Users running `compute deploy` or `compute publish` are immediately presented with a 'Service URL' which when clicked on by a user will typically return a `500 Internal Server` error (as the user's application is still being deployed across Fastly's global network).

## Solution 
Ping the Service URL until an appropriate response is received.

## Notes
The implementation checks for a non-500 response and if it gets one within the specified timeout it will show a 'green tick' to indicate the service URL will work when visited. 

The reason we've chosen this approach, rather than checking for a `200 OK` from the service URL, is that we don't control a user's service and so a 4xx status code for the root path `/` of a service might be a completely valid response! 

Whereas we feel fairly confident that a `500 Internal Server` error is unlikely to be expected behaviour for anyone's service. 

We also ensure our messaging asks users to check their DNS if they have defined a custom domain, as some users will specify a custom domain and activate a service via the CLI and only _then_ will setup the DNS to point to Fastly.

**This is an interim solution until we have either an API or HTTP response header that helps us identify service availability.**

Refer to the following 'example output' gif and see also the updated screenshots in the below thread.

output

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -1112,7 +1112,7 @@ func processService(c *DeployCommand, serviceID string, serviceVersion int, spin
 	if err != nil {
 		return err
 	}
-	msg := "Activating version"
+	msg := fmt.Sprintf("Activating service (version %d)", serviceVersion)
 	spinner.Message(msg + "...")
 
 	_, err = c.Globals.APIClient.ActivateVersion(&fastly.ActivateVersionInput{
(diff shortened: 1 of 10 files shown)


pkg/commands/compute/deploy.go
```diff
@@ -1115,3 +1111,75 @@ func processService(c *DeployCommand, serviceID string, serviceVersion int, spin
...
+func pingServiceURL(serviceURL string) (ok bool, err error) {
+	// gosec flagged this:
+	// G107 (CWE-88): Potential HTTP request made with variable url
+	// Disabling as we trust the source of the variable.
+	// #nosec
+	resp, err := http.Get(serviceURL)
+	if err != nil {
+		return false, err
+	}
+	if resp.StatusCode == http.StatusOK {
```
JakeChampion, line 1181: We don't control the service, which means we can't know what status code the root path should be returning -- a 400 could be a valid response for example -- is there something else we could be using to check service availability?

Integralist: Hmm. Good point!

Integralist: Would checking for something that wasn't a `500 Internal Server Error` be sufficient?

Integralist:
Here are some updated screenshots showing updated remediation messaging and status code responses.

**Success (notice 404 message)**

1  Success (notice 404 message)

**Success (200 OK)**

3  Success (200 OK)

Integralist:
Here's another screenshot demonstrating a timeout (notice the WARNING message)...
4  Timeout (notice the WARNING message)

Integralist:
Here's another screenshot demonstrating the `--status-check-off` flag (for possible automation/CI scenarios)...
5  Success with no status check

Integralist:
Here's another screenshot demonstrating the use of the `--status-check-path` flag which allows a user to configure a specific path to be checked (notice the 404 response status code, because the path I specified wasn't actually handled by the service)...

6  Success with check path flag

This screenshot uses both `--status-check-path` and `--status-check-code`. The latter flag is to indicate a 404 (caused by `--status-check-path`) is what I expect to see from the endpoint being checked...

7  Success with check path and check code flags

Issue #861, opened 2023-03-07
`fastly profile token` doesn't obey `--profile` flag


joeshaw:
**Version**

```
Fastly CLI version v7.0.1 (8ff9a02)
Built with go version go1.20.1 linux/amd64
Viceroy version: viceroy 0.3.2
```

**What happened**

I have several profiles.  I was surprised to see that `fastly profile token -o personal` did not print out my personal token, but instead my default profile's token.

I'll do a little more digging, but at first glance the `TokenCommand` struct has a `profile` field which is never filled in.  https://github.com/fastly/cli/blob/main/pkg/commands/profile/token.go#L17

joeshaw:
Ah, I see there is a `-n` flag, which works.  Any objections to dropping `-n` in favor of `-o`?

JakeChampion:
> Ah, I see there is a `-n` flag to this command, which fills in `c.profile` and works. It's confusing to have both. Any objections to dropping `-n` in favor of `-o`?

I'm in favour of having a single consistent profile selection flag for all commands 👍

Integralist:
Hmm. 

Yeah, I can see why this is confusing.

If we consider the long flag names, then the two possible commands could look like...

```
fastly profile token --name=example
fastly profile token --profile=example
```

I'm not sure if there's too much stutter with `--profile` or not.

PR #862, opened 2023-03-08
breaking(profiles): replace `--name` with positional arg + allow global override

Integralist:
Fixes https://github.com/fastly/cli/issues/861

**NOTE:** Because we're removing a flag from the interface we must consider this a breaking change and consequently the next CLI release will be a new major version.


joeshaw (approved):
👍

PR #873, opened 2023-03-15
fix various lint and CI issues

Integralist:
**NOTE:**
There's lots of warnings in various project CIs related to https://github.com/actions-rs/toolchain and after doing some digging it turns out it's unmaintained and has been for over two years.

So people have been migrating to https://github.com/dtolnay/rust-toolchain (and I'm doing that in this PR).


fgsch (approved):
RGTM 👍

PR #875, opened 2023-03-17
fix(compute/deploy): service availability

Integralist:
This PR includes the following modifications to the service availability check...

1. We only run the service availability check when deploying the service for the first time.
2. We add a countdown to the output message (so users know how long is left before we timeout).
3. We allow `--status-check-code` to be used as the 'expected status code' within the service availability check.

## Screenshots

In the first screenshot, we can see the service availability check (eventually) received a non-500 status code, which signified the service was deployed to/available from the user's nearest Fastly POP...

Screenshot 2023-03-17 at 11 24 06

When running the `fastly compute deploy` command a second time after we already deployed the first time, we see there is no service availability check executed...

Screenshot 2023-03-17 at 11 24 14

In the following screenshot, we can see the service availability check 'in-flight' (i.e. running) and that it contains a timeout which helps users to see that the CLI has not stalled...

Screenshot 2023-03-17 at 11 23 28

In the following screenshot, we can see what happens when a timeout is reached. In the default case we're looking for a non-500 and so the WARNING message indicates this...

Screenshot 2023-03-17 at 11 22 33

Whereas the next screenshot demonstrates how the WARNING message changes when the user sets the `--status-check-code` flag, which tells the CLI the user expects a specific status code from the endpoint being checked...

Screenshot 2023-03-17 at 11 22 40

Finally, the following screenshot demonstrates a contrived example where a user _expects_ their application to return a 500 by setting `--status-check-code` and thus the service availability check finishes almost immediately (but this doesn't mean the service is available and has returned an explicit 500, it's more likely in this scenario that the app hasn't reached the user's nearest POP and so the lookup has failed with a "No service available")...

Screenshot 2023-03-17 at 11 28 44

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -1153,14 +1153,15 @@ func getServiceDomain(apiClient api.Interface, serviceID string, serviceVersion 
 	return name, nil
 }
 
-// checkingServiceAvailability pings the service URL until either there is a 200
-// OK or if the configured timeout is exceeded.
+// checkingServiceAvailability pings the service URL until either there is a
+// non-500 (or whatever status code is configured by the user) or if the
+// configured timeout is reached.
 func checkingServiceAvailability(
 	serviceURL string,
 	spinner text.Spinner,
 	c *DeployCommand,
 ) (status int, err error) {
-	remediation := "The service has been successfully deployed and activated, but the service 'availability' check %s (last status code response was: %d). If using a custom domain, please be sure to check your DNS settings. Otherwise, your application might be taking longer than usual to deploy across our global network. Please continue to check the service URL and if still unavailable please contact Fastly support."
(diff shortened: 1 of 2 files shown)


JakeChampion (approved):
LGTM! I like that this now only happens on service creation and not each deploy ☺️

bstewart00:
Question: I'm creating Fastly services in my application and need to do a similar status check. I was surprised to see that the service initially responds with 500 responses with a plain body "Internal Server Error"  until (presumably) it gets things initialized. Is there a reason Fastly doesn't use `503 Service Unavailable` or something that would be more easy to distinguish from other kinds of 500 e.g. from the origin?

Integralist:
Hi @bstewart00 

Great question, I'm not sure why a 503 wasn't used. I'll enquire internally and see if I can find out the situation, because (agreed) the use of a 500 is not ideal.

PR #878, opened 2023-03-17
🗣️  Pass verbosity flag along to viceroy binary

itsrainy:
Putting this up as a possible solution to https://github.com/fastly/Viceroy/issues/235.

With the 0.4.0 release, the default Viceroy verbosity is "OFF" rather than "INFO" ([discussion here](https://github.com/fastly/Viceroy/pull/229#discussion_r1131582459)). This change would pass along the `--verbose` flag to viceroy when a user calls `fastly compute serve --verbose`.

Alternatively, we could just unconditionally pass the `-v` flag so the `fastly compute serve` output remains ~the same as before that Viceroy version.

### DIFF
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -591,6 +591,7 @@ func local(bin, file, addr, env string, debug, watch bool, watchDir cmd.Optional
 		text.Break(out)
 		text.Output(out, "Wasm file: %s", file)
 		text.Output(out, "Manifest: %s", manifestPath)
+		args = append(args, "-v")
 	}
 
 	s := &fstexec.Streaming{


guybedford (approved):

pkg/commands/compute/serve.go
```diff
@@ -591,6 +591,7 @@ func local(bin, file, addr, env string, debug, watch bool, watchDir cmd.Optional
 		text.Break(out)
 		text.Output(out, "Wasm file: %s", file)
 		text.Output(out, "Manifest: %s", manifestPath)
+		args = append(args, "-v")
```
Integralist, line 594: :wave: Thanks for opening this PR to address this issue. 

CLI users very rarely use --verbose mode because of the extra noise, so I think the default behaviour should be to pass the `-v` flag regardless of whether the verbose mode was set in the CLI (we've already had a few users confused by the missing output because they no longer see the local server address so they don't know where to go).

An alternative would be to keep this change `-v` (as it might address other issues raised specifically for Viceroy, see this PR's description) and instead have the CLI additionally print the address unless the user sets CLI to use verbose mode. This is because on line 71 we set a default address `.Default("127.0.0.1:7676")` so the CLI does _know_ the address up front but we'd want to avoid the duplication of information if viceroy also is in verbose mode.

Integralist: Actually, thinking about this a little more I'm going to go with the 'alternative' option as I think that makes the most sense. So I'll approve/merge this PR and open a separate PR to address the other aspects of the alternative approach.

PR #904, opened 2023-04-11
breaking(objectstore): rename `object-store` command to `kv-store`

dennismartensson:
We are renaming object store to kv store so doing the needed changes here.

Blocked by: https://github.com/fastly/go-fastly/pull/422

### DIFF
--- a/pkg/commands/compute/deploy.go
+++ b/pkg/commands/compute/deploy.go
@@ -126,7 +126,7 @@ func (c *DeployCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return nil
 	}
 
-	domains, backends, dictionaries, loggers, objectStores, err := constructSetupObjects(
+	domains, backends, dictionaries, loggers, kvStores, err := constructSetupKVs(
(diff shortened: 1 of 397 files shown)


Integralist (commented):
LGTM. Only a couple of minor change requests.

Integralist (approved):
LGTM. Will merge once CI is finished (and passing).

pkg/commands/compute/deploy.go
```diff
@@ -839,7 +839,7 @@ func pkgUpload(spinner text.Spinner, client api.Interface, serviceID string, ver
 	return spinner.Stop()
 }
 
-func constructSetupObjects(
+func constructSetupKVs(
```
awilliams-fastly, line 842: Wondering if this should have been left as `constructSetupObjects`? Since `setup.Backends`, for example, aren't key-value stores. Working through merge conflicts in #769 and noticed this.

Integralist: @awilliams-fastly correct! I'm actually fixing this at the moment 🙂

PR #907, opened 2023-04-13
Support latest fastly.toml manifest data model

Integralist:
This PR addresses a few separate issues...

1. Renaming `[setup.dictionaries]` and `[local_server.dictionaries]` to the `config_stores` equivalent.
2. Warning the user if their config contains `[setup.dictionaries]` and provide relevant remediation steps.
3. Splitting the `manifest` package up into separate files for easier maintenance.
4. Refactor the manifest tests.
5. Refactor the manifest `Read()` method, which includes...
    - Removing a bug fix for a very old `v0.25.0` issue.
    - Not auto-updating `manifest_version` (it made incorrect assumptions about manifest interface changes).
6. Updating error remediation message to reference the correct documentation link.

> **WARNING:** We've bumped `manifest_version` to a new version `3` but we're waiting for @triblondon to land an internal documentation update that bumps the version before we cut a new release.


pkg/commands/compute/setup/config_store.go
```diff
@@ -11,54 +11,52 @@ import (
 	"github.com/fastly/go-fastly/v8/fastly"
 )
 
-// Dictionaries represents the service state related to dictionaries defined
+// ConfigStores represents the service state related to config stores defined
 // within the fastly.toml [setup] configuration.
 //
 // NOTE: It implements the setup.Interface interface.
-type Dictionaries struct {
+type ConfigStores struct {
```
JakeChampion, line 18: Config stores are different to dictionaries, they have different naming rules and are also a shared resource like object stores and secret stores. Does change create config store resources and link them to the newly created fastly service?

pkg/manifest/version.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return fmt.Errorf("error parsing manifest_version '%s': %w", s, err)
+	}
+
+	if version > ManifestLatestVersion {
+		return fsterr.ErrUnrecognisedManifestVersion
+	}
+	*v = Version(version)
+	return nil
+}
```
awilliams-fastly, line 79: I see this was moved verbatim from another file. But we could consider reducing the logic here so that there's only 1 path, i.e. skip first trying to parse as int then as float, and instead start directly with `segs := strings.Split(s, ".")`.

I think that'd make the code more maintainable. As is, the logic is slightly different for each method: for example, the `strconv.Atoi` path doesn't enforce that `version` be less than or equal to `ManifestLatestVersion`.

awilliams-fastly (approved):
LGTM

pkg/commands/compute/setup/config_store.go
Integralist: Good shout. I'll double-check in the morning.

pkg/manifest/version.go
Integralist: Sounds good. I'll take a look at a refactor 👍🏻

awilliams-fastly (approved):

PR #927, opened 2023-05-04
feat(kvstore): Bulk Import

Integralist:
This PR introduces both breaking changes _and_ the following features for bulk importing data into the Fastly KV Store.

The breaking changes introduced in this PR are the rename of the `--key-name` flag to `--key` to be consistent with the other 'stores' supported within the CLI.

> **NOTE:** The first two approaches require the value to be base64 encoded as they are being streamed to a Fastly 'batch' API endpoint. The last approach (directory structure) makes individual ["set-value-for-key"](https://developer.fastly.com/reference/api/services/resources/object-store-item/#set-value-for-key) calls that are processed concurrently.

```shell
# stdin
echo '{"key":"example","value":"VkFMVUU="}' | fastly kv-store-entry create -s olwenufofbiklsaaz4i1l8 --stdin

# file
#
# {
#   "key": "file-example-1",
#   "value": "VkFMVUU="
# }
# {
#   "key": "file-example-2",
#   "value": "VkFMVUU="
# }
fastly kv-store-entry create -s olwenufofbiklsaaz4i1l8 --file data.json

# directory
# filename == key, file contents == value
fastly kv-store-entry create -s olwenufofbiklsaaz4i1l8 --dir ./data/
```

Each example is documented in `pkg/app/metadata.json`.

## Screenshots

Example of successful insertion:

Screenshot 2023-05-09 at 15 58 25

Example of failed insertion:

Screenshot 2023-05-09 at 15 58 41


pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +36,63 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+		if s == lookup.SourceUndefined {
+			return fsterr.ErrNoToken
+		}
+
+		// IMPORTANT: Input could be large so we must buffer the reads.
+		// This will avoid loading all of the data into memory at once.
+		body := bufio.NewReader(in)
+
+		resp, err := undocumented.Call(
+			host, path, http.MethodPut, token, io.TeeReader(body, os.Stdout), c.Globals.HTTPClient,
```
Integralist, line 66: I'm only tee'ing the reads so I can see the data get buffered into the HTTP client call.

Integralist:
@awilliams-fastly This is draft for the moment just while I wait for the batch API endpoint to be documented on https://developer.fastly.com/reference/api/ (there is an internal PR for that work). Once it's documented, then I can look to implement the endpoint in the go-fastly API client, and once in a published release I can replace the `undocumented.Call()` with an actual API client method. 

But for the time being, you might be able to assist with a quick review.

Thanks!

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+				filePath := filepath.Join(c.dirPath, file.Name())
+				dir, filename := filepath.Split(filePath)
+				index := strings.Index(dir, base)
+				filename = filepath.Join(dir[index:], filename)
+
+				// G304 (CWE-22): Potential file inclusion via variable
+				// #nosec
+				fileContent, err := os.ReadFile(filePath)
+				if err != nil {
+					errors = append(errors, ProcessErr{
```
awilliams-fastly, line 190: There's a datarace here on `errors`

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+	err = spinner.Start()
+	if err != nil {
+		return err
+	}
+	fileLength := len(files)
+	msg := "Processed %d of %d files"
+	spinner.Message(fmt.Sprintf(msg, 0, fileLength) + "...")
+
+	base := filepath.Base(path)
+	errors := []ProcessErr{}
```
awilliams-fastly, line 153: I typically see this as `var errors []ProcessErr`

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+				sem <- struct{}{}
+				defer func() {
+					<-sem
+				}()
+				defer wg.Done()
+
+				filePath := filepath.Join(c.dirPath, file.Name())
+				dir, filename := filepath.Split(filePath)
+				index := strings.Index(dir, base)
+				filename = filepath.Join(dir[index:], filename)
```
awilliams-fastly, line 184: Wonder if [`strings.TrimPrefix`](https://pkg.go.dev/strings#TrimPrefix) could be used here? Might reduce the number of operations.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+
+				// G304 (CWE-22): Potential file inclusion via variable
+				// #nosec
+				fileContent, err := os.ReadFile(filePath)
```
awilliams-fastly, line 188: I'm guessing that these files could be quite large, and with the default concurrency of 100, could result in a lot of memory usage reading the entire contents into memory.

I'm not though if there's an alternate approach that's possible. Would require a kv store API that accepts something like an `io.Reader`

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+				processed <- struct{}{}
+			}(file)
+		}
+	}
+
+	wg.Wait()
+
+	// NOTE: We add a two second sleep to allow the final goroutine to increment.
+	// Otherwise the StopMessage below is called before filesProcessed is updated.
+	time.Sleep(time.Second * 2)
```
awilliams-fastly, line 219: Consider using a channel or other synchronization method to determine when the `filesProcessed` goroutine has completed.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+					return
+				}
+
+				err = c.Globals.APIClient.InsertKVStoreKey(&fastly.InsertKVStoreKeyInput{
+					ID:    c.Input.ID,
+					Key:   filename,
+					Value: string(fileContent),
+				})
+				if err != nil {
+					errors = append(errors, ProcessErr{
```
awilliams-fastly, line 203: Similar datarace here

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+						File: filePath,
+						Err:  err,
+					})
+					return
+				}
+
+				processed <- struct{}{}
```
awilliams-fastly, line 210: Should this be done regardless of if there's an error processing the file? The related message is "Processed a of b files" which doesn't necessarily mean it was uploaded successfully.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+	for _, file := range files {
+		if !file.IsDir() {
+			wg.Add(1)
+
+			go func(file fs.DirEntry) {
+				// Restrict resource allocation if concurrency limit is exceeded.
+				sem <- struct{}{}
+				defer func() {
+					<-sem
+				}()
```
awilliams-fastly, line 178: Nice use of channels!

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+	// Otherwise the StopMessage below is called before filesProcessed is updated.
+	time.Sleep(time.Second * 2)
+
+	spinner.StopMessage(fmt.Sprintf(msg, filesProcessed, fileLength))
+	err = spinner.Stop()
+	if err != nil {
+		return err
+	}
+
+	outputMsg := "Inserted %d keys into KV Store"
```
awilliams-fastly, line 227: Consider moving into the `if` statement, since that's the only place it's used.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+		return nil
+	}
+
+	text.Error(out, "Encountered the following errors")
+	text.Break(out)
+	for _, err := range errors {
+		fmt.Printf("File: %s\nError: %s\n\n", err.File, err.Err.Error())
+	}
+
+	return nil
```
awilliams-fastly, line 240: Should this return `nil`? If yes, please provide a comment.

pkg/commands/compute/deploy.go
```diff
@@ -445,7 +445,8 @@ type activator func(customerID string) error
 func preconfigureActivateTrial(endpoint, token string, httpClient api.HTTPClient) activator {
 	return func(customerID string) error {
 		path := fmt.Sprintf(undocumented.EdgeComputeTrial, customerID)
-		_, err := undocumented.Get(endpoint, path, token, httpClient)
+		var body io.Reader // nil: no body needed to be sent
+		_, err := undocumented.Call(endpoint, path, http.MethodPost, token, body, httpClient)
```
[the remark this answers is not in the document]
Integralist: Readability. 

I dislike looking at an unfamiliar function and wondering what a `nil` represents 🙂 

I admit it's not the end of the world having the `nil` inlined (because I know I can find it out very quickly) but the variable also lets me document that there is no need to send a body to whatever `Call` is.

I guess I could improve this by having `Call` accept an options struct and then that would...

A.) Allow the code to be clearer as I'm naming the fields.
B.) Allow me to just omit the Body altogether and nil will automatically be inferred.

pkg/commands/kvstoreentry/create.go
Integralist: Yeah, I knew you were going to say that 😄 

I was just being lazy, I'll fix that up 👍🏻

pkg/commands/kvstoreentry/create.go
Integralist: Oh, interesting. Did you find that just using `go build -race` or did it bubble up in some other place?

pkg/api/undocumented/undocumented.go
```diff
@@ -65,14 +74,14 @@ func Get(host, path, token string, c api.HTTPClient) (data []byte, err error) {
...
-		return data, NewError(fmt.Errorf("non-2xx response"), res.StatusCode)
-	}
-
 	data, err = io.ReadAll(res.Body)
 	if err != nil {
 		return []byte{}, NewError(err, res.StatusCode)
 	}
 
+	if res.StatusCode != http.StatusOK {
+		return data, NewError(fmt.Errorf("non-2xx response"), res.StatusCode)
```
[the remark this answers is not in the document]
Integralist: I tweaked this slightly so it checks for 4xx and above (as some undocumented API calls might well return a `201` just as an example) and also to include the body data as part of the response.

pkg/commands/compute/deploy.go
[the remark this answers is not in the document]
Integralist: Fixed in [3d0ff37](https://github.com/fastly/cli/pull/927/commits/3d0ff372d48e7d7552596fb9d80209d61fa234de)

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+	if in == nil || text.IsTTY(in) {
+		return fsterr.ErrNoSTDINData
+	}
+	return c.CallBatchEndpoint(in, out)
+}
+
+// ProcessFile streams a JSON file content to the batch API endpoint.
+func (c *CreateCommand) ProcessFile(out io.Writer) error {
+	f, err := os.Open(c.filePath)
+	if err != nil {
```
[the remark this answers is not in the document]
Integralist: Fixed in [cc94671](https://github.com/fastly/cli/pull/927/commits/cc9467106bc93c76aeb408476e6c914034d0b9e7)

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +28,233 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+
+	go func() {
+		for range processed {
+			filesProcessed++
+			spinner.Message(fmt.Sprintf(msg, filesProcessed, fileLength) + "...")
+		}
+	}()
+
+	for _, file := range files {
+		if !file.IsDir() {
```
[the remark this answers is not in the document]
Integralist: Fixed in [0c451f8](https://github.com/fastly/cli/pull/927/commits/0c451f8827ed3f5f7afcdc9af695cb061f4c4cc1)

pkg/commands/kvstoreentry/create.go
Integralist: Fixed in [4f74c01](https://github.com/fastly/cli/pull/927/commits/4f74c01f64974c1cd398af98a0a1c07a080fab76)

pkg/commands/kvstoreentry/create.go
Integralist: There is a batch API endpoint (not documented) but it was problematic due to the need to be able to adequately audit what specific files failed and with the batch endpoint it wasn't possible to determine that without making additional API calls to list all keys after each batch call which itself is timely/costly. So we opted to avoid the batch endpoint for concurrent processing.

I'll have a think about this more.

pkg/commands/kvstoreentry/create.go
Integralist: Fixed in [71a5369](https://github.com/fastly/cli/pull/927/commits/71a5369b115a9fc3927f3ed9212c8e01a8098752)

pkg/commands/kvstoreentry/create.go
Integralist: The `processed` channel is only sent a message if there were no errors processing the file. 

This means if there was an error then the final message will reflect that the file was not processed.

pkg/commands/kvstoreentry/create.go
Integralist: Yeah I always consult my old gist where I detail three ways to limit concurrency (depending on what I need): https://gist.github.com/Integralist/00ee13f2c2ede5e2200e6ea129d3c43d

pkg/commands/kvstoreentry/create.go
Integralist: Fixed in [0c71ea9](https://github.com/fastly/cli/pull/927/commits/0c71ea924d43bb9d6bba2d2d752d1ea4423632d7)

pkg/commands/kvstoreentry/create.go
Integralist: Ignore me. Fixed in [ffbb29d](https://github.com/fastly/cli/pull/927/commits/ffbb29d7c8ebe3b95c27f754cacac32d5f1b182b)

pkg/commands/kvstoreentry/create.go
Integralist: Fixed in [ffbb29d](https://github.com/fastly/cli/pull/927/commits/ffbb29d7c8ebe3b95c27f754cacac32d5f1b182b)

pkg/commands/kvstoreentry/create.go
Integralist: Ignore me. I replied without actually looking at the code you said there was a data race with. The moment I looked at it I realised of course there would be 🤦🏻 

Fixed in [7f29d04](https://github.com/fastly/cli/pull/927/commits/7f29d041df5f4d4f4cb824dc6833884838c53160)

pkg/commands/kvstoreentry/create.go
Integralist: Fixed in [7f29d04](https://github.com/fastly/cli/pull/927/commits/7f29d041df5f4d4f4cb824dc6833884838c53160)

pkg/commands/kvstoreentry/create.go
Integralist: I have a draft PR for go-fastly trying to figure out what's possible:
https://github.com/fastly/go-fastly/pull/437

pkg/api/undocumented/undocumented.go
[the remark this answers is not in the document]
awilliams-fastly: I meant to say that `fmt.Errorf` could be replaced with `errors.New` unless you plan on adding some variable(s) to the message.

pkg/commands/compute/deploy.go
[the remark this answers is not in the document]
awilliams-fastly: The readability makes sense. Although when I read it, I assume that the body is non-nil since if it were me I would have used `nil`. Can go either way here, but I think your point about using an options struct is valid.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +29,275 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+			filesProcessed++
+			spinner.Message(fmt.Sprintf(msg, "Processing", filesProcessed, fileLength) + "...")
+			if filesProcessed >= uint64(len(files)) {
+				done <- true
+			}
+		}
+	}()
+
+	for _, file := range files {
+		if file.IsDir() {
```
awilliams-fastly, line 179: Wonder if `!file.IsRegular()` would be more appropriate to prevent things like symlinks from being included?

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +29,275 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+		filesProcessed   uint64
+		mu               sync.Mutex
+		wg               sync.WaitGroup
+	)
+
+	go func() {
+		for range processed {
+			filesProcessed++
+			spinner.Message(fmt.Sprintf(msg, "Processing", filesProcessed, fileLength) + "...")
+			if filesProcessed >= uint64(len(files)) {
```
awilliams-fastly, line 172: Wondering what it'd look like if instead of using this condition,  the `processed` channel were closed, causing the for loop to exit? Might be able to remove `IMPORTANT ` warning below in that case.

pkg/commands/kvstoreentry/create.go
```diff
@@ -26,20 +29,275 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+
+	base := filepath.Base(path)
+	processed := make(chan struct{}, c.dirConcurrency)
+	sem := make(chan struct{}, c.dirConcurrency)
+	done := make(chan bool)
+
+	var (
+		processingErrors []ProcessErr
+		filesProcessed   uint64
+		mu               sync.Mutex
```
awilliams-fastly, line 164: Consider a comment about what `mu` is protecting.

pkg/commands/kvstoreentry/create.go
Integralist: @awilliams-fastly I've got the changes to use the go-fastly PR locally and I'll push them up once you've had a chance to review/approve the go-fastly PR. Thanks!

pkg/commands/kvstoreentry/create.go
Integralist: Have fixed this locally. Will push up once new go-fastly release is published.

pkg/commands/kvstoreentry/create.go
Integralist: Fixed locally, will push once new go-fastly client is released.

pkg/api/undocumented/undocumented.go
```diff
@@ -40,20 +40,41 @@ func NewError(err error, statusCode int) APIError {
...
+	HTTPClient  api.HTTPClient
+	HTTPHeaders []HTTPHeader
+	Method      string
+	Path        string
+	Token       string
+}
 
-	req, err := http.NewRequest(http.MethodPost, endpoint, nil)
+// Call calls the given API endpoint and returns its response data.
+// func Call(host, path, method, token string, body io.Reader, c api.HTTPClient, headers ...HTTPHeader) (data []byte, err error) {
```
awilliams-fastly, line 61: Should this comment be deleted?

awilliams-fastly (approved):
LGTM.

There's just a the few remaining comments, but the overall change looks good.

pkg/api/undocumented/undocumented.go
Integralist: Fixed locally. Will push shortly.

pkg/commands/kvstoreentry/create.go
Integralist: Fixed this by replacing done channel with atomic increment and load. Will push shortly.

Integralist:
Thanks @awilliams-fastly 

I think I just need to write some tests for the new bulk import logic and then I'm done.

PR #930, opened 2023-05-10
secret store: Add --recreate and --recreate-must options

awilliams-fastly:
This works in combination with https://github.com/fastly/go-fastly/pull/433 which adds support for the `CreateSecretInput.Method` field.

Secret names must be unique within a store.
The method effects how duplicate names are handled:

- `POST`:  Default. Create a secret; error if one already exists with the same name.
- `PUT` / `--recreate-allow`:   Create or recreate a secret.
- `PATCH` / `--recreate`: Recreate a secret and error if one does not already exist with the same name.

~**DRAFT** until https://github.com/fastly/go-fastly/pull/433 is merged and a new release is created.~

### DIFF
--- a/pkg/commands/secretstoreentry/create.go
+++ b/pkg/commands/secretstoreentry/create.go
@@ -62,14 +62,14 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
 	c.RegisterFlagBool(c.JSONFlag())              // --json
 	c.RegisterFlagBool(cmd.BoolFlagOpts{
 		Name:        "recreate",
-		Description: "Create or recreate if a secret of the same name already exists within the store",
+		Description: "Recreate secret by name (errors if secret doesn't already exist)",
 		Dst:         &c.recreate,
 		Required:    false,
 	})
 	c.RegisterFlagBool(cmd.BoolFlagOpts{
-		Name:        "recreate-must",
-		Description: "Recreate secret of the same name which must already exist within the store",
-		Dst:         &c.recreateMust,
+		Name:        "recreate-allow",
+		Description: "Create or recreate secret by name",
+		Dst:         &c.recreateAllow,
 		Required:    false,
 	})
 	c.RegisterFlagBool(secretStdinFlag(&c.secretSTDIN)) // --stdin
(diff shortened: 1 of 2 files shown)


Integralist:
@awilliams-fastly https://github.com/fastly/go-fastly/releases/tag/v8.2.0

pkg/commands/secretstoreentry/create.go
```diff
@@ -56,8 +57,20 @@ func NewCreateCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *C
...
+	c.RegisterFlag(secretFileFlag(&c.secretFile)) // --file
+	c.RegisterFlagBool(c.JSONFlag())              // --json
+	c.RegisterFlagBool(cmd.BoolFlagOpts{
+		Name:        "recreate",
+		Description: "Create or recreate if a secret of the same name already exists within the store",
+		Dst:         &c.recreate,
+		Required:    false,
+	})
+	c.RegisterFlagBool(cmd.BoolFlagOpts{
+		Name:        "recreate-must",
```
Integralist, line 69: I'm not really a fan of the naming of these flags because of the overloaded nature of the first one.

To me "recreate" means the thing exists and I want it to be recreated. The fact that it will internally behave like a "create", if _by chance_ I happen to have misspelt the name, is more of an unexpected side-effect I think.

It makes `recreate-must` a bit confusing.

I wonder if we rename `recreate` to `recreate-allowed` and `recreate-must` to `recreate`, like so...

```shell
fastly secrets-store-entry create
# Create secret but error if secret name already exists (i.e. POST) 
# e.g. I think the secret doesn't exist, so I'm OK getting an error back telling me it does

fastly secrets-store-entry create --recreate
# Recreate secret but error if secret name doesn't already exists (i.e. PATCH) 
# e.g. I think the secret already exists, so I explicitly state my expectation is a recreate.

fastly secrets-store-entry create --recreate-allowed
# Create or Recreate secret (i.e. PUT)
# e.g. I don't know if the secret already exists, so I'm saying "allow a replacement to happen"
```

awilliams-fastly: I've made your suggested change in https://github.com/fastly/cli/pull/930/commits/263ef6267bf6f6433e6b7482f4850a77d795c32e

pkg/commands/secretstoreentry/secretstoreentry_test.go
```diff
@@ -94,6 +95,10 @@ func TestCreateSecretCommand(t *testing.T) {
 			args:      fmt.Sprintf("create --store-id %s --name %s --stdin", storeID, secretName),
 			wantError: "unable to read from STDIN",
 		},
+		{
+			args:      fmt.Sprintf("create --store-id %s --name %s --stdin --recreate --recreate-must", storeID, secretName),
```
Integralist, line 99: I think this test is referencing the old flag name and I expect CI to fail.

Integralist (approved):

Issue #935, opened 2023-05-16
`fastly backend create` gets rejected by the Fastly API when pointing address to an IP


dcaba:
**Version**

Fastly CLI version v9.0.3 (f6047d9b)
Built with go version go1.20.3 linux/amd64
Viceroy version: viceroy 0.3.2

**What happened**

```
$ fastly backend create --address "1.2.3.4" --port 443 --name=origin --version=active -s serviceID --autoclone --verbose
Fastly API token provided via config file (profile: user)
Fastly API endpoint: https://api.fastly.com

Service ID (via --service-id): serviceID

Service version 1 is not editable, so it was automatically cloned because --autoclone is
enabled. Now operating on version 10.

ERROR: the Fastly API returned 400 Bad Request: Bad request ('' is not a valid Host header).
```

But we were expecting a correct backend creation. I cheated and looked to our internal logs, that point to `override_host` being set but as an empty string, [which is not acceptable (ref to internal codebase)](https://github.com/fastly/Heavenly/blob/master/lib/Heavenly/Model/Backend.pm#L547-L550)

Browsing the code, [this looks the problematic piece](https://github.com/fastly/cli/blob/413c8c23691b99afd3067073540b061f74feabb1/pkg/commands/backend/create.go#L216-L220) as, in absence of an explicit `--override-host` arg, it sets overrideHost to the result of [`SetBackendHostDefaults`](https://github.com/fastly/cli/blob/413c8c23691b99afd3067073540b061f74feabb1/pkg/commands/backend/create.go#L250-L259); the conditional around `LookupAddr()` is not going to set a value for `overrideHost`, which will become the nil value for strings, `""`, rather than letting `input.OverrideHost` (a string pointer) to stay nil.

Bug confirmed with [this extra validation in the createBackend mock](https://github.com/fastly/cli/compare/main...dcaba:fastly-cli:dcaba/backendOverrideHostCannotBeEmpty?expand=1#diff-fb437c006e4c444cb9dceb11e4947a313327312ff42040709e257596c22c27d4R359-R360)

PR #936, opened 2023-05-16
fastly backend create: override host cannot be an empty string

dcaba:
Fixes https://github.com/fastly/cli/issues/935

But there may be better ways of doing this (revisiting `SetBackendHostDefaults()` role or signature would be maybe a more elegant choice?)

### DIFF
--- a/pkg/commands/backend/backend_test.go
+++ b/pkg/commands/backend/backend_test.go
@@ -67,7 +67,7 @@ func TestBackendCreate(t *testing.T) {
 			},
 			WantOutput: "Created backend www.test.com (service 123 version 4)",
 		},
-		// We test that setting a host override does not result in an error
+		// We test that setting an invalid host override does not result in an error
 		{
 			Args: args("backend create --service-id 123 --version 1 --address 127.0.0.1 --override-host overrite-set --name www.test.com --autoclone --port 8080"),
 			API: mock.API{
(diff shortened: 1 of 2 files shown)


pkg/commands/backend/backend_test.go
```diff
@@ -67,6 +67,16 @@ func TestBackendCreate(t *testing.T) {
 			},
 			WantOutput: "Created backend www.test.com (service 123 version 4)",
 		},
+		// We test that setting a host override does not result in an error
```
Integralist, line 70:
```suggestion
		// We test that setting an invalid host override does not result in an error
```

pkg/commands/backend/backend_test.go
```diff
@@ -67,6 +67,16 @@ func TestBackendCreate(t *testing.T) {
...
+		{
+			Args: args("backend create --service-id 123 --version 1 --address 127.0.0.1 --override-host overrite-set --name www.test.com --autoclone --port 8080"),
```
Integralist, line 72:
```suggestion
			Args: args("backend create --service-id 123 --version 1 --address 127.0.0.1 --override-host invalid-host-override --name www.test.com --autoclone --port 8080"),
```

Integralist (approved):
LGTM. Thanks!

PR #940, opened 2023-05-17
fix(kvstoreentry): support JSON output for bulk processing

Integralist:


pkg/commands/kvstoreentry/create.go
```diff
@@ -136,20 +136,34 @@ func (c *CreateCommand) ProcessStdin(in io.Reader, out io.Writer) error {
...
 	if err != nil {
 		c.Globals.ErrLog.Add(err)
 		return err
 	}
 	defer func() {
 		_ = f.Close()
 	}()
-	return c.CallBatchEndpoint(f, out)
+
+	var in io.Reader
```
awilliams-fastly, line 161: Wonder if this line is redundant?

pkg/commands/kvstoreentry/create.go
```diff
@@ -201,13 +215,14 @@ func (c *CreateCommand) ProcessDir(in io.Reader, out io.Writer) error {
 	if err != nil {
 		return err
 	}
-	fileLength := len(filteredFiles)
+	filesTotal := len(filteredFiles)
```
awilliams-fastly, line 218: 👍

awilliams-fastly (approved):

pkg/commands/kvstoreentry/create.go
Integralist: No, I think that's still needed because `f` is a `io.ReadCloser`, not a `*os.File`.

So if I was to do `in := f` and not declare the `var in io.Reader` first, then `in` would be inferred as an `io.ReadCloser` and I couldn't reassign to it from the `io.TeeReader` which only returns an `io.Reader`.

That said, I can simplify this logic actually to...

```go
func (c *CreateCommand) ProcessFile(out io.Writer) error {
  f, err := os.Open(c.filePath)
	if err != nil {
		c.Globals.ErrLog.Add(err)
		return err
	}
	defer func() {
		_ = f.Close()
	}()

  var in io.Reader
  in = f
	if c.Globals.Verbose() {
		in = io.TeeReader(f, out)
	}
	return c.CallBatchEndpoint(in, out)
}
```

PR #952, opened 2023-06-14
Add support for file_max_bytes field to s3 logging endpoint

kellymclaughlin:


pkg/commands/logging/s3/create.go
```diff
@@ -159,6 +161,10 @@ func (c *CreateCommand) ConstructInput(serviceID string, serviceVersion int) (*f
 		input.Domain = &c.Domain.Value
 	}
 
+	if c.FileMaxBytes.WasSet {
```
Integralist, line 164: As mentioned in the related Terraform PR: is this safe to send `0` to the API?

If yes, then nothing more to do here.

If no, then we could either add `&& c.FileMaxBytes.Value > 0` to the condition, or check the value is == 0 within the if block and return an error notifying the user that the input was unexpected.

pkg/commands/logging/s3/describe.go
```diff
@@ -97,6 +97,7 @@ func (c *DescribeCommand) Exec(_ io.Reader, out io.Writer) error {
 	lines := text.Lines{
 		"Bucket":                            o.BucketName,
 		"Compression codec":                 o.CompressionCodec,
+		"File max bytes":                    o.FileMaxBytes,
```
Integralist, line 100: Does the API always return the `file_max_bytes` field? 

If no, then I'm wondering if the go-fastly implementation will cause a zero value (i.e. `0`) to be set on the `FileMaxBytes` response field. In that case we don't want to show `File max bytes: 0` in the CLI output (as demonstrated in your test file changes) because that's misleading the user into thinking something has gone wrong and that they won't be able to have any logs uploaded to S3.

So in that case I think we should only print the values here (and in the list file below) if the value is > 0.

This value could be `0` because the API didn't return the field

Integralist: Ignore my comment. Based on your feedback in the associated Terraform PR it seems the API understands `0` to mean "no size limit" and that's fine as I'm going to presume the API will always send a value for this if it was set and if it's not set it would default to zero.

PR #967, opened 2023-06-27
Correctly check if the package is up to date

fgsch:
Use the files hash instead of the (botched) hashsum approach; the
latter never really worked due to discrepancies between the API and this
tool.


fgsch:
After this PR, trying to deploy the same package shows:

```
INFO: Skipping package deployment, local and service version are identical.
(service REDACTED, version 123)
```

PR #987, opened 2023-07-28
build(scripts): use /usr/bin/env bash to retrieve system bash path

ereslibre:
Use bash as returned by `/usr/bin/env bash`. In systems such as NixOS this returns a path to a valid bash executable in the current environment.


Integralist (approved):

Integralist:
Thanks @ereslibre for this fix 👍🏻

PR #995, opened 2023-08-14
feat(compute/build): support native go

Integralist:
This PR adds support for native Go when using the latest SDK (0.2.0) and an updated Go starter kit ([not yet released](https://github.com/fastly/compute-starter-kit-go-default/pull/8/files)).
For screenshots and details see [here](https://github.com/fastly/cli/pull/995#issuecomment-1680605524)

**This PR is blocking:** https://github.com/fastly/compute-starter-kit-go-default/pull/8

This is NOT a breaking-release, even though we bump the constraints for tinygo and standard go (required for the new Go SDK). That's because we parse the go.mod to see if the user has an SDK version `< 0.2.0` and if so we (at runtime) downgrade the constraint to `>= 0.26.0` rather than use the new constraint of `>= 0.28.1`.

---

_Original PR description/details below..._ 

---

This PR switches from using TinyGo to native Go.

> 🚨 **NOTE:** ~~This appears to be a 'breaking' change (and thus will require a major version release) because ultimately although the user can configure settings in their manifest to control the use of TinyGo, out of the box, the CLI will try and use native Go and that is going to break the experience for existing users.~~ See [update here](https://github.com/fastly/cli/pull/995#issuecomment-1680605524)

~~Users are able to continue using TinyGo by adding a `[scripts.toolchain]` into their `fastly.toml` manifest file:~~

```toml
[scripts]
toolchain = "tinygo"
```

## Examples

⚠️ The only example that might require discussion is the last one.

### 1

The following example demonstrates a new user, with go1.21 installed, and no other non-standard configuration.

We can see that the native Go build command is used (as it now supports WASI).

We can also see that we display a message to the user informing them of what they need to do to continue using TinyGo (if they are trying to build a pre-existing project that previously would have used TinyGo).

1  new user with go1 21

### 2

The following example demonstrates the error the user will see if they have go1.21 installed but have `[scripts.toolchain]` set to TinyGo. 

Notice that TinyGo itself doesn't yet support go1.21.

2  new user with go1 21 but tinygo toolchain

### 3

The following example demonstrates a new user, with go1.18 installed, who has `[scripts.toolchain]` set to TinyGo.

3  new user with go1 18 and tinygo toolchain

### 4

The following example demonstrates the error the user will see if they have go1.18 installed but have `[scripts.toolchain]` set to native Go. 

Notice the CLI has an explicit constraint on `toolchain = "go"` using Go 1.21 and that the version of Go the user has installed doesn't yet support WASI.

4  new user with go1 18 but go toolchain

### 5

The following example demonstrates a new user, with go1.21 installed, who has no `[scripts.toolchain]` set but does have a custom `[scripts.build]` configured to use native Go.

Screenshot 2023-08-14 at 15 04 06

### 6

The following example demonstrates a new user, with go1.18 installed, who has no `[scripts.toolchain]` set but does have a custom `[scripts.build]` configured to use TinyGo.

Notice how although native Go is now the default behaviour, because the user has go1.18 installed, and TinyGo, and they've configured their custom build script to use TinyGo, the CLI does indicate a warning that the constraint wasn't met (and also that the user should add `toolchain = "tinygo"`) but ultimately the build succeeded as they were using all the right tools and scripts.

The reason this is _slightly_ problematic is because the toolchain constraint defaults to native Go (i.e. go1.21) and only changes to go1.18 when `[scripts.toolchain]` is configured. If that's not configured, like in this example, we don't then want to get into the business of trying to manually parse `[scripts.build]` to understand what a user has entered. We could try and do a glob search for "tinygo" but we don't know if a user has tinygo aliased or they downloaded it manually as `tgo` or whatever the situation might be. We can't presume to know the context of the custom build script.

Screenshot 2023-08-14 at 15 08 10


Integralist:
**NOTE:** I had an internal discussion with reviewer @joeshaw and after a lengthy back-and-forth about how to avoid a major release, we decided that a major release was probably the best approach (as none of the alternatives were clear winners and each approach had its own set of caveats).

Integralist:
@joeshaw once you've merged https://github.com/fastly/compute-starter-kit-go-default/pull/8 let me know so I can quickly retest this PRs changes to be sure `compute serve` and deployments work.

.fastly/config.toml
```diff
@@ -1,12 +1,13 @@
...
+config_version = 3
 
 [fastly]
 api_endpoint = "https://api.fastly.com"
 
 [language]
 [language.go]
-tinygo_constraint = ">= 0.26.0-0" # NOTE -0 indicates to the CLI's semver package that we accept pre-releases (TinyGo users commonly use pre-releases).
-toolchain_constraint = ">= 1.18"
+tinygo_constraint = ">= 0.26.0-0"       # NOTE -0 indicates to the CLI's semver package that we accept pre-releases (TinyGo users commonly use pre-releases).
```
joeshaw, line 8: I mentioned this in https://github.com/fastly/compute-starter-kit-go-default/pull/8 but version 0.2.0 of the SDK brings the TinyGo requirement up to >= 0.28.1.

pkg/commands/compute/language_go.go
```diff
@@ -24,7 +25,14 @@ import (
...
 // This makes the experience less confusing as users didn't expect file changes.
-const GoDefaultBuildCommand = "tinygo build -target=wasi -gc=conservative -o bin/main.wasm ./"
+const TinyGoDefaultBuildCommand = "tinygo build -target=wasi -gc=conservative -o bin/main.wasm ./"
+
+// GoDefaultBuildCommand is a build command compiled into the CLI binary so it
+// can be used as a fallback for customer's who have an existing C@E project and
+// who don't have a `scripts.build` defined.
+//
+// NOTE: Only used if `toolchain = "go"` and if Go 1.21+ is installed.
+const GoDefaultBuildCommand = "env GOARCH=wasm GOOS=wasip1 go build -o bin/main.wasm ./"
```
joeshaw, line 35: I am not confident `env` is going to work on Windows, so it's worth checking this if CI doesn't already.

Integralist: Yeah, I wasn't confident this would work on Windows but CI (which runs the full test suite) seems to think there were no issues 🤔 

In general would we just remove the `env` and just prefix the build script with GOARCH/GOOS? Not sure if exec.Command needs `env`.

joeshaw: `env` is required because `exec.Command` doesn't run a shell.  We could set the environment outside of the command-line, but that obviously complicates things if we want the end user control over things.

dgryski: You can set the environment with https://pkg.go.dev/os/exec#Cmd.Environ , although that would require parsing the command line with https://pkg.go.dev/github.com/google/shlex or something to extract the environment variable settings.

Integralist: @dgryski I think it's fine to set the environment without shlex because we know what the _default_ values for those should be (i.e. they're already hard-coded into the binary `GOARCH=wasm GOOS=wasip1`). So I'll look at doing that on Wednesday when I'm back from PTO.

Once the user starts defining their own build script, then things are a little out of our hands but I imagine if someone wants to define their own build script, then they could just do an `export` directly in their shell as a workaround. I'm not sure in practice how many people are going to stray from the pre-defined default paths in the CLI (for either Go or TinyGo)

JakeChampion:
Question: Is there some reasons about why `toolchain` is within `scripts`? I was surprised to see it within `scripts` as it is not like the other entries within `scripts`, which are meant to be commands ran within a shell such as the `build` and `post_init` fields.

Integralist:
@JakeChampion I originally had it top-level but moved it to scripts as it felt out of place at the top level.

Also, as per https://developer.fastly.com/reference/compute/fastly-toml/ `scripts` is for...

> Customisation options for the Fastly CLI build step.

So as this is customising the build process it made sense (to me) to have toolchain added to scripts.

triblondon:
I'm quite confused about scripts.toolchain. I expect my build to be executed using scripts.build, and so I wonder at what point does the scripts.toolchain command get executed on the shell?

triblondon:
Oh, catching up on your comment about the definition of "scripts" @integralist... that's not how I interpret scripts, scripts are surely executable commands?

Integralist:
@triblondon I'm happy to move this back to the top-level if you feel it's more appropriate there.

triblondon:
I am not a Go developer, so feel free to override me if this is expected in the Go community.  I would have a hard time understanding how `toolchain` interacts with `build` (and after reading the PR description I'm only 80% sure I get it).

Is it not possible to just change the default build command in the CLI? It seems to me that we are already going to switch tinygo users to go automatically, with a warning about adding `toolchain` if they want to opt out.  So instead, couldn't we do this?  If the user has no build script set, we say:

```
No build script found in fastly.toml.  Fastly now recommends the standard Go compiler to build 
Compute@Edge projects, because Go added support for WASI in version x.y.z.  If you were previously
using TinyGo, and want to continue to do so, cancel this build, add the following to your fastly.toml, 
then re-run this build:

[command]

If you are happy to use standard Go, press enter to proceed. Add a scripts.build value in your fastly.toml
file to suppress this message on future builds.
```

Integralist:
The latest commit reflects the new direction we're going in as far as implementation is concerned.

## 1. Pre-existing project with go1.18

Pre-existing projects cloned from our Go starter kit has no explicit built script specified.

2  pre-existing project user still on go1 18

## 2. Pre-existing project with go1.21

Pre-existing projects cloned from our Go starter kit has no explicit built script specified.

1  pre-existing project user upgrades to go1 21

## 3. New project with go1.18

New projects cloned from Go starter kit will have an explicit build script specified.

3  new go project user still on go1 18

## 4. New project with go1.21

New projects cloned from Go starter kit will have an explicit build script specified.

4  new go project user now on go1 21

## 5. New project with go1.18 and custom tinygo build script

The build succeeds because although we have the new Go starter kit bumped to use the new 0.2.0 SDK, the user's tinygo custom script will be using TinyGo not native Go (as per the logic/constraints defined in the latest commit of this PR).

6  new project with new sdk but go1 18 and tinygo build script

## 6. New project with go1.21 and custom tinygo build script

The build fails because we have used the new Go starter kit (bumped to use the new 0.2.0 SDK), and so the user's tinygo custom script will be using TinyGo not native Go and TinyGo < 0.29.0 (not released) doesn't support go1.21.

But luckily TinyGo's error message is very clear, and it indicates to the user that their Go version is not supported and what is supported.

5  new project with new sdk but go1 21 and tinygo build script

Integralist:
@joeshaw @dgryski 

```
        Build script to execute:
        	cmd.exe /C GOARCH=wasm GOOS=wasip1 go build -o bin/main.wasm ./
        
        
        INFO: Command output:
        --------------------------------------------------------------------------------
        'GOARCH' is not recognized as an internal or external command,
        operable program or batch file.
        --------------------------------------------------------------------------------
```

Looks like we'll have to go down the https://pkg.go.dev/os/exec#Cmd.Environ route. 

This would mean _not_ setting those environment variables in the build script command for Windows users.

Integralist:
@joeshaw @dgryski [last commit](https://github.com/fastly/cli/pull/995/commits/78cc351086eab0264698b9d99e8fbf8fdbf3b799) fixed Windows issue

joeshaw:
@Integralist did `env` work previously?

joeshaw:
It looks like `env` does indeed work on Windows, see https://github.com/fastly/compute-starter-kit-go-default/actions/runs/5882217274/job/15952358049 where I test it out.

I think we might want to not automatically set `GOARCH`/`GOOS` because there will be a variation in behavior for TinyGo.  See https://github.com/tinygo-org/tinygo/pull/3861.

pkg/commands/compute/language_go.go
```diff
@@ -125,14 +160,74 @@ func (g *Go) Build() error {
...
+// before and would stop their build from working).
+//
+// NOTE: The `configConstraint` is the latest CLI application config version.
+// If there are any errors trying to parse the go.sum we'll default to the
+// config constraint.
+func identifyTinyGoConstraint(configConstraint string) string {
+	moduleName := "github.com/fastly/compute-sdk-go"
+	version := ""
+
+	f, err := os.Open("go.sum")
```
joeshaw, line 181: Why examine `go.sum` rather than `go.mod`?  I think the module file would be a more reliable indicator of the version the user wanted.

There is a parser for `go.mod` files available in https://pkg.go.dev/golang.org/x/mod/modfile.

`go.sum` can contain multiple versions, so if you stick with `go.sum` you'll need to handle that case.

pkg/commands/compute/language_go.go
```diff
@@ -85,33 +89,64 @@ type Go struct {
 	spinner text.Spinner
 	// timeout is the build execution threshold.
 	timeout int
+	// toolchain indicates the toolchain used for compiling your program.
+	toolchain string
```
joeshaw, line 93: leftover?

pkg/manifest/file.go
```diff
@@ -188,7 +188,13 @@ func appendSpecRef(w io.Writer) error {
...
+	// Build is a custom build script.
+	Build string `toml:"build,omitempty"`
+	// PostBuild is executed after the build step.
 	PostBuild string `toml:"post_build,omitempty"`
-	PostInit  string `toml:"post_init,omitempty"`
+	// PostInit is executed after the init step.
+	PostInit string `toml:"post_init,omitempty"`
+	// Toolchain indicates the toolchain used for compiling your program.
+	// Supported languages: Go (allowed inputs: go, tinygo).
+	Toolchain string `toml:"toolchain,omitempty"`
```
joeshaw, line 199: leftover?

joeshaw:
> because there will be a variation in behavior for TinyGo

Specifically, `tinygo -target=wasi` pretends to be `GOARCH=arm` `GOOS=linux` because the standard library didn't have real support for wasi, so some different code is selected due to those implicit build tags.

pkg/commands/compute/language_go.go
```diff
@@ -85,33 +89,64 @@ type Go struct {
...
+		toolchainConstraint string
+	)
+
 	if g.build == "" {
-		g.build = GoDefaultBuildCommand
-		noBuildScript = true
+		g.build = TinyGoDefaultBuildCommand
+		tinygoToolchain = true
+		toolchainConstraint = g.config.ToolchainConstraintTinyGo
+		text.Info(g.output, "No [scripts.build] found in fastly.toml. Visit https://developer.fastly.com/learning/compute/go/ to learn how to target native Go vs TinyGo.")
```
dgryski, line 109: Is `native Go` really what we want to call this compiler option?

jaskiratr: Can we use `standard Go` ?

pkg/commands/compute/language_go.go
Integralist: I'll switch to `go.mod`

I opted for `go.sum` originally as I thought it would be easier to parse (as go.mod had different ways of defining a dependency) but I neglected to realise that the dependency could appear multiple times.

pkg/commands/compute/language_go.go
Integralist: Whoops! Yes. I'll remove this.

pkg/manifest/file.go
Integralist: Whoops! Yes. I'll remove this.

pkg/commands/compute/language_go.go
Integralist: 👍🏻

Integralist:
@joeshaw `env` works fine for Linux but not Windows...

Screenshot 2023-08-18 at 10 48 30

I'm not sure what problem exists in the test suite implementation that would cause it to fail for the tests but work in practice (as demonstrated by the CI job you linked to).

Integralist:
@joeshaw looks like my tests neglected to include `env` 🤦🏻 🙂 

So all is good now and ready for your final review.

Integralist:
**UPDATE:** The use of `env` doesn't work on a real/fresh Windows laptop (likely GitHub Actions installing tools to make it work). So we're going to add `[scripts.env_vars]` to the Go Starter Kit and add support for that in this CLI PR. This will allow the CLI to set the environment variables required for standard Go to compile correctly, while also allowing a user to override the env vars with their own settings if they so choose.

I'm PTO for just over a week, so I'll wrap up this outstanding work when I return.

joeshaw:
An additional feature request as part of this PR: we should ensure that the binary output is actually a wasm file.  If you ran a `fastly compute build` on a project that supported `scripts.build` but not `scripts.env_vars` you would get a native binary written to `bin/main.wasm`, because `GOARCH` and `GOOS` aren't set.

There's nothing we can do about this for existing versions (and `fastly compute serve` and `fastly compute publish` fail because of this), but going forward this would be a helpful validation step.

The binary format (including the magic and version info at the start of a file) are defined here: https://webassembly.github.io/spec/core/binary/modules.html#binary-module

Integralist:
@joeshaw OK I've implemented the `scripts.env_vars` and the Wasm validation logic.

Have another review and give this PR an official 'approval' if you're happy with it.

Thanks!

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+	_, err = f.Read(magic)
+	if err != nil {
+		return bt.handleError(err)
+	}
+	if len(magic) > 0 && magic[0] == 0x00 {
+		magic = magic[1:] // Remove the first byte which is a null character "\x00"
+	}
+	if string(magic) != "asm" {
+		return bt.handleError(err)
+	}
```
joeshaw, line 226: IMO it'd be better to declare the expected bytes here (`[]bytes{0x00, 0x61, 0x73, 0x6d}`) and using `bytes.Equal` to compare them rather than chopping the nul byte off and doing a string comparison.

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+	// Parse the version
+	versionBytes := make([]byte, wasmBytes)
+	_, err = f.Read(versionBytes)
+	if err != nil {
+		return bt.handleError(err)
+	}
+	version := int(binary.LittleEndian.Uint32(versionBytes))
+	if version <= 0 {
+		return bt.handleError(err)
+	}
```
joeshaw, line 241: This is fine, but you can also do

```go
var version uint32
if err := binary.Read(f, binary.LittleEndian, &version); err != nil {
	return bt.handleError(err)
}
```

to save some lines.

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+	if len(magic) > 0 && magic[0] == 0x00 {
+		magic = magic[1:] // Remove the first byte which is a null character "\x00"
+	}
+	if string(magic) != "asm" {
+		return bt.handleError(err)
+	}
+	if bt.verbose {
+		text.Break(bt.out)
+		text.Description(bt.out, "Wasm module 'magic'", string(magic))
```
joeshaw, line 229: And to add on to my comment above, the "magic" here is the 4 bytes including the nul, so it's important to print it out as such.

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+
+	// Parse the version
+	versionBytes := make([]byte, wasmBytes)
+	_, err = f.Read(versionBytes)
+	if err != nil {
+		return bt.handleError(err)
+	}
+	version := int(binary.LittleEndian.Uint32(versionBytes))
+	if version <= 0 {
+		return bt.handleError(err)
```
joeshaw, line 240: if you leave this as is, `err` will be `nil` here as well.

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+// The encoding of a module starts with a preamble containing a 4-byte magic
+// number (the string '\0asm') and a version field.
+//
+// Reference:
+// https://webassembly.github.io/spec/core/binary/modules.html#binary-module
+func (bt BuildToolchain) validateWasm(path string) error {
+	// gosec flagged this:
+	// G304 (CWE-22): Potential file inclusion via variable
+	// Disabling as we trust the source of the path variable.
+	// #nosec
```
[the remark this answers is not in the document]
Integralist: Looks like gosec is happy with a constant because it can predictably identify the value as hard-coded.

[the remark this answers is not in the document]
Integralist: I've made the change but I personally prefer to inject dependencies wherever possible rather than rely on globals (although I appreciate a constant is a bit different in its intent)

pkg/commands/compute/language_go.go
```diff
@@ -125,14 +159,92 @@ func (g *Go) Build() error {
...
+	}
+
+	// 0.2.0 introduces the break by bumping the TinyGo minimum version to 0.28.1
+	breakingSDKVersion, err := semver.NewVersion("0.2.0")
+	if err != nil {
+		return configConstraint
+	}
+
+	if gomodVersion.LessThan(breakingSDKVersion) {
+		return ">= 0.26.0-0"
```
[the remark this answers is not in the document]
Integralist: I've moved this value to the config.

pkg/commands/compute/language_toolchain.go
Integralist: I wasn't sure how best to display the value so opted for `fmt.Sprintf("%#v", magic)` which displays it as...

```
[]byte{0x0, 0x61, 0x73, 0x6d}
```

If I was to use `%v` instead of `%#v`, then it would have displayed as...

```
[0 97 115 109]
```

Which I didn't feel was as appropriate.

I also didn't want to hard-code the bytes into a string (e.g. `"0x00, 0x61, 0x73, 0x6d"`) as it means effectively having it defined twice (once in a hard-coded string and the other as the assignment to `expectedMagic`).

I also didn't want to have to `range` over the `magic` variable just to get it to display like `"0x00, 0x61, 0x73, 0x6d"` so opted for just `fmt.Sprintf("%#v", magic)`.

pkg/commands/compute/language_toolchain.go
Integralist: Done

pkg/commands/compute/language_toolchain.go
Integralist: oh nice! thanks

Integralist:
@joeshaw one more review if you can please 🙂 🙇🏻

pkg/commands/compute/language_go.go
```diff
@@ -125,14 +159,92 @@ func (g *Go) Build() error {
...
+// version requirement to 0.28.1 and we want to avoid any scenarios where a
+// bump in SDK version causes the user's build to break (which would happen for
+// users with a pre-existing project who happen to update their CLI version: the
+// new CLI version would have a TinyGo constraint that would be higher than
+// before and would stop their build from working).
+//
+// NOTE: The `configConstraint` is the latest CLI application config version.
+// If there are any errors trying to parse the go.mod we'll default to the
+// config constraint.
+func identifyTinyGoConstraint(configConstraint, fallback string) string {
```
dgryski, line 176: If this turns out to be too fragile, we can switch to https://pkg.go.dev/golang.org/x/mod/modfile

pkg/commands/compute/language_toolchain.go
```diff
@@ -179,6 +196,55 @@ func (bt BuildToolchain) Build() error {
...
+	magic := make([]byte, wasmBytes)
+	_, err = f.Read(magic)
+	if err != nil {
+		return bt.handleError(err)
+	}
+	if len(magic) > 0 && magic[0] == 0x00 {
+		magic = magic[1:] // Remove the first byte which is a null character "\x00"
+	}
+	if string(magic) != "asm" {
+		return bt.handleError(err)
```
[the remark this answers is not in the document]
Integralist: Just pushed a change.

pkg/commands/compute/language_go.go
Integralist: Cool thanks. I think our use case is simple enough but yeah if in the real-world it's starts being a problem we can pick up a tool like that 👍🏻

joeshaw (approved):
thank you for your persistence on this. 👍 LGTM

Issue #841, opened 2023-02-23
[FEATURE REQUEST] Please add CONDITIONS to the CLI


leoncowle:
**Is your feature request related to a problem? Please describe.**
Missing the ability to use the CLI to create/manage CONDITIONS.

**Describe the solution you'd like**
Implement a command in the CLI to manage CONDITIONS.

**Describe alternatives you've considered**
Closed issue #497 mentions using it in VCL, but I'm unclear how -- when it's a Request condition that you want to attach to a backend. Alternatively, we can call the API directly. However, this is not the proper solution (if you want to use the CLI at all). Right now, my approach is to have a script which does:
```
curl -X POST "https://api.fastly.com/service/<svcid>/version/<ver>/condition" ... ... ...
fastly backend create --request-condition <condition created using cURL above> ... ... ...
```
Would be much better if we were able to use the CLI for all steps.

Integralist:
Hi @leoncowle 

Nearly all of the UI features (for a 'Deliver' service) are shortcuts to help avoid writing your own VCL code. 
A 'Deliver' service uses Varnish/VCL and so that's why all these UI features essentially generate VCL code.

So when you run something like...

```shell
# Create a 'condition' object
$ curl -i  -X POST "https://api.fastly.com/service/<SERVICE_ID>/version/<VERSION>/condition" -H "Fastly-Key: $FASTLY_API_TOKEN" -H "Content-Type: application/x-www-form-urlencoded" -H "Accept: application/json" -d "name=test-request-condition&statement=req.url.basename+%3D%3D+%22index.html%22&type=REQUEST&priority=10"

# Create a 'backend' object (and attach the condition object to it)
$ fastly backend create --request-condition test-request-condition --name foo --address <ADDRESS> --service-id <SERVICE_ID> --version <VERSION>
```

What is happening is, a 'condition' object is created and within your service a corresponding chunk of VCL code is generated and added inside of a _default_ `vcl_recv` subroutine that Fastly generates for your service. 

That chunk of VCL code looks something like the following, which is the VCL generated from the above example curl/cli command shown above...

```vcl
if (req.url.basename == "index.html") {
  set req.backend = F_foo;
}
```

You can see all the VCL code is really doing is saying "once the condition is met, make sure the 'backend' object 'foo' is used.

> **NOTE:** You can check this yourself by going into the Fastly management UI and clicking on the "{} Show VCL" link (so for example you'll also see the `fastly backend create` command will have produced a 'backend' VCL object).

[...]

PR #1008, opened 2023-09-08
add condition commands

noguxun:
For https://github.com/fastly/cli/issues/841


pkg/commands/condition/create.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Optional flags
+	c.RegisterAutoCloneFlag(cmd.AutoCloneFlagOpts{
+		Action: c.autoClone.Set,
+		Dst:    &c.autoClone.Value,
+	})
+	c.CmdClause.Flag("name", "Condition name").Short('n').Action(c.name.Set).StringVar(&c.name.Value)
+	c.CmdClause.Flag("priority", "Condition priority").Action(c.priority.Set).IntVar(&c.priority.Value)
+	c.CmdClause.Flag("statement", "Condition statement").Action(c.statement.Set).StringVar(&c.statement.Value)
+	c.CmdClause.Flag("type", "Condition type").Action(c.conditionType.Set).StringVar(&c.conditionType.Value)
```
Integralist, line 57: We should use an enum type here as the data model https://developer.fastly.com/reference/api/vcl-services/condition/ for `type` suggests the values are constrained to `REQUEST`, `CACHE`, `RESPONSE`, `PREFETCH`.

See authtoken 'scope' for an example...

- https://github.com/fastly/cli/blob/main/pkg/commands/authtoken/create.go#L18-L20
- https://github.com/fastly/cli/blob/main/pkg/commands/authtoken/create.go#L49

Integralist, line 57: We should use an enum type here as the data model https://developer.fastly.com/reference/api/vcl-services/condition/ for `type` suggests the values are constrained to `REQUEST`, `CACHE`, `RESPONSE`, `PREFETCH`.

See authtoken 'scope' for an example...

- https://github.com/fastly/cli/blob/main/pkg/commands/authtoken/create.go#L18-L20
- https://github.com/fastly/cli/blob/main/pkg/commands/authtoken/create.go#L49

pkg/commands/condition/list.go
```diff
@@ -0,0 +1,119 @@
...
+		Name:        cmd.FlagServiceName,
+		Description: cmd.FlagServiceDesc,
+		Dst:         &c.serviceName.Value,
+	})
+
+	return &c
+}
+
+// Exec invokes the application logic for the command.
+func (c *ListCommand) Exec(in io.Reader, out io.Writer) error {
```
Integralist, line 58: As per GitHub's linter report (below) let's underscore the `in` parameter.

Integralist (changes requested):
Look good. Just two minor edits required to be made and then this will be good to go.

One other thing was that the associated API is a [VCL Object](https://developer.fastly.com/reference/api/vcl-services/) and so I'm wondering if instead of `fastly condition ...` it should be nested under the vcl package like 'custom' and 'snippet' are so the command changes to `fastly vcl condition` (I think that might make it clearer what a condition is).

Thanks!

pkg/commands/condition/create.go
noguxun: > We should use an enum type..

Thanks for the example, I will try to make it an enum.  out of the scope of this PR, but I notice that in go-fastly SDK, the Type of condition is not an enum either. 

>  it should be nested under the vcl package like 'custom' and 'snippet' are so the command changes to fastly vcl condition (I think that might make it clearer what a condition is).

I see, that is a very good point,  will try to make it a sub-command under the `vcl` , which can be clearly maps to[ vcl objects ](https://developer.fastly.com/reference/api/vcl-services/) groups

noguxun:
Hi @Integralist , I've made a new commit according to your review.

pkg/app/run_test.go
```diff
@@ -62,6 +62,7 @@ acl-entry
 auth-token
 backend
 compute
+condition
```
Integralist, line 65: This should now be removed.

Integralist (changes requested):
Looks great. Just a couple of minor tweaks required.

Integralist (approved):
🎉❤️

LGTM. Thanks!

PR #1011, opened 2023-09-14
add tinygo starter kit

joeshaw:


dgryski (commented):
LGTM but I'll leave this for someone who actually works on this code.

Integralist (commented):
I think you're going to need to also bump the `config_version` inside of the `.fastly/config.toml` (which is merged with the output of this script to produce the final config.toml that's embedded into the CLI -- https://github.com/fastly/cli/blob/main/DEVELOP.md#fastlyconfigtoml).

If we don't bump the version, then the CLI will just end up using the config file already on the user's machine. 

Where as bumping the version causes this line to get executed when the user invokes a CLI command (as the config has to be loaded when using the CLI): https://github.com/fastly/cli/blob/main/pkg/config/config.go#L325-L335.

Integralist (approved):

PR #1013, opened 2023-09-18
feat(compute): add `metadata` subcommand

Integralist:
This PR introduces a new `fastly compute metadata` command.
It should only be merged _after_ https://github.com/fastly/cli/pull/1016 is merged.

> **NOTE:** No metadata is currently collected.  
> This PR simply introduces the options _first_, then in a later PR we will implement the metadata calls.

Below are some screenshots highlighting how the new `compute metadata` command works...

> **NOTE:** These screenshots were taken back when the command was a top-level `telemetry` command. We've since changed it to be nested as `metadata` under the `compute` command because we're not instrumenting the CLI. We're simply annotating the Wasm binary and so we don't want to mislead users into thinking this is something bigger than it is.

Screenshot 2023-09-18 at 14 47 51

Screenshot 2023-09-18 at 14 48 05

Below is a screenshot highlighting the new `compute metadata` messaging for users who install the CLI for the first time or who upgrade to the CLI version that includes metadata collection...

Monosnap  cts:testing-fastly-cli 2023-09-21 15-48-50


Integralist:
I've converted this PR to draft until we've had time to agree on what the 'default' behaviour is going to be and also the sequencing of events to the rollout of this feature.

pkg/commands/compute/metadata.go
```diff
@@ -0,0 +1,95 @@
...
+		c.Globals.Config.WasmMetadata.MachineInfo = "disable"
+	}
+	if c.disablePackage {
+		c.Globals.Config.WasmMetadata.PackageInfo = "disable"
+	}
+	err := c.Globals.Config.Write(c.Globals.ConfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to persist metadata choices to disk: %w", err)
+	}
+	text.Success(out, "configuration updated (see: `fastly config`)")
```
joeshaw, line 85: IMO it'd be better to print out the values for the 3 fields here rather than pointing them to the toml output.

joeshaw (commented):
LGTM except for the FIXME for the URL.  One suggestion below if you like it.

Integralist:
@joeshaw PR has been amended as per your feedback.

The new output will look something like...

Screenshot 2023-10-02 at 11 52 31

joeshaw:
Thanks. LGTM other than the FIXME.

joeshaw (approved):

pkg/app/run.go
```diff
@@ -113,6 +114,18 @@ func Run(opts RunOpts) error {
 		return nil
 	}
 
+	if !g.Config.CLI.MetadataNoticeDisplayed && commandCollectsData(commandName) {
+		// FIXME: We need the actual URL to point users to.
+		text.Important(g.Output, "The Fastly CLI is configured to collect data related to Wasm builds (e.g. compilation times, resource usage, and other non-identifying data). To learn more about what data is being collected, why, and how to disable it: https://www.fastly.com/")
```
Integralist, line 119: @jaskiratr we'll need an actual URL here (or should this be https://bit.ly/wasm-metadata which points to our community forum thread on the topic)?

jaskiratr: We'll update the [CLI reference page](https://developer.fastly.com/reference/cli/) with the information about what data is being collected, why, and how to disable it

PR #1016, opened 2023-09-20
feat(telemetry): add wasm-tools wasm binary annotations

Integralist:
The intention is for this PR to be merged before https://github.com/fastly/cli/pull/1013. Once this PR is merged (read: _merged_ but _NOT_ released), then the latter PR can be hooked into the metadata collection flows. 

## Context

Ultimately if we did publish both this PR and the latter as part of a new CLI release, they would still not collect any data yet as the annotations (which are part of this PR) are locked down to a feature flag to prevent users from accidentally sending metadata without the proper/appropriate consent.

Also worth noting there is nothing on the Fastly platform side to actually _consume_ this metadata, so nothing would be 'analysed' until appropriate communications had been sent out and users given adequate time to adjust to the upcoming CLI release.

Although the metadata collection is behind a feature flag, this PR does still support disabling _all_ metadata using an environment variable `FASTLY_WASM_METADATA_DISABLE=true`. This is just to be prepared for an eventual CLI release where the metadata collection becomes opt-out and so this will be required for adding to CI environments like the Fastly Compute Actions repo that customers use for installing the CLI and compiling/deploying Compute projects.

There should be a third PR at some point to remove the feature flag, so that the annotations are dependent solely on the user configuring their consent/non-consent via `fastly compute metadata` (which is implemented in the [latter PR](https://github.com/fastly/cli/pull/1013)).

## Screenshots

The following screenshot shows the default metadata on a compiled Wasm binary (i.e. metadata HAS NOT been enabled)...

Screenshot 2023-10-25 at 11 35 59

The following screenshot shows the metadata attached when passing the `--metadata-enable` flag (i.e. metadata collection HAS BEEN opted into)...

Screenshot 2023-10-25 at 13 00 33

> **NOTE:** You'll notice a bunch of secrets have been redacted. We use a combination of a very open-ended regex pattern to err on the side of caution (and this helps catch a large majority of popular formats) + a hardcoded list of env vars that don't fit those formats + any env vars not catered for, the user can add those using `--metadata-filter-envvars` to ensure they get scrubbed also + we use [TruffleHog](https://github.com/trufflesecurity/trufflehog) to identify and redact any secret 'values' it happens to notice.

The following screenshot demonstrates that even if I set `--metadata-enable` no metadata is collected because I've also set `FASTLY_WASM_METADATA_DISABLE=true`, so although it's an unlikely situation, we always err on the side of caution and will give precedence to _disabling_  over the flag (typically flags are given highest priority, but I've opted against that here)...

Screenshot 2023-10-25 at 14 24 57


pkg/commands/compute/language_go.go
```diff
@@ -123,6 +128,50 @@ func (g *Go) Dependencies() map[string]string {
 	return deps
 }
 
+// Imports returns all source code imported packages.
+func (g *Go) Imports() []string {
```
joeshaw, line 132: How precise are you aiming this to be?  This wouldn't take into account unused vendoring, build tags, etc.  A `go list` invocation with the appropriate flags might give a more accurate reading here.  I'd think you'd also need to call it with `scripts.env_vars` set.

Integralist: This is just a _rough_ best effort. I don't think we need to be that specific with the details.

My understanding is that we're just trying to get a feel for popular frameworks that people are using. Which admittedly we'd get from the `Dependencies()` but I wanted to account for the difference between a popular package showing up as a non-direct dependency vs the user actually importing and _using_ that dependency explicitly.

The reason for choosing go.sum over go.mod was to try and account for that scenario where it might be useful to know if there's a popular package underlying a bunch of projects (even if they're not directly used).

pkg/github/github.go
```diff
@@ -343,3 +381,74 @@ func SetBinPerms(bin string) error {
...
+func (m Metadata) URL() string {
+	platform := runtime.GOOS
+	if platform == "darwin" {
+		platform = "macos"
+	}
+
+	arch := runtime.GOARCH
+	if arch == "arm64" {
+		arch = "aarch64"
+	}
```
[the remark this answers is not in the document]
Integralist: Yup, agreed. This was designed to work specifically for wasm-tools and I think somewhere in the code comments I even say something like "once we add more tools this pattern has to be abstracted".

pkg/commands/compute/build.go
```diff
@@ -81,6 +91,12 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return err
 	}
 
+	wasmtools, err := GetWasmTools(spinner, out, c.wasmtoolsVersioner, c.Globals)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(io.Discard, "%#v\n", wasmtools)
```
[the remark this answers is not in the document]
Integralist: Yeah this was an old commit now you've commented on, but at the time I needed to reference `wasmtools` otherwise I couldn't compile the program to run tests but couldn't print to stdout as it would end up _breaking_ said tests 😆 so decided to just discard it.

pkg/env/env.go
```diff
@@ -26,6 +26,20 @@ const (
...
 )
 
+// Parse transforms the local environment data structure into a map type.
+func Parse(environ []string) map[string]string {
+	env := map[string]string{}
+	for _, kv := range environ {
+		toks := strings.SplitN(kv, "=", 2)
+		if len(toks) != 2 {
+			continue
+		}
```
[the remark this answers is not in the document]
Integralist: Ah OK. I didn't write this particular chunk of code I just moved it from one package to another. But like the idea of refactoring it.

pkg/commands/compute/language_go.go
Integralist: But as mentioned, I'll take a look at parsing `go.mod` instead to get both explicit and implicit dependencies.

I don't think there is the same for Rust/JavaScript unfortunately, so those will have to be manually parsed.

joeshaw: > The reason for choosing go.sum over go.mod was to try and account for that scenario where it might be useful to know if there's a popular package underlying a bunch of projects (even if they're not directly used).

`go.sum` will be such a firehose without context that I don't think it'll give you useful data.  The output of `go mod graph` would be better because you'd have the context of what modules depend on other modules.  Combined with the data from `go.mod` you'd be able to filter out what's actually used vs. what's depended upon.

> But as mentioned, I'll take a look at parsing go.mod for Import() instead, to have an easier time parsing out the explicit dependencies rather than using the ast package.

Some confusion here: for `Dependencies()` I think you need to take `go.mod` into account, even if you are reading `go.sum` or running `go mod graph`.  For `Import()` I think what you have is ok, but `go list -f {{.Imports}}` with the appropriate platform settings will be more accurate.

jaskiratr (approved):

pkg/commands/compute/language_go.go
Integralist: **UPDATE:** After some internal discussions, the following will be happening...

1. `Import()` will be removed completely (i.e. no parsing of source code files, and no `CLI_PackageInfoImported%d`).
2. `Dependencies()` for Go I'll going to switch to `go.mod`.
3.  The config.toml `[telemetry]` section will become `[wasm-metadata]`.
4. Environment variable `TelemetryDisable` will be renamed to align with '[wasm-metadata]' naming.
5. The `telemetry` command (from https://github.com/fastly/cli/pull/1013) will be moved to `compute metadata`.

pkg/commands/compute/build.go
```diff
@@ -188,59 +195,68 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
-		fmt.Sprintf("--processed-by=SCRIPTS_POSTINIT=%s", c.Manifest.File.Scripts.PostInit),
-		fmt.Sprintf("--processed-by=SCRIPTS_POSTBUILD=%s", c.Manifest.File.Scripts.PostBuild),
-	}
+	if c.enableTelemetry {
+		endTime := time.Now()
+		runtime.ReadMemStats(&memAfter)
+
+		args := []string{
+			"metadata", "add", "bin/main.wasm",
+			"--language=CLI_" + strings.ToUpper(language.Name),
```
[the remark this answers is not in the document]
Integralist: Done (locally), will push shortly.

pkg/commands/compute/build.go
```diff
@@ -188,59 +195,68 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+			fmt.Sprintf("--processed-by=RUNTIME_OS=%s", runtime.GOOS),
```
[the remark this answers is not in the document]
Integralist: Reworked naming.

Integralist:
@joeshaw made changes, please take another look once you get a chance (also if you could review https://github.com/fastly/cli/pull/1013) as once both those PRs are approved/merged I can then open a 3rd PR to tie them together.

pkg/commands/compute/language_go.go
```diff
@@ -106,15 +106,51 @@ func (g *Go) DefaultBuildScript() bool {
...
+			// length greater than three segments.
+
+			if len(parts) >= 2 && parts[0] == "require" && parts[1] == "(" {
+				insideRequireBlock = true
+				continue
+			}
+			if len(parts) == 1 && parts[0] == ")" {
+				insideRequireBlock = false
+				continue
+			}
```
joeshaw, line 139: I won't die on this hill but I'll make one final pitch for `golang.org/x/mod/modfile`.

```go
data, err := os.ReadFile("go.mod")
if err != nil {
	return
}

f, err := modfile.ParseLax("go.mod", data, nil)
if err != nil {
	return
}

for _, req := range f.Require {
	if req.Indirect {
		continue
	}

	deps[req.Mod.Path] = req.Mod.Version
}
```

pkg/commands/compute/build.go
```diff
@@ -168,13 +195,98 @@ func (c *BuildCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+			text.Break(out)
+
+			fmt.Fprintln(out, string(wasmtoolsOutput))
+			text.Break(out)
+		}
+	} else {
+		// NOTE: If not collecting metadata, just record the CLI version.
+		args := []string{
+			"metadata", "add", "bin/main.wasm",
+			fmt.Sprintf("--language=CLI_%s: %s", revision.AppVersion, strings.ToUpper(language.Name)),
```
joeshaw, line 282: This feels like an odd way to encode the CLI version.  When I run `wasm-tools metadata show main.wasm` for a Rust program I get:

```
module:
    language:
        Rust
    processed-by:
        clang: 15.0.6
        rustc: 1.71.0 (8ede3aae2 2023-07-12)
```

with the code here I think that would appear as:

```
module:
    language:
        Rust
        CLI_10.4.0: Rust
    processed-by:
        clang: 15.0.6
        rustc: 1.71.0 (8ede3aae2 2023-07-12)
```

whereas something like this makes more sense IMO:

```
module:
    language:
        Rust
    processed-by:
        clang: 15.0.6
        rustc: 1.71.0 (8ede3aae2 2023-07-12)
        fastly: 10.4.0 (Rust)
```

Looking at `wasm-tools metadata add --help`:

```
      --language <NAME>
          Add a programming language to the producers section

      --processed-by <NAME=VERSION>
          Add a tool and its version to the producers section

      --sdk <NAME=VERSION>
          Add an SDK and its version to the producers section
```

A few things stand out to me.  `language` appears to just freeform text and not key-value like `processed-by` and `sdk`.

There doesn't appear to be a generic metadata section, so stuffing telemetry into `processed-by` feels a little weird to me, but without any better option it seems ok.  I'd consider writing it as JSON though, so that you're only using one key to do it and it's easily removed.

pkg/commands/compute/language_go.go
Integralist: Oh! Sorry, I think the use of `golang.org/x/mod/modfile` got lost in the shuffle as that definitely is a better solution using an existing `x` package! 

I'll update the code, thanks for triple-checking I saw this 👍🏻

Integralist: Done.

pkg/commands/compute/build.go
Integralist: Done.

Integralist:
@joeshaw I've made amendments as per your feedback.

The output now looks something like...

Screenshot 2023-10-02 at 11 46 06

joeshaw:
The value of the memory allocated doesn't seem right to me, like maybe it underflowed the uint64 there.  I think maybe a signed value there, but it warrants investigation.

Integralist:
@joeshaw You're right. I didn't take into account that the memory allocated could be _less_ than when it started 🤦🏻 

I've modified it to be a snapshot of the memory directly after the build completes.

Let me know if after this change (and https://github.com/fastly/cli/pull/1016/commits/0e71aa13885bbddbc2a4b5130d234b95e369cb59) if this is OK to merge.

Thanks!

joeshaw (approved):
👍

Issue #1020, opened 2023-09-22
[fastly compute build] scripts.build ignored with rust lang


fpalmier:
**Version**

Fastly CLI version v10.4.0 (4b869ab3)
Built with go version go1.21.0 linux/amd64
Viceroy version: viceroy 0.5.1

**What happened**

I'm trying to use a cargo workspace for my Compute@Edge project.

`fastly compute build` fail because of an invalid cargo command

Here is my Cargo.toml
```
[workspace]
resolver = "2"
members = [
    "handler",
    "core",
]
```

Here is my fastly.toml
```
authors = []
description = "todo"
language = "rust"
manifest_version = 3
name = "handler"
service_id = ""

[scripts]
  build = "cargo build --bin handler --release --target wasm32-wasi --color always"

```

Here are the logs
```
Fastly API token provided via config file (profile: user)
Fastly API endpoint: https://api.fastly.com

✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain

INFO: Creating ./bin directory (for Wasm binary)

Command output for 'cargo locate-project --quiet': {"root":"/Users/francois/fastly-rust/Cargo.toml"}

INFO: The Fastly CLI requires a Rust version '>= 1.56.1'.

Build script to execute:
        sh -c cargo build --bin  --release --target wasm32-wasi --color always


INFO: Command output:
--------------------------------------------------------------------------------
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/francois/fastly-rust/handler/Cargo.toml
workspace: /Users/francois/fastly-rust/Cargo.toml
error: "--bin" takes one argument.
Available binaries:
    handler

--------------------------------------------------------------------------------

✗ Running [scripts.build]

ERROR: error during execution process (see 'command output' above): exit status 101.

```

It seem that `modifyCargoPackageName` in `/pkg/commands/compute/language_rust.go` overwrite the build script in fastly.toml with the default script.

Moreover, `m.Package.Name` is blank when using a cargo workspace, I think it should fallback to the first binarie

Integralist:
Thanks for opening this issue @fpalmier 

I've created a ticket internally to track this and will update the ticket once we have more information or have a resolution.

Integralist:
Hi @fpalmier 

Thanks for your patience. I've been off ill the past few days so I've only just managed to circle back around to this issue. 

I'm just trying at the moment to reproduce the error you've reported but I'm not having any luck in doing so. 

[...]

PR #1023, opened 2023-09-29
feat(compute/build): support Cargo Workspaces

Integralist:
Fixes https://github.com/fastly/cli/issues/1020

## Problem

The logic flow prior to this PR would cause a user's custom build script to be replaced with a default build script if the package name they set in their Cargo.toml didn't match the 'default' package name of `fastly-compute-project`. 

This issue was further exacerbated when a customer used a [Cargo Workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) to structure their project. As part of a reduced test case I discovered a bug that caused the user's custom build script to be unexpectedly modified and also changed it in such a way that it would break the build step. 

## Solution

This PR modifies the logic flow for a Rust project. 

The updated flow looks like this...

- If using a standard project structure, then the root Cargo.toml won't be a Cargo Workspace, and so it will contain a package name. The CLI will assign that name to `r.packageName` and later on it will use that name to look up the compiled binary and move it to the required `bin/` directory.

- If the user has set up a Cargo Workspace but there is no custom build script, then the CLI needs to identify which Workspace package is their application. It does this by checking if the package has a rust-toolchain.toml containing the target `wasm32-wasi` and if so we extract the package name from that package's relevant Cargo.toml manifest. The CLI will assign that name to `r.packageName` and later on it will use that name to look up the compiled binary and move it to the required `bin/` directory.

  > **NOTE:** This is sufficient for now, but I can see us needing to update this logic in the future when we start supporting the linking of multiple Wasm binaries on the Fastly platform (aka Wasm Components). Because at that point, a Cargo Workspace could well contain multiple 'binary' projects, each producing a Wasm binary that are linked up in some fashion.

- If the user has set up a Cargo Workspace _and_ they define a custom build script, then the CLI will trust their custom script aligns with the relevant Workspace package name (i.e. their custom script is set up to reference the right binary). The CLI will parse the package name specified in their custom script by looking at the value assigned to the `--bin` flag. The CLI will assign that name to `r.packageName` and later on it will use that name to look up the compiled binary and move it to the required `bin/` directory.

  > **NOTE:** The implementation is very basic and presumes the `--bin` value is inserted directly and isn't required to be resolved (e.g. `$some_environment_variable`).

- If there is no custom build script (i.e. the CLI will use a pre-defined default build script), then the CLI will check to see if the package name (assigned to `r.packageName`) matches the default package name `fastly-compute-project`. If not, then it will update the default build script to use whatever package name is assigned to `r.packageName`.

## NOTES

I've tested this PR with different project set ups to validate the new logic flows.

Below is an example Cargo Workspace project I tested this with...

```
.
├── Cargo.lock
├── Cargo.toml
├── bin
│   └── main.wasm
├── compute
│   ├── Cargo.lock
│   ├── Cargo.toml
│   ├── README.md
│   ├── rust-toolchain.toml
│   └── src
│       ├── main.rs
│       └── welcome-to-compute@edge.html
├── fastly.toml
└── pkg
    └── compute.tar.gz

5 directories, 11 files
```

Below is the root workspace `Cargo.toml`...

```toml
[workspace]

members = ["compute"] # << matches compute/ directory name
```

Below is the `compute/Cargo.toml`...

```toml
[package]
name = "my-compute-app"
version = "0.1.0"
authors = []
edition = "2018"
# Remove this line if you want to be able to publish this crate as open source on crates.io.
# Otherwise, `publish = false` prevents an accidental `cargo publish` from revealing private source.
publish = false

[profile.release]
debug = 1

[dependencies]
fastly = "0.9.4"
```

Below is the fastly.toml file I used with the Cargo Workspace (notice it was moved to the root directory, i.e. I moved it out from the `compute/` directory which contains my Compute application)...

```toml
authors = ["integralist@fastly.com"]
description = ""
language = "rust"
manifest_version = 3
name = "compute"
service_id = ""

[scripts]
  build = "cargo build --bin my-compute-app --release --target wasm32-wasi --color always -vv"
```

### DIFF
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -142,7 +142,7 @@ func (r *Rust) Build() error {
 	return bt.Build()
 }
 
-// RustToolchain models a [toolchain] from a rust-toolchain.toml manifest.
+// RustToolchainManifest models a [toolchain] from a rust-toolchain.toml manifest.
 type RustToolchainManifest struct {
 	Toolchain RustToolchain `toml:"toolchain"`
 }
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -89,6 +89,8 @@ type Rust struct {
 	nonInteractive bool
 	// output is the users terminal stdout stream
 	output io.Writer
+	// packageName is the resolved package name from the project Cargo.toml
+	packageName string
 	// postBuild is a custom script executed after the build but before the Wasm
 	// binary is added to the .tar.gz archive.
 	postBuild string
@@ -110,7 +112,7 @@ func (r *Rust) Build() error {
 		noBuildScript = true
 	}
 
-	err := r.modifyCargoPackageName()
+	err := r.modifyCargoPackageName(noBuildScript)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/commands/compute/language_rust.go
```diff
@@ -140,10 +142,20 @@ func (r *Rust) Build() error {
 	return bt.Build()
 }
 
+// RustToolchain models a [toolchain] from a rust-toolchain.toml manifest.
```
doramatadora, line 145:
```suggestion
// RustToolchainManifest models a [toolchain] from a rust-toolchain.toml manifest.
```

doramatadora (approved):

PR #1032, opened 2023-10-06
refactor(compute/deploy): simplify logic flows

Integralist:
**REVIEWER NOTE:** Probably easier to avoid the diffs and to just read the actual code files.

I've both manually and programmatically validated the code still works (see screenshot for a single example):

Screenshot 2023-10-06 at 14 38 49


fgsch (approved):
I've been poking around but it's a big PR.  Let's merge it now and do more testing in tree.

PR #1033, opened 2023-10-09
Update after parameter name change in Viceroy

fgsch:
Sync with Viceroy 0.9.0.


pkg/commands/compute/serve.go
```diff
@@ -568,7 +568,7 @@ func local(opts localOpts) error {
 		if opts.profileGuestDir.WasSet {
 			directory = opts.profileGuestDir.Value
 		}
-		args = append(args, "--profile-guest="+directory)
+		args = append(args, "--profile=guest,"+directory)
```
Integralist, line 571: Will this cause a break for users with an older Viceroy version?

Integralist: Ah yes, https://github.com/fastly/cli/pull/1019 was merged but hasn't actually been published in a release yet. 

So this PR doesn't break anything because the thing it changes hasn't yet been released. 

So by the time we cut a new CLI release, most users will have the newer version of Viceroy installed. 

As for users who have an older Viceroy pinned, I think you could update the `--profile-guest` flag description to state that the flag is for `Viceroy 0.9.1+`.

fgsch: Good idea. Thank you. Updated.

Integralist (approved):

PR #1036, opened 2023-10-09
feat(product_enablement): add products command

Integralist:
This PR adds a new `products` command which implements the following API endpoints...
https://developer.fastly.com/reference/api/products/enablement/

### DIFF
--- a/pkg/commands/products/root.go
+++ b/pkg/commands/products/root.go
@@ -109,71 +109,57 @@ func (c *RootCommand) Exec(_ io.Reader, out io.Writer) error {
 		return nil
 	}
 
-	var brotliEnabled, diEnabled, fanoutEnabled, ioEnabled, oiEnabled, wsEnabled bool
+	ps := ProductStatus{}
 
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductBrotliCompression,
 		ServiceID: serviceID,
 	}); err == nil {
-		brotliEnabled = true
+		ps.BrotliCompression = true
 	}
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductDomainInspector,
 		ServiceID: serviceID,
 	}); err == nil {
-		diEnabled = true
+		ps.DomainInspector = true
 	}
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductFanout,
 		ServiceID: serviceID,
 	}); err == nil {
-		fanoutEnabled = true
+		ps.Fanout = true
 	}
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductImageOptimizer,
 		ServiceID: serviceID,
 	}); err == nil {
-		ioEnabled = true
+		ps.ImageOptimizer = true
 	}
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductOriginInspector,
 		ServiceID: serviceID,
 	}); err == nil {
-		oiEnabled = true
+		ps.OriginInspector = true
 	}
 	if _, err = ac.GetProduct(&fastly.ProductEnablementInput{
 		ProductID: fastly.ProductWebSockets,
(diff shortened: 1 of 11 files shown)


pkg/commands/products/root.go
```diff
@@ -0,0 +1,189 @@
...
+	"origin_inspector",
+	"websockets",
+}
+
+// NewRootCommand returns a new command registered in the parent.
+func NewRootCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *RootCommand {
+	var c RootCommand
+	c.Globals = g
+	c.manifest = m
+	c.CmdClause = parent.Command("products", "Enable, disable, and check the enablement status of products on your services")
```
awilliams-fastly, line 43:
```suggestion
	c.CmdClause = parent.Command("products", "Enable, disable, and check the enablement status of products")
```

Not sure this is the best suggestion, but I think that dropping _your services_ would be consistent with other command descriptions.

pkg/commands/products/root.go
```diff
@@ -0,0 +1,189 @@
...
+		ImageOptimizer    bool `json:"image_optimizer"`
+		OriginInspector   bool `json:"origin_inspector"`
+		WebSockets        bool `json:"websockets"`
+	}{
+		BrotliCompression: brotliEnabled,
+		DomainInspector:   brotliEnabled,
+		Fanout:            brotliEnabled,
+		ImageOptimizer:    brotliEnabled,
+		OriginInspector:   brotliEnabled,
+		WebSockets:        brotliEnabled,
```
awilliams-fastly, line 155: Replace `brotliEnabled` with corresponding variables.

pkg/commands/products/root.go
```diff
@@ -0,0 +1,189 @@
...
+			ProductID: identifyProduct(c.disableProduct),
+			ServiceID: serviceID,
+		}); err != nil {
+			return fmt.Errorf("failed to disable product '%s': %w", c.disableProduct, err)
+		}
+		text.Success(out, "Successfully disabled product '%s'", c.disableProduct)
+		return nil
+	}
+
+	var brotliEnabled, diEnabled, fanoutEnabled, ioEnabled, oiEnabled, wsEnabled bool
```
awilliams-fastly, line 103: nitpick: These variables could be replaced by an inline struct. It could even be the same struct as used to generate the JSON output.

pkg/commands/products/root.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	serviceID, _, _, err := cmd.ServiceID(c.serviceName, c.manifest, c.Globals.APIClient, c.Globals.ErrLog)
+	if err != nil {
+		return fmt.Errorf("failed to identify Service ID: %w", err)
+	}
+
+	if c.enableProduct != "" {
+		if _, err := ac.EnableProduct(&fastly.ProductEnablementInput{
+			ProductID: identifyProduct(c.enableProduct),
```
awilliams-fastly, line 83: Wondering if we should check for `fastly.ProductUndefined` before making the API call?

pkg/commands/products/root.go
Integralist: 🤦🏻

pkg/commands/products/root.go
Integralist: We already use the `EnumVar` so we know we're going to be getting a value that corresponds to the switch statement. I guess checking for `ProductUndefined` would only be useful if we're concerned the switch statement had a typo (which is completely possible, so I'll look at adding a check 👍🏻 )

awilliams-fastly: Got it -- my concern was avoiding unnecessary API calls. But if that's handled within the client, then it's not a concern.

Integralist: No worries. I've added it because the concern actually comes from us translating the enum flag's value (e.g. "brotli_compression") into an actual Go iota/enum, and at that point we could well have a typo hardcoded in the code which no one notices until runtime 😬 so I think you're right and it's best to throw that check in just to be on the safe side.

awilliams-fastly (approved):

pkg/commands/products/root.go
Integralist: Done. Looks much better thanks.

Issue #1004, opened 2023-08-31
[FEATURE REQUEST] Ability to use files for [setup.kv_stores] section


JakeChampion:
**Is your feature request related to a problem? Please describe.**
I would like to be able to have a kv store setup section which will use the contents of files for the kv store entries.

**Describe the solution you'd like**
I'd like the solution to match the solution for [local kv stores](https://developer.fastly.com/reference/compute/fastly-toml/#kv-and-secret-stores)
```toml
    [[local_server.kv_stores.example_store]]
      key = "second"
      file = "../test/data/some-object.txt"
```

**Describe alternatives you've considered**
I haven't considered an alternative

Integralist:
Thanks for opening this issue @JakeChampion 

This is something I believe @dennismartensson was going to open an internal ticket for.

PR #1040, opened 2023-10-11
fix(compute/deploy): support file field for kv_store setup

Integralist:
Fixes https://github.com/fastly/cli/issues/1004

### DIFF
--- a/pkg/commands/compute/setup/kv_store.go
+++ b/pkg/commands/compute/setup/kv_store.go
@@ -3,6 +3,8 @@ package setup
 import (
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 
 	"github.com/fastly/go-fastly/v8/fastly"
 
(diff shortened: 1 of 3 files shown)


JakeChampion (approved):

pkg/commands/compute/setup/kv_store.go
```diff
@@ -60,11 +63,16 @@ func (o *KVStores) Configure() error {
...
+			promptMessage := "Value"
 			dv := "example"
 			if item.Value != "" {
 				dv = item.Value
 			}
-			prompt := text.BoldYellow(fmt.Sprintf("Value: [%s] ", dv))
+			if item.File != "" {
+				promptMessage = "File"
+				dv = item.File
+			}
```
grantstephens, line 74: What happens when somebody has value and file? Should it not be one or the other?

pkg/commands/compute/setup/kv_store.go
```diff
@@ -149,11 +181,16 @@ func (o *KVStores) Create() error {
...
-					Value: item.Value,
-				})
+				input := &fastly.InsertKVStoreKeyInput{
+					ID:  store.ID,
+					Key: item.Key,
+				}
+				if item.Body != nil {
+					input.Body = item.Body
+				} else {
+					input.Value = item.Value
```
grantstephens, line 191: I didn't look too closely, but is it possible to get here without Value having been set?

grantstephens (approved):
Two small queries, but otherwise LGTM

pkg/commands/compute/setup/kv_store.go
Integralist: I'll fix this by checking both fields and returning an error if both are set.

pkg/commands/compute/setup/kv_store.go
Integralist: So it's hard to see in the diff, but we are actually in a different method here (this is `Create()`). In the earlier `Configure()` method if the user doesn't provide a value for 'value' field, then we set a default of `"example"`.

grantstephens (approved):

Issue #770, opened 2023-01-20
[FEATURE REQUEST] Allow the canonical object-store named to be decided during setup


JakeChampion:
**Is your feature request related to a problem? Please describe.**
When deploying the same object-store starter-kit twice, it currently fails due to both trying to create a store of the same name.

```
✗ Creating object store 'my-store'...

ERROR: the Fastly API returned 409 Conflict: Conflict (store with name 'my-store' already exists).
```

**Describe the solution you'd like**
I'd like an interactive option which notices the store already exists and then prompts the user with a few questions such as:
>> An Object Store named `xxyyzz` already exists. Please choose from below options:
> [1] Use the already existing Object Store
> [2] Create a Object Store with a different name
>
>> 1
>Object-Store: [xxyyzz] xxyyzz-jake

**Describe alternatives you've considered**
I don't have an alternative unfortunately

**Additional context**

PR #1041, opened 2023-10-11
feat(compute/deploy): avoid store conflicts

Integralist:
Fixes https://github.com/fastly/cli/issues/770


pkg/commands/compute/setup/config_store.go
```diff
@@ -113,91 +147,73 @@ func (o *ConfigStores) Create() error {
...
+			}
+		}
 
-				o.Spinner.StopMessage(msg)
-				err = o.Spinner.Stop()
+		if len(configStore.Items) > 0 {
+			for _, item := range configStore.Items {
+				err = o.Spinner.Process(fmt.Sprintf("Creating config store item '%s'", item.Key), func(_ *text.SpinnerWrapper) error {
+					_, err = o.APIClient.UpdateConfigStoreItem(&fastly.UpdateConfigStoreItemInput{
+						Upsert:  true, // Use upsert to avoid conflicts when reusing a starter kit.
```
grantstephens, line 188: Is this documented somewhere- could cause some confusion if data keeps being overwritten.

pkg/commands/compute/setup/kv_store.go
```diff
@@ -97,15 +151,14 @@ func (o *KVStores) Configure() error {
...
-				text.Break(o.Stdout)
 			}
 			if value == "" {
 				value = dv
 			}
 
 			var f *os.File
 			if item.File != "" {
-				abs, err := filepath.Abs(item.File)
+				abs, err := filepath.Abs(value)
```
grantstephens, line 161: Why did this change?

grantstephens (commented):
Couple of small comments, otherwise LGTM

pkg/commands/compute/init.go
```diff
@@ -73,8 +73,7 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		introContext = " (using --from to locate package template)"
 	}
 
-	text.Output(out, "Creating a new Compute@Edge project%s.", introContext)
-	text.Break(out)
+	text.Output(out, "Creating a new Compute@Edge project%s.\n\n", introContext)
```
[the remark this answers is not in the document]
Integralist: Separate PR 🙂 https://github.com/fastly/cli/pull/1044

pkg/commands/compute/setup/config_store.go
```diff
@@ -96,8 +128,10 @@ func (o *ConfigStores) Configure() error {
 		}
 
 		o.required = append(o.required, ConfigStore{
-			Name:  name,
-			Items: items,
+			Name:              name,
+			Items:             items,
+			LinkExistingStore: linkExistingStore,
+			ExistingStoreID:   existingStoreID,
```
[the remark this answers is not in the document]
Integralist: Correct. That's expected. We only try and use `ExistingStoreID` if `LinkExistingStore` is set to `true`.

pkg/commands/compute/setup/config_store.go
Integralist: Hmm, good point actually. I might stick in a warning notice to let the user know we'll be upserting.

pkg/commands/compute/setup/config_store.go
```diff
@@ -48,13 +50,45 @@ type ConfigStoreItem struct {
...
+		)
+
+		for _, store := range existingStores {
+			if store.Name == name {
+				if o.AcceptDefaults || o.NonInteractive {
+					linkExistingStore = true
+					existingStoreID = store.ID
+				} else {
+					text.Warning(o.Stdout, "\nA Config Store called '%s' already exists\n\n", name)
+					prompt := text.Prompt("Use a different store name (or leave empty to use the existing store): ")
```
grantstephens, line 71: Is it worth saying here that data will be upserted?

grantstephens (approved):
Small, thing, but happy to defer to your judgement. LGTM

pkg/commands/compute/setup/config_store.go
Integralist: That's a fair point actually, as it's only relevant when dealing with an existing store 👍🏻

PR #1043, opened 2023-10-13
feat(compute/build): allow user to specify project directory to build

Integralist:
Customers have often asked for the ability to specify a project directory when building and deploying compute applications.

## compute build

Screenshot 2023-10-13 at 16 48 54

## compute deploy

Screenshot 2023-10-13 at 16 40 41

## compute serve (i.e. build, then serve)

Screenshot 2023-10-13 at 16 49 11

## compute publish (i.e. build, then deploy)

Screenshot 2023-10-13 at 16 46 32

### DIFF
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -276,12 +276,7 @@ func Content(flagval string) string {
 	content := flagval
 	if path, err := filepath.Abs(flagval); err == nil {
 		if _, err := os.Stat(path); err == nil {
-			// gosec flagged this:
-			// G304 (CWE-22): Potential file inclusion via variable
-			//
-			// Disabling as we require a user to configure their own environment.
-			/* #nosec */
-			if data, err := os.ReadFile(path); err == nil {
+			if data, err := os.ReadFile(path); err == nil /* #nosec */ {
 				content = string(data)
 			}
 		}
--- a/pkg/commands/compute/build.go
+++ b/pkg/commands/compute/build.go
@@ -56,7 +56,7 @@ func NewBuildCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *Bu
 
 	// NOTE: when updating these flags, be sure to update the composite commands:
 	// `compute publish` and `compute serve`.
-	c.CmdClause.Flag("dir", "Project directory to build (default: current directory)").StringVar(&c.Flags.Dir)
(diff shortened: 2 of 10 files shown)


pkg/cmd/flags.go
```diff
@@ -276,12 +276,7 @@ func Content(flagval string) string {
 	content := flagval
 	if path, err := filepath.Abs(flagval); err == nil {
 		if _, err := os.Stat(path); err == nil {
-			// gosec flagged this:
-			// G304 (CWE-22): Potential file inclusion via variable
-			//
-			// Disabling as we require a user to configure their own environment.
-			/* #nosec */
-			if data, err := os.ReadFile(path); err == nil {
+			if data, err := os.ReadFile(path); err == nil /* #nosec */ {
```
Integralist, line 279: **NOTE:** Latest gosec (released today) changed the way `#nosec` is parsed. It has to be _on_ the affected line.

pkg/commands/compute/build.go
```diff
@@ -50,11 +51,12 @@ type BuildCommand struct {
 func NewBuildCommand(parent cmd.Registerer, g *global.Data, m manifest.Data) *BuildCommand {
 	var c BuildCommand
 	c.Globals = g
-	c.Manifest = m
+	c.Manifest = m // TODO: Stop passing a non-mutable 'copy' in any commands.
 	c.CmdClause = parent.Command("build", "Build a Compute@Edge package locally")
 
 	// NOTE: when updating these flags, be sure to update the composite commands:
 	// `compute publish` and `compute serve`.
+	c.CmdClause.Flag("dir", "Project directory to build (default: current directory)").StringVar(&c.Flags.Dir)
```
awilliams-fastly, line 59: Other projects, such as git and Go, use `-C` when changing the relative directory. Wondering if we should/could include that as an alias for these flags?

awilliams-fastly (approved):

pkg/commands/compute/build.go
Integralist: Oh nice, didn't know `-C` was a convention 👍🏻

JakeChampion: `Make` also uses -C for this 👍 

```
-C dir, --directory=dir
            Change to directory dir before reading the makefiles or doing
            anything else.  If multiple -C options are specified, each is
            interpreted relative to the previous one: -C / -C etc is
            equivalent to -C /etc.  This is typically used with recursive
            invocations of make.
```

PR #1044, opened 2023-10-13
branding: Replace all Compute@Edge/C@E references with Compute.

kpfleming:


kpfleming:
I intentionally left the contents of CHANGELOG.md alone since that is historical information, and I also did not modify the binary files in the `testdata` directories as I wasn't sure what would be needed to ensure that tests would still pass.

Integralist:
The CI is failing because gosec released a new version today that no longer respects the `#nosec` directive. It's a simple fix so I can sort that out on Monday.

pkg/commands/logtail/root.go
```diff
@@ -480,7 +480,7 @@ type (
 		stream string
 	}
 
-	// Log defines the message envelope that compute@edge (C@E) wraps the
+	// Log defines the message envelope that Compute (Compute) wraps the
```
Integralist, line 483:
```suggestion
	// Log defines the message envelope that the Compute platform wraps the
```

Integralist (approved):
LGTM. Just one suggested edit.

PR #1046, opened 2023-10-16
feat: support `--env` flag

Integralist:
We support `--env` in `compute serve`, which acts as a switch for the fastly.toml manifest that is loaded.

For example, `fastly compute serve --env stage` will attempt to use a `fastly.stage.toml` manifest file.

Customers have requested the same support for `--env` with the `compute deploy` command.

This PR modifies the following `compute` commands because it made sense to add support for `--env` to them...

- `compute build`
- `compute deploy`
- `compute publish`
- `compute hashsum`
- `compute hash-files`

I've manually tested these commands to ensure they're working as intended.

**NOTE:** The other reason for modifying `hashsum` and `hash-files` is that I realised if we support modifying a build to use `--env` then it means we need to be able to reproduce that build for the hashing commands.


pkg/commands/compute/serve.go
```diff
@@ -456,40 +462,27 @@ func installViceroy(
 		// Viceroy is already installed, so we check if the installed version matches the latest.
 		// But we'll skip that check if the TTL for the Viceroy LastChecked hasn't expired.
 
-		stale := g.Config.Viceroy.LastChecked != "" && g.Config.Viceroy.LatestVersion != "" && check.Stale(g.Config.Viceroy.LastChecked, g.Config.Viceroy.TTL)
+		stale := check.Stale(g.Config.Viceroy.LastChecked, g.Config.Viceroy.TTL)
```
grantstephens, line 465: Just confirming that the latest version no longer matters here?

grantstephens: Oh, is it because it is being checked below?

pkg/commands/compute/validate.go
```diff
@@ -49,6 +49,14 @@ func (c *ValidateCommand) Exec(_ io.Reader, out io.Writer) error {
 		return fmt.Errorf("error reading file path: %w", err)
 	}
 
+	manifestFilename := manifest.Filename
+	if c.env != "" {
+		manifestFilename = fmt.Sprintf("fastly.%s.toml", c.env)
+		if c.Globals.Verbose() {
+			text.Info(out, "Using the '%s' environment manifest (it will be packaged up as %s)\n\n", manifestFilename, manifest.Filename)
```
grantstephens, line 56: Should we print this outside of the c.env if statement?

pkg/commands/compute/validate.go
```diff
@@ -77,8 +85,8 @@ type ValidateCommand struct {
 // NOTE: This function is also called by the `deploy` command.
 func validatePackageContent(pkgPath string) error {
 	files := map[string]bool{
-		"fastly.toml": false,
-		"main.wasm":   false,
+		manifest.Filename: false,
```
grantstephens, line 88: Think I'm missing something, but how does manifest.Filename get set- it is a global variable?

grantstephens (commented):
Two small comments. One other thing: Should we always print the `env` if it is set just to make it very clear that it's set and we're reading it?

pkg/commands/compute/serve.go
Integralist: So this was me trying to address a bug I had noticed where my config (for whatever reason, I don't know why or when it happened) had ended up with empty string values for both `LastChecked` and `LatestVersion` and because of the conditional check we had, which was...

```go
g.Config.Viceroy.LastChecked != "" && g.Config.Viceroy.LatestVersion != "" && check.Stale(g.Config.Viceroy.LastChecked, g.Config.Viceroy.TTL)
```

It meant I was never checking if the TTL was stale (and so I was left with a message telling me the TTL hadn't expired when I knew it had).

Now, in this PR, if we just pass the empty value to `check.Stale()` then it's going to calculate `LastChecked` as `time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)` and the TTL will definitely have passed by that and so it will force a recheck for a newer Viceroy version.

pkg/commands/compute/validate.go
Integralist: We _could_ do. The feedback I've typically had from users has been the CLI is too noisy.

But on the flip-side I've also had users say things like "why do I have to use --verbose to see certain information".

It's all a balance and we won't be able to keep everyone happy unfortunately.

**That said, in this particular case my thinking was along these lines...**

Setting the `--env` flag is something the user knows they've done, because they've explicitly set it when calling the CLI command (unless they're using some kind of custom abstraction around the CLI and someone has added it as part of that custom abstraction). So if the user has had to explicitly type `--env` as part of the command invocation, then it should be reasonable to presume they know what they've set and why, and that's why I decided, in this case, to keep the extra INFO message behind the `--verbose` flag.

pkg/commands/compute/validate.go
Integralist: Correct! It's a global constant https://github.com/fastly/cli/blob/main/pkg/manifest/manifest.go#L9

grantstephens (approved):

PR #1081, opened 2023-11-07
refactor(compute/serve): replace log.Fatal usage with channel

Integralist:


pkg/commands/compute/serve.go
```diff
@@ -648,6 +648,25 @@ func local(opts localOpts) error {
...
+	// asynchronously in a goroutine. We also need to be able to signal the
+	// viceroy process to be killed, and we do that using `s.Signal(os.Kill)` from
+	// within the relevant error handling blocks in `watchFiles`, where upon the
+	// below `select` statement will pull the error message from the `failure`
+	// channel and return it to the user. If we fail to kill the Viceroy process
+	// then we still want to pull an error from the `failure` channel and so we
+	// have a separate `select` statement to check for any initial errors prior to
+	// the Viceroy executable starting and an error occurring in `watchFiles`.
+	select {
+	case channelErr := <-failure:
```
awilliams-fastly, line 663: What would you think of renaming these 2 variables (here and elsewhere)? Looking only here, it's not clear what a receive on the channel indicates. Maybe something like `fileWatchErrs`? Likewise for `channelErr`, maybe `fileWatchErr`?

Integralist: Ah so I actually purposefully chose generic naming because although it's handling file watch errors right now, maybe in the future I want other types of errors to cause the stopping of the process.

Integralist: I could maybe name it something like `asyncErrs := <-failure`

Integralist: Done

awilliams-fastly (approved):
Similar comment as this https://github.com/fastly/cli/pull/1082#discussion_r1386738551.

Otherwise, LGTM

Integralist:
Thanks @awilliams-fastly I think that makes sense for the other PR but for this one I think there's a wrinkle where we have...

```
eventHandler := func(modifiedFile string, _ fsnotify.Op) { ... } 
```

...which can't return an error in its current form and I can't convert it to return an error because it's contained by another function debounced(func() {}) that we don't control and can't convert to handle an error. So it gets harder to reason about at that point. 

So I think I'll leave it as is for this PR but for I'll definitely look at doing your suggestion for the other PR.

PR #1082, opened 2023-11-08
refactor(logtail): replace log.Fatal usage with channel

Integralist:


pkg/commands/logtail/root.go
```diff
@@ -101,31 +101,39 @@ func (c *RootCommand) Exec(_ io.Reader, out io.Writer) error {
...
+	}
 
 	return nil
 }
 
 // Tail starts the virtual tail process. Tail fetches data from the eventbuffer
 // API. It hands off the requested logs to the outputloop for the actual
 // printing.
-func (c *RootCommand) tail(out io.Writer) {
+func (c *RootCommand) tail(out io.Writer, failure chan error) {
```
awilliams-fastly, line 130: Wonder if this could be `failure chan<- error`?

pkg/commands/logtail/root.go
```diff
@@ -101,31 +101,39 @@ func (c *RootCommand) Exec(_ io.Reader, out io.Writer) error {
...
 
 	// Start the output loop.
 	go c.outputLoop(out)
 
 	// Start tailing the logs.
-	go c.tail(out)
-
-	<-sigs
-	close(c.dieCh)
+	go c.tail(out, failure)
```
awilliams-fastly, line 112: An alternative to passing `failure` into `tail` would be something like
```
go func() {
    failure <- c.tail(out)
}()
```
I bring it up in case you think it'd simplify `c.tail`.

Integralist: Done

awilliams-fastly (approved):

PR #1115, opened 2024-01-11
Add support for passing file path to certificate for certificate create command

rdoherty-fastly:
It is common (and easier) to pass a file path into a CLI tool vs trying to pass in file contents. This change allows the --cert-blob argument to take a file path. It is backwards compatible via the `argparser.Content` function, which determines if the flag value is a file path or file contents.

This follows the existing behavior in `pkg/commands/vcl/snippet/create.go` and a few others. I've tested with a file path and raw cert contents with the `--debug` flag and it does send the correct cert data both ways.

**Questions**:

* Not sure if it's a good idea to overload the `--cert-blob` field, but seems benign?
* It feels weird to use the `argparser.OptionalString` type for the certBlob field that is required, but I couldn't find anything else that might work but could have missed it.
* How should this be tested? I'm new to Go and don't see existing tests for this functionality. Is there anything I can use as a guide?
* Is this considered backwards compatible? I think so, existing tooling would be passing in the full cert contents and keep working, this would allow users to use a file path if they like.

I'm fairly new to Go, any feedback is appreciated!

### DIFF
--- a/pkg/commands/tls/custom/certificate/certificate_test.go
+++ b/pkg/commands/tls/custom/certificate/certificate_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"path/filepath"
 	"testing"
 
 	"github.com/fastly/go-fastly/v8/fastly"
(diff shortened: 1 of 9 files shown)


pkg/commands/tls/custom/certificate/create.go
```diff
@@ -17,7 +17,7 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
 	c.Globals = g
 
 	// Required.
-	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob").Required().StringVar(&c.certBlob)
+	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob. Can be passed as file path or content, e.g. \"$(< certificate.crt\" or --cert-blob=path/to/certificate.crt").Action(c.certBlob.Set).Required().StringVar(&c.certBlob.Value)
```
Integralist, line 20: Slight tweak to the description, but importantly I've removed the `.Value` as I think the type should be a `string` and not an `argparser.OptionalString` (because the flag is marked as required).

```suggestion
	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob. Can be passed as file path (e.g. \"path/to/certificate.crt\") or content (e.g. \"$(< certificate.crt)\")").Action(c.certBlob.Set).Required().StringVar(&c.certBlob)
```

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -30,7 +30,7 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
 type CreateCommand struct {
 	argparser.Base
 
-	certBlob string
+	certBlob argparser.OptionalString
```
Integralist, line 33: I think this should stay as a string as the flag is marked as 'required'
```suggestion
	certBlob string
```

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -59,7 +59,11 @@ func (c *CreateCommand) constructInput() *fastly.CreateCustomTLSCertificateInput
 	if c.id != "" {
 		input.ID = c.id
 	}
-	input.CertBlob = c.certBlob
+
+	if c.certBlob.WasSet {
+		input.CertBlob = *fastly.String(argparser.Content(c.certBlob.Value))
+	}
```
Integralist, line 65: The flag is required so there must be a value provided, so we'll just pass the `string` directly to `argparser.Content()`

```suggestion
	input.CertBlob = *fastly.String(argparser.Content(c.certBlob))
```

Integralist (changes requested):
Thanks @rdoherty-fastly for opening this PR! Much appreciated 🙂 

I've made some small suggested edits for you to review.

In answer to your questions...

- **Not sure if it's a good idea to overload the `--cert-blob` field, but seems benign?**
  - I think this is fine. It's backwards compatible and we've used a similar approach of overloading a flag elsewhere in the CLI.

- **It feels weird to use the `argparser.OptionalString` type for the certBlob field that is required, but I couldn't find anything else that might work but could have missed it.**
  - I made suggested edits to revert it back to the original type of a `string`. I'm not sure why you needed to change the type, but I presume (as you're new to Go) that maybe you simply made a mistake somewhere earlier and it just evolved into using OptionalString. But you're instinct is correct, and we shouldn't use it for a 'required' flag.
  
- **How should this be tested? I'm new to Go and don't see existing tests for this functionality. Is there anything I can use as a guide?**
  - In Go the tests always exist in a file alongside the code file itself, and is always suffixed with `_test.go` as the filename. So you'll see in the same directory https://github.com/fastly/cli/blob/2d6c936cda7d2acdbd7709ec7c67c11758fdf894/pkg/commands/tls/custom/certificate/certificate_test.go 
  
  I would suggest making a copy of the error/success test cases https://github.com/fastly/cli/blob/2d6c936cda7d2acdbd7709ec7c67c11758fdf894/pkg/commands/tls/custom/certificate/certificate_test.go#L34-L54 and modify the copies so that you're passing a relative file path instead of `example` (which in the current tests is standing in for the actual PEM contents, which we mock the API call to succeed). But you can look into adding a test file into a `./testdata/` directory inside of this same directory (e.g. add `pkg/commands/tls/custom/certificate/testdata/certificate.crt`). In Go, you use testdata directories for holding fixture data...
  
  > Test files that declare a package with the suffix "_test" will be compiled as a
separate package, and then linked and run with the main test binary. The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests. -- https://pkg.go.dev/cmd/go/internal/test
  
- **Is this considered backwards compatible? I think so, existing tooling would be passing in the full cert contents and keep working, this would allow users to use a file path if they like.**
  - Yes. I think this is backwards compatible. It shouldn't break for a user who passes a blob instead of a file path.

Thanks!

rdoherty-fastly:
Thanks @Integralist ! Will work on these things today or tmrw!

rdoherty-fastly:
Thanks for the feedback @Integralist! I've updated the PR with your changes.

rdoherty-fastly:
Oops I need to update tests.

rdoherty-fastly:
Tests added @Integralist , would love a review of them, thanks!

pkg/commands/tls/custom/certificate/certificate_test.go
```diff
@@ -52,6 +55,28 @@ func TestTLSCustomCertCreate(t *testing.T) {
...
+		{
+			Name: validateFilePathSuccess,
+			API: mock.API{
+				CreateCustomTLSCertificateFn: func(_ *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+					return &fastly.CustomTLSCertificate{
+						ID: mockResponseID,
+					}, nil
+				},
+			},
+			Args:       args(fmt.Sprintf("tls-custom certificate create --cert-blob %s", filepath.Join("pkg", "config", "testdata", "certificate.crt"))),
```
Integralist, line 77: I'm not sure how this is passing because from what I can tell there isn't a file at `pkg/config/testdata/certificate.crt`.

rdoherty-fastly: Uh yep that is not good. I forgot to commit that file, but it should fail in CI.

Integralist: Yeah let's make sure ci fails first then commit the file to see if it passes. You can of course run the tests locally to help debug things

rdoherty-fastly: It's failing b/c if `argparser.Content` can't find the file it thinks the argument is the content. So the cert-blob being uploaded is `pkg/config/testdata/certificate.crt` . A filename passed in that doesn't exist becomes the raw cert blob contents. Will think about how to handle this.

rdoherty-fastly: https://github.com/fastly/cli/blob/46fefb53dac87aaef3898c9d1094baac4281f26c/pkg/argparser/flags.go#L276

rdoherty-fastly: So the question is, do we do more heuristics or validation on the cert-blob before uploading to check if we're uploading a cert? Or do we do some heuristics on if the cert-blob looks like a file path? 

My concern is if someone gives an invalid path, the client will upload a 'cert' that's invalid. They then get this error:

```
➜  cli git:(rdoherty/certificate-create-accept-file-path) ✗ ./fastly tls-custom certificate create --cert-blob=bad/path

ERROR: the Fastly API returned 400 Bad Request: Invalid value for cert_blob (The contents provided do not appear to be a valid X.509 certificate.).

```

Which isn't too helpful. Only if they use `--debug-mode` would they see that the 'contents' are the file path. We could use `pem.Decode()` and if it fails return an error. Thoughts?

Integralist: OK, I think we need to start handling the error case in `argparser.Content`...

https://github.com/fastly/cli/blob/main/pkg/argparser/flags.go#L281C43-L281C43

...so this ^^ should change to `!= nil` rather than the happy path `== nil`.

e.g.

```go
func Content(flagval string) (string, error) {
	content := flagval
	if path, err := filepath.Abs(flagval); err == nil {
		if _, err := os.Stat(path); err != nil {
			return fmt.Errorf("error looking up provided path '%s': %q", path, err)
		}
		if data, err := os.ReadFile(path); err == nil /* #nosec */ {
			content = string(data)
		}
	}
	return content
}
```

Notice the function signature has changed to returning two arguments now `(string, error)`.

This will mean updating the callers to handle an error now as well. I think it's only called in like 5 places in the codebase though so shouldn't be a massive amount of work.

Well done spotting this bug 👍🏻 🙂

rdoherty-fastly:
I'm closing this PR in favor of opening up a new one that will add a separate flag (`--cert-file`?) that is mutually exclusive to `--cert-blob`. One of the two will be required.

Integralist:
Thanks @rdoherty-fastly

PR #1130, opened 2024-02-14
fix(compute/build): normalise and bucket heap allocations

Integralist:
## Problem

We started annotating Wasm binaries with additional metadata and part of that data was the number of memory heap allocations. Including this metric means every time you build a package (even if you have changed ZERO lines of code) it'll produce a package with a different hash (e.g. `fastly compute hash-files`). This behaviour breaks Terraform's ability to reason about whether the package has changed, as each time you build the amount of memory used to create the package is variable/fluctuating.

## Solution

We're going to bucket the memory allocation numbers to provide a consistent output. If the memory allocations increase significantly then the bucket value will change (as we're considering the change significant enough to be considered some kind of important code/dependency change).

## Screenshots

Below demonstrates me building a package twice (no code changes) and the memory allocation value is now consistent (`2-5MB`):

Screenshot 2024-02-14 at 17 07 53

The following demonstrates me building a package twice (no code changes) and the hash produced is thus the same:

Screenshot 2024-02-14 at 17 23 27

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.21.x
+          go-version: 1.22.x
       # NOTE: Manage GitHub Actions cache via https://github.com/fastly/cli/actions/caches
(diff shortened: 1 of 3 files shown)


Integralist:
Thanks @cee-dub 👍🏻

PR #1134, opened 2024-02-16
build: change go module to include /v10 path

Integralist:
## Problem

Unable to install latest CLI version using `go install`.

For example, the following command will install the latest v1.x.x release and _not_ the actual latest release (i.e. v10.8.2):

```
go install github.com/fastly/cli/cmd/fastly@latest
```

## Solution

Update the go.mod module name to include `/v10` (the current major release) and update all code to reference it.

This should mean that we can install the latest version of v10.x.x with:

```
go install github.com/fastly/cli/v10/cmd/fastly@latest
```

## Notes

This would normally be considered a breaking-change but as the CLI is a binary and not a library package to be consumed by another application, this should be fine to release (i.e. it's highly unlikely any one is importing the `/pkg` packages, although we should move those to `/internal` at some point).

Also, I've removed the `.tmpl` files (and any related scripts/logic). I created them years ago as a helper and never ended up using them in practice, and no one else uses them either and because of that, they've gone _very_ stale (judging by the go-fastly version used, I've not used these templates in almost 2 and half years).


Integralist:
@ydnar @joeshaw @dgryski if any of you have opinions on this PR I'd love to hear them (see the description for details).

My big concerns are, every time we release a new major version:

- We have to bump _a lot_ of files to update the import paths to match the go.mod module name.
- A user has to specify the specific version as part of the `go install` invocation (which sucks).
- And due to prev point it means any documentation for installing the CLI via `go install` needs constant updating. 

I just want to be sure there isn't a better way to solve the 'go install latest' problem.

cee-dub:
For one thing, can't the implementation files (that are obviously v10: that's what they do at the current revision) live in a directory that doesn't move? The main that compiles to the binary v10 may need to be in an actual/v10 directory, but I don't believe all the files need to be.

Integralist:
Thanks for the feedback @cee-dub 

Unfortunately, I don't believe your suggestion will work (at least from how I've interpreted it).

i.e.

> can't the implementation files (that are obviously v10: that's what they do at the current revision) live in a directory that doesn't move?

No, they can't as far as I know. Their import path is coupled to the module name.

I tried doing what I think you've suggested here https://github.com/Integralist/test-go-binary and it didn't work.

In that example repo I have a go module named `github.com/integralist/test-go-binary` with the following tree structure...

```
.
├── README.md
├── cmd
│   └── test-go-binary
│       └── v4
│           └── main.go
├── go.mod
└── internal
    └── flags
        └── flags.go
```

I've tagged the commit with `v4.0.0` but if I try to do the following...

```
$ go install github.com/integralist/test-go-binary@latest
```

...then the v1.x.x will be installed.

If I try to install the tag explicitly (which isn't the goal, the goal is to use `@latest`, but if we try it nonetheless), then we'll get an error...

```
$ go install github.com/integralist/test-go-binary@v4.0.0

go: github.com/integralist/test-go-binary@v4.0.0: 
github.com/integralist/test-go-binary@v4.0.0: 
invalid version: module contains a go.mod file, 
so module path must match major version ("github.com/integralist/test-go-binary/v4")
```

If I was to change the module name to `github.com/integralist/test-go-binary/v4`, then that would require the go install to look like this...

```
go install github.com/integralist/test-go-binary/v4/cmd/test-go-binary@latest
```

...like I've shown with the Fastly CLI example, but I then am also forced to have all packages within the CLI imported as the module name `github.com/integralist/test-go-binary/v4/...` because that's what the module name is.

joeshaw (approved):
Go's module versioning is really terrible, _especially_ when talking about a CLI program as opposed to a library.

The only option here is to change the module path to include the version number, if you ever want `@latest` to work (but even then you still have to put the version number in the module path).  So `go install github.com/fastly/cli@latest` will never work as you'd want it to -- you'll only ever get the latest 1.x version.  It'll have to be `go install github.com/fastly/cli/v10@latest`, and folks will need to somehow know to switch it to `v11`, `v12`, etc. as major version numbers change.  Which they mostly should not have to care about with a CLI program.

So, this is the best of a bad situation if you want folks to be able to `go install` anything. 👍

Integralist:
I've decided this isn't worth the effort.

We know a user can install the v1 easily and can just run `fastly update` to get the latest version.

I _started_ doing this work because I wanted a way to avoid GitHub API rate limits as part of an internal project (due to the number of times the CI pipeline will run) but considering we still need to know the major version in the path, I'm thinking now I might be better off just doing the `fastly update` dance and instead caching the API releases lookup.

ydnar:
Sorry to chime in late here, but here’s my (probably unhelpful) $0.02:

I think in some ways the decision to mangle the import path with `/vMajor` was a subtle disincentive designed by the Go team to discourage breaking changes to published packages. See their blog post on "there will never be a Go 2.0." I think the way that the `go` command and related tooling (`gopls` among them) poorly handle module versions > v1 is evidence of this.

The solution I’ve (regrettably) arrived at is forever `v0.N.M` versions, updating the minor version when introducing breaking changes.

ydnar:
More $0.02, from Slack:

If customers need to do something to move to a breaking `v10` version, it’s worth considering breaking the URL into something we can iterate faster on, e.g. a vanity module URL.

Softball proposal: move the CLI (and other Fastly Go packages) under `go.fastly.com/`. Each can be a separate module, with a separate version as necessary.

Second: breaking the import path gives us a chance to reset the version. Instead of `v10`, we set the version of this module to `v0.10.0` and introduce breaking changes at the minor release. This avoids package path mangling with `vN`.

Concretely: `go get go.fastly.com/cli@latest` would fetch the `go.fastly.com/cli@v0.10.0`.

ydnar:
Third (somewhat related, but not necessary): move the new vanity-URLed Fastly Go modules under a monorepo, e.g. `github.com/fastly/go-modules`. Subdirectories of that repo would be top-level Go modules under `go.fastly.com/`.

grantstephens:
Wouldn't we be able to use the vanity url to point to the latest version, i.e. along the lines of `go.fastly.com/cli` -> `github.fastly.com/cli/v10` ?

ydnar:
> Wouldn't we be able to use the vanity url to point to the latest version, i.e. along the lines of `go.fastly.com/cli` -> `github.fastly.com/cli/v10` ?

We could, but I’m making the argument that Go module versions > 1.0 are harmful.

If we reset the entire module import path, we can reset the version number as well, since it’s an entirely new package/module.

PR #1136, opened 2024-02-20
fix(github): update wasm-tools path

Integralist:
Looks like the latest release of wasm-tools https://github.com/bytecodealliance/wasm-tools/releases/tag/v1.200.0 changed its release assets path.

### DIFF
--- a/Dockerfile-rust
+++ b/Dockerfile-rust
@@ -15,8 +15,6 @@ RUN rustup toolchain install ${RUST_TOOLCHAIN} \
   && useradd -ms /bin/bash fastly
 
 USER fastly
-# this forces the download of the crates index; it will save build time if the image was recently built
-RUN cargo search --limit 0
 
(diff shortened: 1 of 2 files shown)


Dockerfile-rust
```diff
@@ -15,8 +15,6 @@ RUN rustup toolchain install ${RUST_TOOLCHAIN} \
   && useradd -ms /bin/bash fastly
 
 USER fastly
-# this forces the download of the crates index; it will save build time if the image was recently built
-RUN cargo search --limit 0
```
Integralist, line 19: This change is unrelated to the PR itself but was needed as crates.io now returns a `500 Internal Server Error` for this call (I'm able to replicate locally). I've pinged our Rust experts internally about whether something changed on the Crates end of things, but as it's a performance operation and doesn't directly affect functionality I decided the best thing to do is to just remove the line.

PR #1142, opened 2024-03-01
fix(compute/build): avoid persisting old metadata

Integralist:
**Problem:** If you ever compiled a `main.wasm` binary with metadata, and then later compiled with metadata disabled, then the binary would continue to be annotated with the old metadata.
**Solution:** Completely remove the old `main.wasm` binary if it exists before re-compiling.


fgsch (approved):
Thank you!

PR #1169, opened 2024-03-27
fix(stats/historical): avoid runtime SIGSEGV

Integralist:
CLI was triggering a SIGSEGV panic whenever a flag was being passed (like `--from`, regardless of value)...

```
go run cmd/fastly/main.go stats historical --service-id <SOME_ID> --token $FASTLY_API_KEY --from '2024-03-20 12:00:00'
```

The fix was to change how the Kingpin package writes the flag value to memory.


kpfleming (approved):
This looks straightforward to me, if you've verified that it resolves the bug then we're off and running.

PR #1179, opened 2024-04-15
fix(compute/deploy): check compute product entitlement when creating service

Integralist:
A few changes are happening in this PR but the primary change is to **_check the entitlement the user has to the Compute product and to display a message informing them that the new service they're about to create will be part of a trial access_** (unless they are already on a trial, or they've paid for access).

Additional changes worth noting:

- Refactoring the code to avoid passing a function through _multiple_ call sites (just wasn't necessary).
- Removing the use of a function type that used partial application (a result of the above refactor).
- Replacing the "trial activation" logic with a simpler "check entitlement" API call.
- Refactoring the 'deploy' tests so we have a default behaviour for mocking the entitlement call **

> ** The original test code was duplicating the same mocking code everywhere and it just wasn't necessary. Setting a failure mock mode was the only special case that needed explicit overriding.


pkg/commands/compute/deploy.go
```diff
@@ -1211,7 +1165,7 @@ func (c *DeployCommand) ExistingServiceVersion(serviceID string, out io.Writer)
 		})
 		return serviceVersion, fsterr.RemediationError{
 			Inner:       fmt.Errorf("invalid service type: %s", serviceType),
-			Remediation: "Ensure the provided Service ID is associated with a 'Wasm' Fastly Service and not a 'VCL' Fastly service. " + fsterr.ComputeTrialRemediation,
+			Remediation: "Ensure the provided Service ID is associated with a 'Wasm' Fastly Service and not a 'VCL' Fastly service. " + fsterr.ComputeAccessRemediation,
```
kpfleming, line 1168:
```suggestion
			Remediation: "Ensure the provided Service ID is associated with a 'Compute' Fastly Service and not a 'CDN' Fastly service. " + fsterr.ComputeAccessRemediation,
```

pkg/api/undocumented/undocumented.go
```diff
@@ -17,12 +17,30 @@ import (
 	"github.com/fastly/cli/pkg/useragent"
```
Integralist, line 17: @kpfleming feel free to rebase and take over this PR if you want to.

kpfleming:
This is no longer necessary as 'trials' are gone.

PR #1182, opened 2024-04-17
Add support for specifying location when creating KV stores

hrmsk66:
:wave: This PR adds the ability to specify the regional location when creating a new KV store using the Fastly CLI.

Changes made:

- Update go-fastly to 9.3.0
- Add a new `location` flag to the fastly kvstore create command. Allows specifying the region from a list of supported locations (US, EU, ASIA, AUS).
- Add new test cases to validate creating a KV store with a specific location.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.2.2
+	github.com/fastly/go-fastly/v9 v9.3.0
 	github.com/hashicorp/cap v0.6.0
 	github.com/kennygrant/sanitize v1.2.4
 	github.com/mholt/archiver v3.1.1+incompatible
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.3.0
+	github.com/fastly/go-fastly/v9 v9.3.1
 	github.com/hashicorp/cap v0.6.0
 	github.com/kennygrant/sanitize v1.2.4
 	github.com/mholt/archiver v3.1.1+incompatible
--- a/pkg/commands/kvstore/create.go
+++ b/pkg/commands/kvstore/create.go
@@ -19,6 +19,9 @@ type CreateCommand struct {
 	Input fastly.CreateKVStoreInput
 }
 
+// locations is a list of supported regional location options.
+var locations = []string{"US", "EU", "ASIA", "AUS"}
+
 // NewCreateCommand returns a usable command registered under the parent.
 func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
 	c := CreateCommand{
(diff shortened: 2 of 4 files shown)


fgsch:
I will defer to @Integralist but this LGTM 👍

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.2.2
+	github.com/fastly/go-fastly/v9 v9.3.0
```
Integralist, line 30: I've just published `9.3.1` if you wouldn't mind bumping to that please.

pkg/commands/kvstore/create.go
```diff
@@ -28,6 +31,7 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
 	}
 	c.CmdClause = parent.Command("create", "Create a KV Store")
 	c.CmdClause.Flag("name", "Name of KV Store").Short('n').Required().StringVar(&c.Input.Name)
+	c.CmdClause.Flag("location", "Regional location of KV Store").Short('l').HintOptions(locations...).EnumVar(&c.Input.Location, locations...)
```
Integralist, line 34: Please move this to sit above `name` flag so the help output shows flags in sorted order (this is something I've been doing wherever/whenever possible and I'd like it to be as consistent as possible moving forward).

**EDIT:** So with that in mind I see `--json` flag will need reordering too.

pkg/commands/kvstore/kvstore_test.go
```diff
@@ -71,6 +72,37 @@ func TestCreateStoreCommand(t *testing.T) {
...
+					}, nil
+				},
+			},
+			WantOutput: fstfmt.EncodeJSON(&fastly.KVStore{
+				StoreID:   storeID,
+				Name:      storeName,
+				CreatedAt: &now,
+				UpdatedAt: &now,
+			}),
+		},
```
Integralist, line 105: I'm not sure of the value of the tests here other than they verify a `--location` flag can be provided.

It might be worth just adding a NOTE above the tests to indicate they just validate the flag is allowed as no other indicator is present in either the `fastly.KVStore` struct or subsequently the CLI output.

e.g.

```
// NOTE: The following tests only validate support for the --location flag.
// No other location/region indicators are exposed for us to validate.
```

Unless, you decide to change the CLI output to say something like...

```
fstfmt.Success("Created KV Store '%s' (%s, %s)", storeName, storeID, storeLocation)
```

...so we can validate the location.

hrmsk66: Thank you for your suggestion. IMHO, the API response should ideally include the location, and the CLI should use that value for the create command output.
(Furthermore, I think it would be preferable to include the location in the describe command output as well.)

Since the current API response does not include that value, I added a comment to the test as suggested.

I just noticed that the absence of location in the API response has been brought up internally. Hopefully, it will become available soon, and it can be incorporated into the CLI 🤞

PR #1187, opened 2024-04-26
Add ability to init from existing compute service

jsocol:
The '--from' argument to 'fastly compute init' now accepts a Fastly
Service ID, which must refer to an existing compute service. The
fastly.toml manifest will be populated with details from this existing
service.


pkg/app/run.go
```diff
@@ -622,7 +622,7 @@ func commandRequiresAuthServer(command string) bool {
 // requires an API token.
 func commandRequiresToken(command string) bool {
 	switch command {
-	case "compute build", "compute hash-files", "compute init", "compute metadata", "compute serve":
+	case "compute build", "compute hash-files", "compute metadata", "compute serve":
```
Integralist, line 625: This is an interesting scenario because by making `compute init` return `true` (i.e. "it requires a token") this potentially causes issues for users who want to mess around with the Fastly platform locally but who don't have an account or have a token created yet (e.g. they want to try out Fastly and to spin up some application code locally, i.e. `fastly compute serve` without needing to create an account).

I wonder if we could add additional logic to https://github.com/fastly/cli/blob/f3790ccee587fd63f3879f7e8a860091f16d4e29/pkg/app/run.go#L232 so that we check if the `--from` flag was passed and if it's not a Service ID, then we'll skip that conditional branch that tries to validate a token.

pkg/commands/compute/init.go
```diff
@@ -68,14 +70,22 @@ func NewInitCommand(parent argparser.Registerer, g *global.Data) *InitCommand {
 // Exec implements the command interface.
 func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	var introContext string
+	var isExistingService bool
```
Integralist, line 73: NIT: maybe combine into single `var`...

```suggestion
	var (
		introContext string
		isExistingService bool
	)
```

pkg/commands/compute/init.go
```diff
@@ -142,13 +152,18 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
-	name, desc, authors, err := c.PromptOrReturn(email, in, out)
-	if err != nil {
-		c.Globals.ErrLog.AddWithContext(err, map[string]any{
-			"Description": desc,
-			"Directory":   c.dir,
-		})
-		return err
+	var name string
+	var desc string
+	var authors []string
```
Integralist, line 157: NIT: maybe combine into single `var`...

```suggestion
	var (
		name string
		desc string
		authors []string
	)
```

pkg/commands/compute/init.go
```diff
@@ -196,25 +211,104 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+					"Tag":       tag,
+					"Directory": c.dir,
+				})
+				return err
+			}
+		} else {
+			profileName, _, _ := c.Globals.Profile()
+
+			var service *fastly.Service
+			var pack *fastly.Package
```
Integralist, line 243: NIT: maybe combine into single `var`...

```suggestion
			var (
				service *fastly.Service
				pack *fastly.Package
			)
```

pkg/commands/compute/init.go
```diff
@@ -196,25 +211,104 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+				c.Globals.ErrLog.AddWithContext(err, map[string]any{
+					"From":      from,
+					"Branch":    branch,
+					"Tag":       tag,
+					"Directory": c.dir,
+				})
+				return err
+			}
+		} else {
+			profileName, _, _ := c.Globals.Profile()
```
Integralist, line 240: This presumes a profile is being used.

For example, I have profiles defined but I typically have them overridden by the `FASTLY_API_TOKEN` environment variable. See the following logic flow for more details: https://github.com/fastly/cli/blob/6c1ec982475e698bd347c94622f857d9e4f44c90/pkg/global/global.go#L125-L133

...basically it breaks down to...

- `--token` returns `lookup.SourceFlag`
- `FASTLY_API_TOKEN` returns `lookup.SourceEnvironment`
- Everything else (i.e. anything to do with a profile) returns `lookup.SourceFile`
- If no token found it returns `lookup.SourceUndefined`

pkg/commands/compute/init.go
```diff
@@ -196,25 +211,104 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+						"Directory": c.dir,
+					})
+					if hErr, ok := err.(*fastly.HTTPError); ok && hErr.IsNotFound() {
+						return fmt.Errorf("the service %s could not be found for profile %s", c.cloneFrom, profileName)
+					}
+					return err
+				}
+
+				if service.Type == nil || *service.Type != "wasm" {
+					return fmt.Errorf("service %s is not a compute service (type is %s)", c.cloneFrom, *service.Type)
```
Integralist, line 260: I think this needs to be capitalized as it's a product name (best to check throughout the PR in case any other references need updating)...

```suggestion
					return fmt.Errorf("service %s is not a Compute service (type is %s)", c.cloneFrom, *service.Type)
```

pkg/commands/compute/init.go
```diff
@@ -196,25 +211,104 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+				}
+
+				if service.ActiveVersion == nil {
+					return fmt.Errorf("service %s does not have an active version to clone", c.cloneFrom)
+				}
```
Integralist, line 265: We could return a remediation error message that lets the user know they can pass the autoclone flag:
https://github.com/fastly/cli/blob/main/pkg/errors/remediation_error.go#L106-L109

e.g.

```go
return fsterr.RemediationError{
	Inner:       fmt.Errorf("service %s does not have an active version to clone", c.cloneFrom),
	Remediation: fsterr.AutoCloneRemediation,
}
```

Then if the flag is passed, we can add the logic to make an API call to auto-clone their service for them.

pkg/manifest/file.go
```diff
@@ -225,6 +228,13 @@ func (f *File) Write(path string) error {
 	return fp.Close()
 }
 
+// FilePath returns the original path of the file that was read, if any,
+// and a boolean indicating whether the file was read (true) or constructed
+// from env vars or other sources (false).
+func (f *File) FilePath() (string, bool) {
```
Integralist, line 234: Is this used? or was this something you added for future dev/debugging purposes?

pkg/testutil/pointers.go
```diff
@@ -0,0 +1,6 @@
+package testutil
+
+// NullInt returns a pointer to the given int.
+func NullInt(i int) *int {
```
Integralist, line 4: Do we need this or could we just use go-fastly's `ToPointer()` method?

pkg/testutil/stdout.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"bytes"
+	"sync"
+)
+
+// MockStdout provides a concurrent-safe wrapper around bytes.Buffer to
+// replicate the behavior of os.Stdout (or os.Stderr) without
+// introducing potential data races from the buffer resizing itself
+// during Write calls (see bytes.(*Buffer).tryGrowByReslice()).
```
Integralist, line 11: Might be worth adding an extra bit of context to this implementation so it's clear we're referring to the yacc Spinner (as I believe that was what originally prompted this new code).

pkg/commands/compute/init.go
```diff
@@ -27,12 +27,14 @@ import (
...
 	"github.com/fastly/cli/pkg/profile"
 	"github.com/fastly/cli/pkg/text"
+	"github.com/fastly/go-fastly/v9/fastly"
 )
 
 var (
 	gitRepositoryRegEx        = regexp.MustCompile(`((git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)?(/)?`)
 	fastlyOrgRegEx            = regexp.MustCompile(`^https:\/\/github\.com\/fastly`)
 	fastlyFileIgnoreListRegEx = regexp.MustCompile(`\.github|LICENSE|SECURITY\.md|CHANGELOG\.md|screenshot\.png`)
+	fastlyIDRegEx             = regexp.MustCompile("^[0-9a-zA-Z]{22}$")
```
phamann, line 37: Warning: you can't rely on Fastly UUIDs _always_ being `22` characters in length.

pkg/commands/compute/init.go
jsocol: Is it OK to leave `profile` in `fastly.toml` blank? (I'm not really sure how it's used since profiles are defined locally outside of whatever project repo)

Integralist: `profile` is part of the CLI config and not the fastly.toml (the latter being a project-specific configuration file, where as the CLI's internal configuration file is used to store profile information: https://github.com/fastly/cli/blob/main/pkg/config/config.go#L199-L200)

jsocol: It is [included in the manifest file](https://github.com/fastly/cli/blob/main/pkg/manifest/file.go#L31-L32), though I'm not 100% sure why. If we can leave it blank there, I'll drop it from this completely, which seems safest

pkg/app/run.go
jsocol: Yeah, I'm going to take a look at what this would take. I also thought about what it would look like to defer this check until an authenticated API request needs to happen, to avoid needing to pre-list the potential commands. The simplest approach (check to see if `command` matches `/^compute init --from(=|\s)[a-zA-Z0-9]{22}$/` (ish, pending updates from [Patrick's comment](https://github.com/fastly/cli/pull/1187#discussion_r1582989336)) also seems brittle, but may be fine for now, at least?

pkg/commands/compute/init.go
jsocol: I'm not sure if cloning would help in this situation (I should use a different word in the error message). We need a version that will allow us to [fetch the package data](https://www.fastly.com/documentation/reference/api/services/package/#get-package). I tried that with a new compute service with no active versions and I get a 404.

pkg/manifest/file.go
jsocol: It was for handling a potential error that turned out to be impossible, ty!

pkg/testutil/pointers.go
jsocol: Oh I missed that, thank you!

pkg/commands/compute/init.go
Integralist: I don't really recall to be honest but I have a suspicion that _maybe_ it's related to data collection?

jsocol: I took it out with no obvious negative effects. We could follow up with a prompt to include it if we find it's valuable

pkg/commands/compute/init.go
Integralist: Oh, I see. I think I may have misread the code. Yeah you'd need at least one active version of the service for a package to be pullable from it.

jsocol: There's a slightly surprising case where:

* the service exists
* it has no active versions
* it has an _inactive_ version with a package and metadata uploaded

In order to do this you have to go into the UI, create a compute service, upload a package, then never activate or publish it. It's certainly possible, even if unlikely, and we want to make it more unlikely.

So I reworked this to try to find the newest service version that has package data associated with it. If no versions satisfy that requirement, we (currently) bail. We could create a `fastly.toml` with just the service ID in it, or fall back to prompting. My inclination is to do that as a fast-follow UX improvement, but I can certainly do it now, too.

pkg/commands/compute/init.go
```diff
@@ -196,25 +215,130 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+					"From":      from,
+					"Branch":    branch,
+					"Tag":       tag,
+					"Directory": c.dir,
+				})
+				return err
+			}
+		} else {
+			// profileName, _, _ := c.Globals.Profile()
+
```
Integralist, line 245:
```suggestion
```

pkg/commands/compute/init.go
```diff
@@ -196,25 +215,130 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
...
+							if err != nil {
+								if hErr, ok := err.(*fastly.HTTPError); ok {
+									if hErr.IsNotFound() {
+										continue
+									}
+								}
+								return err
+							}
+						}
+					}
```
Integralist, line 297: We should add a `break` directive so we stop the loop once a version is found.

Also, do we need the `*v.Number > serviceVersion` check? I'm guessing if there is any length to `serviceDetails.Version` then that would suggest a valid service version is assigned to `v`. 

We could also use `fastly.ToValue()` instead of the nil check.

```go
for _, v := range serviceDetails.Versions {
	serviceVersion = fastly.ToValue(v.Number)
	
	// ...rest of code...
}
```

pkg/manifest/file.go
```diff
@@ -66,6 +67,8 @@ func (f *File) Read(path string) (err error) {
 		}
 	}()
 
+	f.path = path
+
```
Integralist, line 71: Should this be deleted?

pkg/commands/compute/init.go
jsocol: Ah I might have extrapolated something from the [changelog](https://github.com/fastly/go-fastly/blob/v9.3.1/CHANGELOG.md#v900-2024-02-05) of `go-fastly` here: if the Versions are sorted, then yes we can stop once we find a good version. Otherwise, this is looking for the highest version number with a package attached.

jsocol: Looks like they're sorted by number ascending—I'll revise this to start at the end and break early

pkg/app/run.go
jsocol: tl;dr we needed access to more information here than was available through the command name. This preserves the majority of the logic, switching on the name, but changes the input to take the full command interface.

pkg/app/run.go
```diff
@@ -620,13 +620,19 @@ func commandRequiresAuthServer(command string) bool {
...
-	switch command {
-	case "compute build", "compute hash-files", "compute init", "compute metadata", "compute serve":
+func commandRequiresToken(command argparser.Command) bool {
+	commandName := command.Name()
+	switch commandName {
+	case "compute init":
+		if initCmd, ok := command.(*compute.InitCommand); ok {
+			return compute.IsFastlyID(initCmd.CloneFrom)
+		}
+		return false
```
jsocol, line 630: This really shouldn't happen, it would mean a second `compute init` command was registered somewhere. It might be worth `panic`ing here but that also seems very low-likelihood

jsocol:
@Integralist I'm pretty sure I've caught all the unused bits and addressed all the feedback, mind taking another look?

pkg/app/run.go
```diff
@@ -620,13 +620,19 @@ func commandRequiresAuthServer(command string) bool {
...
 // requires an API token.
-func commandRequiresToken(command string) bool {
-	switch command {
-	case "compute build", "compute hash-files", "compute init", "compute metadata", "compute serve":
+func commandRequiresToken(command argparser.Command) bool {
+	commandName := command.Name()
+	switch commandName {
+	case "compute init":
+		if initCmd, ok := command.(*compute.InitCommand); ok {
+			return compute.IsFastlyID(initCmd.CloneFrom)
```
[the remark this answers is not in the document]
Integralist: yeah I'm not sure the `compute` package is the right place for it either.

pkg/app/run.go
Integralist: Agreed. I think this is fine. I'd rather avoid a panic and it's very unlikely that we'd encounter the command being registered twice.

pkg/app/run.go
Integralist: Yeah this is fine, and nice lateral thinking 👍🏻

Integralist (approved):
LGTM. Going to approve for now, but would be nice to find a better package for `IsFastlyID`. I agree `compute` isn't the right place but I don't think `argparser` is either. Might have to be some other package like https://github.com/fastly/cli/blob/main/pkg/text/service.go (in my head I'm thinking `text` is about handling text input/output and so could make sense to be here)

pkg/app/run.go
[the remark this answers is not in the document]
jsocol: I like the `text` suggestion, I'll move it there pre-merge

PR #1188, opened 2024-04-26
Update tinygo version used in tests

jsocol:
Versions of tinygo before 0.31 do not have precompiled darwin-arm64
binaries. To continue using macos-latest runners, we need to update.

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -88,7 +88,7 @@ jobs:
     needs: [config]
     strategy:
       matrix:
-        tinygo-version: [0.27.0]
+        tinygo-version: [0.31.2]
         go-version: [1.22.x]
         node-version: [18]
         platform: [ubuntu-latest, macos-latest, windows-latest]


Integralist:
Is this related to the CI error in https://github.com/fastly/cli/pull/1187

jsocol:
> Is this related to the CI error in #1187

Yep! I haven't been able to find exactly when, but at some point last week it seems like `macos-latest` Actions runners switched architectures.

I wanted to double check before merging, though, that there are no other reasons we've been using 0.27.0? Are there any potential issues or unknowns we should validate?

Integralist:
Probably just the first tinygo version that we could use when developing the feature.

If I check the default CLI configuration...
https://github.com/fastly/cli/blob/main/.fastly/config.toml#L15-L16

We can see that we set a minimum tinygo version of 0.28.x 🤔

Integralist:
And for clarification, in case you were unfamiliar with this file:
https://github.com/fastly/cli/blob/main/DEVELOPMENT.md#fastlyconfigtoml

jsocol:
Thanks! Just to double check my understanding, `.fastly/config.toml` ends up embedded in `pkg/config.Static` and the data ends up emitted for Go compute projects into the new project's `fastly.toml`, right? I wouldn't think we want to change the minimum tinygo version for WASI, but maybe it's OK to change the _default_ for new `fastly.toml`? Should we do that in the same PR? Happy to add it, unless it deserves it's own write-up

PR #1189, opened 2024-04-29
Pass a cert-file argument instead of full cert blob.

rdoherty-fastly:
WIP, needs help!

Preliminary work in progress to allow a user to specify a filename for their custom cert. I've tested it manually and now I want some good tests to verify that the cli reads the file and passes it to the API.

I'm having trouble with figuring out how to get the contents of the certblob that is passed to the mock api and verifying it is correct. Line 75 in  `pkg/commands/tls/custom/certificate/certificate_test.go` shows where the value can be read, but afaict there's no access to the testcase object.

I'd love some help/advice on what the best way to do this is, I'm still new to Golang, especially wrt tests!

### DIFF
--- a/pkg/commands/tls/custom/certificate/certificate_test.go
+++ b/pkg/commands/tls/custom/certificate/certificate_test.go
@@ -23,17 +23,42 @@ const (
 )
 
 func TestTLSCustomCertCreate(t *testing.T) {
+	var content string
 	args := testutil.Args
 	scenarios := []testutil.TestScenario{
 		{
-			Name:      "validate missing --cert-blob flag",
+			Name:      "validate missing --cert-blob and --cert-path flags",
 			Args:      args("tls-custom certificate create"),
-			WantError: "required flag --cert-blob not provided",
+			WantError: "neither --cert-path or --cert-blob provided, one must be provided",
+		},
+		{
+			Name:      "validate specifying both --cert-blob and --cert-path flags",
+			Args:      args("tls-custom certificate create --cert-blob foo --cert-path bar"),
+			WantError: "cert-path and cert-blob provided, only one can be specified",
+		},
+		{
+			Name:      "validate invalid --cert-path arg",
+			Args:      args("tls-custom certificate create --cert-path ............"),
+			WantError: "error reading cert-path",
+		},
+		{
+			Name: "validate custom cert is submitted",
+			API: mock.API{
+				CreateCustomTLSCertificateFn: func(certInput *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+					content = certInput.CertBlob
+					return &fastly.CustomTLSCertificate{
+						ID: mockResponseID,
+					}, nil
+				},
+			},
+			Args:       args("tls-custom certificate create --cert-path ./testdata/certificate.crt"),
+			WantOutput: fmt.Sprintf("Created TLS Certificate '%s'", mockResponseID),
 		},
 		{
 			Name: validateAPIError,
 			API: mock.API{
-				CreateCustomTLSCertificateFn: func(_ *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+				CreateCustomTLSCertificateFn: func(certInput *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+					content = certInput.CertBlob
 					return nil, testutil.Err
 				},
 			},
@@ -43,7 +68,8 @@ func TestTLSCustomCertCreate(t *testing.T) {
 		{
 			Name: validateAPISuccess,
 			API: mock.API{
-				CreateCustomTLSCertificateFn: func(_ *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+				CreateCustomTLSCertificateFn: func(certInput *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+					content = certInput.CertBlob
 					return &fastly.CustomTLSCertificate{
 						ID: mockResponseID,
 					}, nil
@@ -66,6 +92,7 @@ func TestTLSCustomCertCreate(t *testing.T) {
 			err := app.Run(testcase.Args, nil)
 			testutil.AssertErrorContains(t, err, testcase.WantError)
 			testutil.AssertStringContains(t, stdout.String(), testcase.WantOutput)
+			testutil.AssertPathContentFlag("cert-path", testcase.WantError, testcase.Args, "certificate.crt", content, t)
 		})
 	}
 }
--- a/pkg/commands/tls/custom/certificate/create.go
+++ b/pkg/commands/tls/custom/certificate/create.go
@@ -1,7 +1,10 @@
 package certificate
 
 import (
+	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 
 	"github.com/fastly/go-fastly/v9/fastly"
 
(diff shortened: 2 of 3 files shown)


pkg/commands/tls/custom/certificate/create.go
```diff
@@ -53,16 +63,41 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
...
+		if path, err := filepath.Abs(c.certPath); err == nil {
+			if data, err := os.ReadFile(path); err == nil /* #nosec */ {
+				content := string(data)
+				input.CertBlob = content
+			} else {
+				return nil, fmt.Errorf("error reading cert-path: '%s' . Error: %q", path, err)
+			}
+		} else {
+			return nil, fmt.Errorf("cert file path invalid: '%s' . Error '%q'", path, err)
+		}
```
Integralist, line 95: I think it's more idiomatic to do this in reverse and handle the error cases first (also avoid the extra `else` + make the error message format consistent). The less nesting the better for readability...

```go
path, err := filepath.Abs(c.certPath)
if err != nil {
	return nil, fmt.Errorf("error parsing cert-path '%s': %q", path, err)
}

data, err := os.ReadFile(path) // #nosec	
if err != nil {
	return nil, fmt.Errorf("error reading cert-path '%s': %q", path, err)
}

input.CertBlob = string(data)
```

pkg/commands/tls/custom/certificate/certificate_test.go
```diff
@@ -25,33 +25,63 @@ const (
...
+		//		CreateCustomTLSCertificateFn: func(_ *fastly.CreateCustomTLSCertificateInput) (*fastly.CustomTLSCertificate, error) {
+		//			return nil, testutil.Err
+		//		},
+		//	},
+		//	Args:      args("tls-custom certificate create --cert-blob example"),
+		//	WantError: testutil.Err.Error(),
+		//},
+		//{
+		//	Name: validateAPISuccess,
+		//	API: mock.API{
```
Integralist, line 75: You might want to try using a global variable for this (it's not pretty but it's how I've done this in the past)...

https://github.com/fastly/cli/blob/6c1ec982475e698bd347c94622f857d9e4f44c90/pkg/commands/vcl/snippet/snippet_test.go#L17

https://github.com/fastly/cli/blob/6c1ec982475e698bd347c94622f857d9e4f44c90/pkg/commands/vcl/snippet/snippet_test.go#L49-L50

https://github.com/fastly/cli/blob/6c1ec982475e698bd347c94622f857d9e4f44c90/pkg/commands/vcl/snippet/snippet_test.go#L206

jsocol:
Looks like the test failure is the same thing I hit in #1187. If you rebase on main you should pick up the fix from #1188

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -16,8 +19,10 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
 	c.CmdClause = parent.Command("create", "Create a TLS certificate").Alias("add")
 	c.Globals = g
 
-	// Required.
-	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob").Required().StringVar(&c.certBlob)
+	//Required
+	// cert-blob and cert-file are mutually exclusive. One is required.
+	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob, mutually exclusive with --cert-file").StringVar(&c.certBlob)
+	c.CmdClause.Flag("cert-file", "Filepath to certificate, mutually exclusive with --cert-blob").StringVar(&c.certPath)
```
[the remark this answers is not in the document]
Integralist: I'm presuming he chose this for consistency with other CLI flags where we're historically used `--file`, but I would tend to agree that `--cert-path` sounds better here.

As far as the description I think a slight tweak would help:

```suggestion
	c.CmdClause.Flag("cert-path", "Filepath to a PEM-formatted certificate, mutually exclusive with --cert-blob").StringVar(&c.certPath)
```

Integralist:
As mentioned by @jsocol earlier the failing CI https://github.com/fastly/cli/actions/runs/8881360752/job/24383545720?pr=1189 is because of a change that rolled out in the GitHub runners caused a change in macos-latest architecture. So rebasing `main` should hopefully fix that issue.

rdoherty-fastly:
Thanks @jsocol and @Integralist for the tips! Will be a few days before I have time to work on this again.

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -16,8 +19,10 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
 	c.CmdClause = parent.Command("create", "Create a TLS certificate").Alias("add")
 	c.Globals = g
 
-	// Required.
-	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob").Required().StringVar(&c.certBlob)
+	//Required
```
Integralist, line 22:
```suggestion
	// Required
```

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -16,8 +19,10 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
...
+	// cert-blob and cert-path are mutually exclusive. One is required.
+	c.CmdClause.Flag("cert-blob", "The PEM-formatted certificate blob, mutually exclusive with --cert-path").StringVar(&c.certBlob)
+	c.CmdClause.Flag("cert-path", "Filepath to  a PEM-formatted certificate, mutually exclusive with --cert-blob").StringVar(&c.certPath)
```
Integralist, line 25:
```suggestion
	c.CmdClause.Flag("cert-path", "Filepath to a PEM-formatted certificate, mutually exclusive with --cert-blob").StringVar(&c.certPath)
```

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -31,13 +36,18 @@ type CreateCommand struct {
...
 	id       string
 	name     string
 }
 
 // Exec invokes the application logic for the command.
 func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
-	input := c.constructInput()
+	input, err := c.constructInput()
+
+	if err != nil {
```
Integralist, line 48:
```suggestion
	input, err := c.constructInput()
	if err != nil {
```

pkg/commands/tls/custom/certificate/certificate_test.go
```diff
@@ -23,17 +23,47 @@ const (
...
+		{
+			Name:      "validate invalid --cert-path arg",
+			Args:      args("tls-custom certificate create --cert-path bar"),
+			WantError: "no such file or directory",
+		},
+		{
+			Name:      "validate nonsensical --cert-path arg",
+			Args:      args("tls-custom certificate create --cert-path foobar"),
+			WantError: "no such file or directory",
+		},
```
Integralist, line 48: What's the difference between these two tests? They both look like they're passed an input that will yield the same error. Is there logic inside the CLI that distinguishes the two behaviours more than what I'm seeing in these two test cases?

**EDIT:** Reading ahead I see there are two error cases `error parsing cert-path` and `error reading cert-path` so probably best to rework these two test cases such that we validate each of those logic flow paths.

Integralist (commented):
Looking good. Some comments to be addresses + the CI is failing to run the tests successfully (looks to be failing on `TestTLSCustomCertCreate/validate_custom_cert_is_submitted`)

pkg/commands/tls/custom/certificate/certificate_test.go
rdoherty-fastly: Ok I've tried to trigger `filepath.Abs(c.certPath)` to return an error but have been unable to give it a path that it considers invalid. I'm removing the nonsensical test because of this.

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -53,16 +62,44 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
...
 	}
-	input.CertBlob = c.certBlob
+
+	if c.certBlob != "" {
+		input.CertBlob = c.certBlob
+	}
+
+	if c.certPath != "" {
+		path, err := filepath.Abs(c.certPath)
+
```
Integralist, line 86:
```suggestion
```

pkg/commands/tls/custom/certificate/create.go
```diff
@@ -53,16 +62,43 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
...
+	}
+
+	if c.certPath != "" {
+		path, err := filepath.Abs(c.certPath)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing cert-path '%s': %q", c.certPath, err)
+		}
+
+		data, err := os.ReadFile(path) // #nosec
+
```
Integralist, line 91:
```suggestion
```

PR #1190, opened 2024-04-30
feat(compute): notify the API of progress to update the UI

jsocol:
Creates a new internal `beacon` package to send certain events and results back to the API so that we can update the UI based on actions from the CLI. We're adding these to `compute init` and `compute deploy` initially.

Not ready for review but wanted to get the full cross-platform test suite running.

TODO:
- [x] Fill in the rest of the `compute deploy` payload and test
- [x] Decide how/if to deal with errors
- [x] Add the beacon to `compute init`

### DIFF
--- a/pkg/commands/compute/deploy_test.go
+++ b/pkg/commands/compute/deploy_test.go
@@ -1,6 +1,7 @@
 package compute_test
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


pkg/commands/compute/deploy_test.go
```diff
@@ -2239,6 +2240,95 @@ func TestDeploy(t *testing.T) {
...
+
+	stdout := threadsafe.Buffer{}
+	args := testutil.Args("compute deploy --auto-yes --non-interactive")
+	recordingHTTP := testutil.NewRecordingHTTPClient()
+	recordingHTTP.SingleResponse(&http.Response{
+		StatusCode: http.StatusNoContent,
+		Status:     http.StatusText(http.StatusNoContent),
+		Body:       io.NopCloser(bytes.NewBufferString("")),
+	}, nil)
+	fmt.Printf("recording HTTP with...\n%#v\n", recordingHTTP)
```
Integralist, line 2288: Could we have this encapsulated inside of the `NewRecordingHTTPClient`.

pkg/commands/compute/init.go
```diff
@@ -219,6 +221,26 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		c.CloneFrom = from
 	}
 
+	defer func() {
```
Integralist, line 224: Looks like this only tracks users who pass the `--from` flag?

Is there a reason not to define this `defer` earlier in the `Exec()` function and to have it beacon for different types of users?

pkg/commands/compute/init.go
```diff
@@ -234,6 +256,7 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 	// that isn't natively supported by the platform.
 	if c.CloneFrom != "" {
 		if !isExistingService {
+			fmt.Printf("Fetching package from: %s\n", c.CloneFrom)
```
Integralist, line 259: Two general notes to make...

1. We should use `text.Output` or `text.Info` for these messages as it means we can pass in mocked stdout (that is passed in as the `out` argument to the `Exec()` method.

2. We should probably wrap this in a check for the verbose flag (as some users would consider this unnecessary noise).

Otherwise, we already inform the user of this information here:
https://github.com/fastly/cli/blob/8912844aa6dc0aa64f790b5ec40f78fe4c886540/pkg/commands/compute/init.go#L82-L87

pkg/commands/compute/init.go
```diff
@@ -812,7 +835,29 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
...
+	if u.Host == "" && u.Scheme == "" {
+		spinner.StopFailMessage(msg)
+		spinErr := spinner.StopFail()
+		if spinErr != nil {
+			return fmt.Errorf(text.SpinnerErrWrapper, spinErr, err)
+		}
+		return fmt.Errorf("--from url seems invalid: %s", c.CloneFrom)
+	}
+
+	req, err := http.NewRequest("GET", u.String(), nil)
```
Integralist, line 860: I'm interested to know why the extra code here. Was there an unclear error output from `http.NewRequest("GET", c.CloneFrom, nil)` that adding this extra logic was able to provide clarity for?

Integralist (commented):
Looking good. Just had a couple of comments.

pkg/commands/compute/deploy_test.go
jsocol: Oops, that was left over from some debugging

pkg/commands/compute/init.go
jsocol: We actually can't send beacons if the service ID doesn't exist yet, so if someone is running `init` without `--from=<service_id>` we don't have anywhere to even send the notification. It would certainly be possible to move this up—and simplify this condition—since `c.CloneFrom` is a service ID iff `isExistingService` is true.

pkg/commands/compute/init.go
jsocol: This was also leftover debugging, honestly 😓 I'll do another pass to make sure no more of that stayed in

pkg/commands/compute/init.go
jsocol: Ah that's the comment on 849/850. If you pass something that's _almost_ a service ID but is too short or too long into `url.Parse`, you'll get something like

```go
u, _ := url.Parse("abcd1234")
fmt.Printf("%#v\n", u")
// url.URL{
//   Scheme: "",
//   Host:   "",
//   Path:   "abcd1234",
    // ... etc
// }
```

Without this check, the errors trying to fetch the path with no other information were pretty obtuse because they came out of the HTTP client. How could I make this clearer?

pkg/testutil/http.go
```diff
@@ -0,0 +1,86 @@
+package testutil
```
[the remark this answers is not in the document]
jsocol: 👍🏻 I'll do a follow-up. I also found in this PR, but not the last one, the `threadsafe.Buffer` type, which can replace the `testutil.MockStdout`, so that can also be consolidated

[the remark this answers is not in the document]
Integralist: Awesome! Thanks. When I saw `MockStdout` I was like... I'm sure I've fixed this before 🤔 😅

pkg/commands/compute/init.go
Integralist: Gotcha. Makes sense.

pkg/commands/compute/init.go
Integralist: No that's fine, that makes sense now 👍🏻

Integralist (approved):
OK I think we're good here.

PR #1197, opened 2024-05-07
fix(compute/rust) handling of 'cargo version' output

kpfleming:
'cargo version' emits the version string to stdout, so the code should capture only stdout, and parse it for checking the ToolchainConstraint. Capturing stderr and including it in the parsing process can result in bogus version numbers being found and reported (as is happening right now when Rust 1.78 is used in a project directory where .cargo/config exists).

### DIFF
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -312,8 +312,8 @@ func (r *Rust) toolchainConstraint() {
 	// #nosec
 	// nosemgrep
 	cmd := exec.Command(args[0], args[1:]...)
-	stdoutStderr, err := cmd.CombinedOutput()
-	output := string(stdoutStderr)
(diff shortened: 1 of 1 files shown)


Integralist (approved):
LGTM. But can we manually test this change with different Rust/cargo versions to be sure there are no unexpected issues.

kpfleming:
I'll be testing it with 1.78 (the version causing the problem now) and 1.56 (the oldest version we claim to support).

kpfleming:
Tests complete:
1. With Rust 1.78.0, the bogus warning about "version 1.38 didn't meet the constraint" is gone (that version number only appears on stderr, not stdout).
2. With Rust 1.56.0, the constraint-checking code correctly reports that 1.56.0 doesn't meet the constraint.

PR #1199, opened 2024-05-07
feat(compute/rust): Handle Cargo config filename for Rust >=1.78.0

kpfleming:
The preferred Cargo configuration filename has been 'config.toml' since Rust 1.39.0 was released, but 'config' was supported for backwards compatibility (although deprecated).
    
In Rust 1.78.0, the backwards compatibility was removed, so any file with the old name will be ignored. Since Compute package builds require the content in that file, ignoring it results in a build failure.
    
This patch adds checking for this situation. If the old file name is found, and the Rust version is <1.78.0, a warning is issued to let the user know that they should rename the file before upgrading to 1.78.0 or later.  If the old file name is found, and the Rust version is >=1.78.0, an error is emitted and the build process stopped, as it will not succeed.

Example output (with old file name and Rust 1.74.0):
```
kpfleming@kpfleming:~/src/fastly/compute-starter-kit-rust-default$ ~/src/fastly/cli/fastly compute build
✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain
WARNING: The Cargo configuration file name is .cargo/config
WARNING: The file should be renamed to .cargo/config.toml to be compatible with Rust 1.78.0 or later

✓ Running [scripts.build]
✓ Creating package archive
```
Example output (with old file name and Rust 1.78.0):
```
kpfleming@kpfleming:~/src/fastly/compute-starter-kit-rust-default$ ~/src/fastly/cli/fastly compute build
✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain
WARNING: The Cargo configuration file name is .cargo/config

ERROR: The build cannot proceed with Rust version '1.78.0' as the file must be named .cargo/config.toml.

```

### DIFF
--- a/pkg/commands/compute/language_rust.go
+++ b/pkg/commands/compute/language_rust.go
@@ -157,7 +157,14 @@ func (r *Rust) Build() error {
 		text.Info(r.output, "No [scripts.build] found in %s. The following default build command for Rust will be used: `%s`\n\n", r.manifestFilename, r.build)
 	}
 
-	r.toolchainConstraint()
+	version := r.toolchainConstraint()
+
+	if version != nil {
+		err := r.checkCargoConfigFileName(version)
+		if err != nil {
+			return err
+		}
+	}
 
 	bt := BuildToolchain{
 		autoYes:                   r.autoYes,
@@ -294,7 +301,7 @@ func (r *Rust) modifyCargoPackageName(noBuildScript bool) error {
 // NOTE: We don't stop the build as their toolchain may compile successfully.
 // The warning is to help a user know something isn't quite right and gives them
 // the opportunity to do something about it if they choose.
-func (r *Rust) toolchainConstraint() {
+func (r *Rust) toolchainConstraint() *semver.Version {
 	if r.verbose {
 		text.Info(r.output, "The Fastly CLI requires a Rust version '%s'.\n\n", r.config.ToolchainConstraint)
 	}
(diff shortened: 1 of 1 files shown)


pkg/commands/compute/language_rust.go
```diff
@@ -312,32 +319,57 @@ func (r *Rust) toolchainConstraint() {
...
+		return nil
+	}
+
+	text.Warning(r.output, "The Cargo configuration file name is .cargo/config")
+
+	c, _ := semver.NewConstraint(">=1.78.0")
+
+	if c.Check(rustVersion) {
+		return fmt.Errorf("The build cannot proceed with Rust version '%s' as the file must be named .cargo/config.toml", rustVersion)
+	} else {
```
Integralist, line 369: As suggested by the GitHub linter...

Screenshot 2024-05-08 at 06 42 25

pkg/commands/compute/language_rust.go
```diff
@@ -312,32 +319,57 @@ func (r *Rust) toolchainConstraint() {
...
+	if err != nil {
+		r.errlog.Add(err)
+		return fmt.Errorf("getting current working directory: %w", err)
+	}
+
+	if !filesystem.FileExists(filepath.Join(dir, ".cargo", "config")) {
+		return nil
+	}
+
+	text.Warning(r.output, "The Cargo configuration file name is .cargo/config")
```
Integralist, line 363: Let's get some breathing space between the WARNING and the preceding line (as I can see from your example output in the PR description that they're butted up against each other)
```suggestion
	text.Warning(r.output, "\nThe Cargo configuration file name is .cargo/config")
```

Integralist (commented):
This isn't the end of the world but I would prefer not to have two WARNING lines next to each other when they both relate to the same warning.

I can see why that has happened by looking at the code and the fact you're trying to swap out the second warning for an error (depending on the situation) but it just looks a bit weird.

pkg/commands/compute/language_rust.go
Integralist, line 363: To fix the issue of a double WARNING message we should change the flow to...

```go
warnMsg := "\nThe Cargo configuration file name is .cargo/config"

c, _ := semver.NewConstraint(">=1.78.0")

if c.Check(rustVersion) {
    text.Warning(r.output, warnMsg)
    return fmt.Errorf("The build cannot proceed with Rust version '%s' as the file must be named .cargo/config.toml", rustVersion)
}

text.Warning(r.output, warnMsg + ". The file should be renamed to .cargo/config.toml to be compatible with Rust 1.78.0 or later\n\n")
return nil
```

kpfleming: This makes sense, will do it now.

kpfleming:
New output:

```
kpfleming@kpfleming:~/src/fastly/compute-starter-kit-rust-default$ ~/src/fastly/cli/fastly compute build
✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain

WARNING: The Cargo configuration file name is .cargo/config

ERROR: The build cannot proceed with Rust version '1.78.0' as the file must be named .cargo/config.toml.

```

and

```
kpfleming@kpfleming:~/src/fastly/compute-starter-kit-rust-default$ ~/src/fastly/cli/fastly compute build
✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain

WARNING: The Cargo configuration file name is .cargo/config. The file should be renamed to
.cargo/config.toml to be compatible with Rust 1.78.0 or later

✓ Running [scripts.build]
✓ Creating package archive

SUCCESS: Built package (pkg/Default-starter-for-Rust.tar.gz)
```

Integralist (approved):
🎉

PR #1207, opened 2024-05-16
feat(sso): support active session account switching

Integralist:


pkg/auth/auth.go
```diff
@@ -89,17 +93,29 @@ func (s Server) AuthURL() (string, error) {
 		return "", err
 	}
 
+	// We only send login_hint if the user has switched profiles.
```
Integralist, line 96: Not true any more. I've setting it when we create/update too.

phamann (approved):
:+1: LGTM

Issue #1210, opened 2024-05-21
ERROR: failed to configure authentication processes: failed to unmarshal OpenID Connect .well-known metadata: invalid character 'I' looking for beginning of value


marceloboeira:
**Version**

```
Fastly CLI version v10.8.0 (8b3c0372)
Built with go version go1.21.5 linux/amd64 (2024-05-21)
Viceroy version: viceroy 0.9.3

A new version of the Fastly CLI is available.
Current version: 10.8.0
```

**What happened**

We have a pipeline, that hasn't changed in a while, with this command:  
```sh
fastly compute publish --verbose
```

It normally publishes the application to compute @ edge, now it started failing with:

```
ERROR: failed to configure authentication processes: failed to unmarshal OpenID Connect .well-known metadata: invalid character 'I' looking for beginning of value.
```

Note: It seems to be related to https://www.fastlystatus.com/incident/376536, I'm guessing the `invalid character 'I' looking for beginning of value` is for `Internal Server Error` but thought of reporting it here should other people face it too.

Integralist:
Hi @marceloboeira 

You're correct, there was an incident that appears to have caused an issue wit the OICD endpoint used by the CLI (I myself just ran into this). The endpoint is back up already but I'll leave this issue open for now until I get a confirmation that the underlying issue is confirmed resolved.

marceloboeira:
@Integralist looking good now! I'm wondering if that's an opportunity to catch that type of error and improve the message (e.g.: if a 5xx is returned then switch the message to "An unexpected authentication error occurred, check http://fastlystatus.com for any possible incidents" and then some debug details (e.g.: statuscode/body)).

Happy to work on an MR for that if the behaviour is desired. 

In any case, I'll close this issue since the original problem is resolved.

Integralist:
Hi @marceloboeira 

> I'm wondering if that's an opportunity to catch that type of error and improve the message
> Happy to work on an MR for that if the behaviour is desired.

Yes please! That would be very much appreciated 🙂 🙇🏻

PR #1211, opened 2024-05-22
Handle Fastly API unresponsiveness Errors on Authentication

marceloboeira:
When Fastly servers are unresponsive (indicated by a 5xx response), the CLI currently shows a cryptic error message due to unexpected values in the response body (e.g., unable to parse body as JSON).

The last incident caused the CLI to display:

> ERROR: failed to configure authentication processes: failed to unmarshal OpenID Connect .well-known metadata: invalid character 'I' looking for beginning of value

The 'I' referred to the "Internal Server Error" in the response body, which is unclear/unactionable from a user perspective.

This change introduces a validation to handle this specific scenario:

If the authentication request returns a 5xx status code, it is now recognized as an error. The CLI will display a more actionable message, informing clients that the error is likely due to an outage or incident on the Fastly side. Clients can then check the status page or contact support.

Context:
* https://github.com/fastly/cli/issues/1210
* https://www.fastlystatus.com/incident/376536

### DIFF
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -657,7 +657,7 @@ func configureAuth(apiEndpoint string, args []string, f config.File, c api.HTTPC
 	}
 	// Set a more meaningful error message when Fastly servers are unresponsive
 	// check if the response code is a 500 or above
-	if resp.StatusCode >= 500 {
+	if resp.StatusCode >= http.StatusInternalServerError {
 		var body string // default to empty string if we fail to read the body
 		body, _ := io.ReadAll(resp.Body)
 		return nil, fmt.Errorf("the Fastly servers are unresponsive, please check the Fastly Status page (https://fastlystatus.com) and reach out to support if the error persists (HTTP Status Code: %d, Error Message: %s)", resp.StatusCode, body)
--- a/pkg/app/run.go
+++ b/pkg/app/run.go
@@ -658,7 +658,7 @@ func configureAuth(apiEndpoint string, args []string, f config.File, c api.HTTPC
 	// Set a more meaningful error message when Fastly servers are unresponsive
 	// check if the response code is a 500 or above
 	if resp.StatusCode >= 500 {
-		return nil, fmt.Errorf("Fastly servers are unresponsive, please check the Status Page (https://fastlystatus.com) and reach out to support if the error persists — More Details: (HTTP Status Code: %s, Error Message: %s)", resp.StatusCode, resp.Body)
+		return nil, fmt.Errorf("Fastly servers are unresponsive, please check the Status Page (https://fastlystatus.com) and reach out to support if the error persists — More Details: (HTTP Status Code: %d, Error Message: %s)", resp.StatusCode, resp.Body)
 	}
 
 	openIDConfig, err := io.ReadAll(resp.Body)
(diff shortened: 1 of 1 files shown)


Integralist (approved):

pkg/app/run.go
```diff
@@ -655,6 +655,11 @@ func configureAuth(apiEndpoint string, args []string, f config.File, c api.HTTPC
 	if err != nil {
 		return nil, fmt.Errorf("failed to request OpenID Connect .well-known metadata (%s): %w", metadataEndpoint, err)
 	}
+	// Set a more meaningful error message when Fastly servers are unresponsive
+	// check if the response code is a 500 or above
+	if resp.StatusCode >= 500 {
+		return nil, fmt.Errorf("Fastly servers are unresponsive, please check the Status Page (https://fastlystatus.com) and reach out to support if the error persists — More Details: (HTTP Status Code: %s, Error Message: %s)", resp.StatusCode, resp.Body)
```
Integralist, line 661: govet reported...

```
pkg/app/run.go:661:15: fmt.Errorf format %s has arg resp.StatusCode of wrong type int
```

Looks like we need to switch to `%d`...

```suggestion
		return nil, fmt.Errorf("Fastly servers are unresponsive, please check the Status Page (https://fastlystatus.com) and reach out to support if the error persists — More Details: (HTTP Status Code: %d, Error Message: %s)", resp.StatusCode, resp.Body)
```

pkg/app/run.go
```diff
@@ -655,6 +655,11 @@ func configureAuth(apiEndpoint string, args []string, f config.File, c api.HTTPC
 	if err != nil {
 		return nil, fmt.Errorf("failed to request OpenID Connect .well-known metadata (%s): %w", metadataEndpoint, err)
 	}
+	// Set a more meaningful error message when Fastly servers are unresponsive
+	// check if the response code is a 500 or above
+	if resp.StatusCode >= 500 {
+		return nil, fmt.Errorf("Fastly servers are unresponsive, please check the Status Page (https://fastlystatus.com) and reach out to support if the error persists — More Details: (HTTP Status Code: %d, Error Message: %s)", resp.StatusCode, resp.Body)
```
Integralist, line 661: Something else I just realised, we'll want to actually read the body...
```suggestion
		var body string // default to empty string if we fail to read the body
		body, _ := io.ReadAll(resp.Body)
		return nil, fmt.Errorf("the Fastly servers are unresponsive, please check the Fastly Status page (https://fastlystatus.com) and reach out to support if the error persists (HTTP Status Code: %d, Error Message: %s)", resp.StatusCode, body)
```

pkg/app/run.go
```diff
@@ -655,6 +655,13 @@ func configureAuth(apiEndpoint string, args []string, f config.File, c api.HTTPC
 	if err != nil {
 		return nil, fmt.Errorf("failed to request OpenID Connect .well-known metadata (%s): %w", metadataEndpoint, err)
 	}
+	// Set a more meaningful error message when Fastly servers are unresponsive
+	// check if the response code is a 500 or above
+	if resp.StatusCode >= 500 {
```
Integralist, line 660: I'd prefer to use the stdlib constants here...

```suggestion
	if resp.StatusCode >= http.StatusInternalServerError {
```

Integralist:
This will teach me to make edits via the GitHub UI. I'll pull your branch and fix these things up locally 🙂

Integralist:
Superseded by https://github.com/fastly/cli/pull/1212

PR #1212, opened 2024-05-22
feat(app): improve error messaging when Fastly servers are unresponsive

Integralist:
Thanks to @marceloboeira for the [first incarnation](https://github.com/fastly/cli/pull/1211) of these changes which I butchered via the GitHub UI.


marceloboeira:
Nice one!

PR #1213, opened 2024-05-28
feat(compute): clone starter kit source with init --from=serviceID

jsocol:
When we initialize an empty service with `compute init --from=<serviceID>`, we don't have the source code. This change attempts to look at the `cloned_from` metadata and use that to fetch the rest of the project.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.3.2
+	github.com/fastly/go-fastly/v9 v9.4.1-0.20240528142352-98043629c62b
 	github.com/hashicorp/cap v0.6.0
 	github.com/kennygrant/sanitize v1.2.4
 	github.com/mholt/archiver v3.1.1+incompatible
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.4.1-0.20240528142352-98043629c62b
+	github.com/fastly/go-fastly/v9 v9.5.0
 	github.com/hashicorp/cap v0.6.0
 	github.com/kennygrant/sanitize v1.2.4
 	github.com/mholt/archiver v3.1.1+incompatible
--- a/pkg/commands/compute/init.go
+++ b/pkg/commands/compute/init.go
@@ -336,7 +336,7 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			if pack.Metadata != nil {
 				clonedFrom := fastly.ToValue(pack.Metadata.ClonedFrom)
 				if serviceVersion > 1 {
-					out.Write([]byte("Service has active versions, not fetching starter kit source"))
+					text.Info(out, "\nService has active versions, not fetching starter kit source\n\n")
 				} else if gitRepositoryRegEx.MatchString(clonedFrom) {
 					err = spinner.Process("Initializing file structure from selected starter kit", func(*text.SpinnerWrapper) error {
 						err := c.ClonePackageFromEndpoint(clonedFrom, "", "")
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -27,7 +27,7 @@ require (
 )
 
 require (
-	github.com/fastly/go-fastly/v9 v9.3.2
+	github.com/fastly/go-fastly/v9 v9.4.1-0.20240528142352-98043629c62b
```
jsocol, line 30: This is the current `main` branch as of today. We probably want to release actual 9.4.1 (or 9.5)

kpfleming:
The clone operation will pull the 'original' source directly from GitHub? Will it pull a specific commit hash to ensure that it gets the same thing that was used to initialize the service?

If it will be pulling directly from GitHub, the error handling will need to make that clear, since any failure on GitHub's part will make 'fastly compute init' fail and the user will assume it's a Fastly problem.

go.mod
fgsch: Is this needed for this PR?

fgsch:
Sorry for changing the title of this PR. I should have looked a bit more but confused by the description (or lack of) for the `--from` parameter.  

Could we mention the serviceID in here or this was omitted on purpose?

go.mod
Integralist: I've just published https://github.com/fastly/go-fastly/releases/tag/v9.5.0 so let's update this dependency reference

Integralist (commented):
👋🏻 

I feel like the `--from=<ServiceID>` feature is a bit misleading (dare I say... less than useful) or I guess I'm just generally a bit confused by its motivation/purpose 🤔

In my mind if I use `--from=<ServiceID>` then I'm expecting the code for that package to be made available to me so I can continuing developing the package. But from what I can tell this is just downloading the starter kit that the package was _originally_ based on, and I'm not sure how that's useful.

Now if the original developer of the package had set `--include-source` as part of the `fastly compute build` subcommand, then I guess we might be able to extract the original source code from the uploaded package (e.g. download the package, and then extract code from the `package.tar.gz` file).

fgsch:
> > Could we mention the serviceID in here or this was omitted on purpose?
> 
> Sorry, I'm not totally sure I follow. Mention it where, in the PR, or somewhere in code/messaging?

In the help description:

```
  -f, --from=FROM            Local project directory, or Git repository URL,
                             or URL referencing a .zip/.tar.gz file, containing
                             a package template
```

jsocol:
@Integralist @kpfleming @fgsch Updated with feedback addressed. After some internal discussions, there are three changes here:

1. Messaging around fetching the starter kit source has changed to make it clearer that's what we're using
2. Updated the logic to only fetch the source if the active version is 1
3. Added details to the `--help` text that mention service IDs and the starter kits

pkg/commands/compute/init.go
```diff
@@ -332,29 +334,50 @@ func (c *InitCommand) Exec(in io.Reader, out io.Writer) (err error) {
 			}
 
 			if pack.Metadata != nil {
+				clonedFrom := fastly.ToValue(pack.Metadata.ClonedFrom)
+				if serviceVersion > 1 {
+					out.Write([]byte("Service has active versions, not fetching starter kit source"))
```
Integralist, line 339:
```suggestion
					text.Info(out, "Service has active versions, not fetching starter kit source")
```

You might have to check the output (both with and without the `--verbose` flag) and add preceding or trailing line breaks to make sure this INFO note has some breathing space around it from other output.

If you add line breaks and they're the same regardless of `--verbose` then just add them inline to the string message (e.g. `"\nService has active versions, not fetching starter kit source\n\n"`). Otherwise, if the line break is dependant on `--verbose`, then use something like...

```go
if c.Globals.Verbose() {
    text.Break(out)
}
```

Integralist (approved):
One small requested change, otherwise LGTM.

pkg/commands/compute/init.go
jsocol: Done, I went with fewer branches/more consistent since this feels important for expectation setting—if you run this against an existing service you aren't going to get source

Integralist (approved):

PR #1214, opened 2024-05-29
 Adds --cert-path argument to tls-custom certificate update command to pass in a path to a certificate file

rdoherty-fastly:
Adds --cert-path argument to `tls-custom certificate update` command to pass in a path to a certificate file
Previously users were required to pass in the full certificate contents via the --cert-blob argument, which is cumbersome and error-prone. This allows users to specify a path to a certificate making this command easier to use.

The new argument --cert-path is mutually exclusive with --cert-blob and the command will return an error if both are specified.

This is a near duplicate of https://github.com/fastly/cli/pull/1189 except for the `update` command vs `create`

### DIFF
--- a/pkg/commands/tls/custom/certificate/certificate_test.go
+++ b/pkg/commands/tls/custom/certificate/certificate_test.go
@@ -255,22 +255,34 @@ func TestTLSCustomCertList(t *testing.T) {
 }
 
 func TestTLSCustomCertUpdate(t *testing.T) {
+	var content string
 	args := testutil.Args
 	scenarios := []testutil.TestScenario{
-		{
-			Name:      "validate missing --cert-blob flag",
-			Args:      args("tls-custom certificate update --id example"),
-			WantError: "required flag --cert-blob not provided",
-		},
 		{
 			Name:      validateMissingIDFlag,
 			Args:      args("tls-custom certificate update --cert-blob example"),
 			WantError: "required flag --id not provided",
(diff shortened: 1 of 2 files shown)


pkg/commands/tls/custom/certificate/certificate_test.go
```diff
@@ -255,22 +255,39 @@ func TestTLSCustomCertList(t *testing.T) {
...
 
 func TestTLSCustomCertUpdate(t *testing.T) {
+	var content string
 	args := testutil.Args
 	scenarios := []testutil.TestScenario{
 		{
 			Name:      "validate missing --cert-blob flag",
 			Args:      args("tls-custom certificate update --id example"),
-			WantError: "required flag --cert-blob not provided",
+			WantError: "neither --cert-path or --cert-blob provided, one must be provided",
```
Integralist, line 264: Feels like this test case can be removed in favour of `"validate missing --cert-blob and --cert-path flags"` below.

Integralist (approved):

pkg/commands/tls/custom/certificate/certificate_test.go
rdoherty-fastly: Done!

Integralist:
Feel free to merge whenever you're ready

rdoherty-fastly:
im_ready_game_of_thrones

PR #1215, opened 2024-05-30
Add option to pass in private key via a file path for `tls-custom private-key create` command via `--key-path`

rdoherty-fastly:
Add option to pass in private key via a file path for `tls-custom private-key create` command via `--key-path`

This option is mutually exclusive with --key.

This mirrors the functionality for the `tls-custom certificate create` command that also now accepts a file path to the certificate.


kpfleming:
At some point it might be worth refactoring this common code, since it appears in a number of places now (some sort of "content or path" helper).

rdoherty-fastly:
Agreed @kpfleming , thanks for merging!

PR #1226, opened 2024-06-12
fix(sso): re-auth on profile switch + support MAUA

Integralist:
When setting up multiple SSO profiles, the refresh token flow is a bit broken. 

The previous refresh token is immediately invalidated when creating the next profile and as a result always forces a re-auth.

The changes here hope to address these issues by forcing a re-auth whenever the user switches profiles.


pkg/commands/sso/root.go
```diff
@@ -92,7 +109,12 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 	// that another session (from a different profile) doesn't cause unexpected
 	// errors for the user flow. This forces a re-auth.
 	if c.InvokedFromProfileCreate || c.InvokedFromProfileUpdate || ForceReAuth {
+		c.Globals.AuthServer.SetParam("prompt", "login select_account")
+	}
+	if c.InvokedFromProfileSwitch {
 		c.Globals.AuthServer.SetParam("prompt", "login")
+		c.Globals.AuthServer.SetParam("login_hint", c.ProfileSwitchEmail)
+		c.Globals.AuthServer.SetParam("account_hint", c.ProfileSwitchCustomerID)
```
phamann, line 117: Maybe we should check whether these are empty before setting? Asusming in the tranistion there may be a possibility they may be empty. However, I don't think there is a harm in sending empty values.

pkg/commands/sso/root.go
```diff
@@ -92,7 +109,12 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
 	// that another session (from a different profile) doesn't cause unexpected
 	// errors for the user flow. This forces a re-auth.
 	if c.InvokedFromProfileCreate || c.InvokedFromProfileUpdate || ForceReAuth {
+		c.Globals.AuthServer.SetParam("prompt", "login select_account")
+	}
+	if c.InvokedFromProfileSwitch {
```
phamann, line 114: We only want to send `select_account` on profile create, all other times (including update) we assume we know the profile email and account id. So i'd change these conditionals too:

```suggestion
	if c.InvokedFromProfileCreate || ForceReAuth {
		c.Globals.AuthServer.SetParam("prompt", "login select_account")
	}
	if c.InvokedFromProfileUpdate || c.InvokedFromProfileSwitch {
```

pkg/commands/sso/root.go
```diff
@@ -188,6 +219,55 @@ func (c *RootCommand) identifyProfileAndFlow() (profileName string, flow Profile
...
+				Key:   "Accept",
+				Value: "application/json",
+			},
+			{
+				Key:   "User-Agent",
+				Value: useragent.Name,
+			},
+		},
+		Method: http.MethodGet,
+		Path:   "/verify",
```
phamann, line 239: I'd personally suggest using `/current_customer` instead, which has less data than than `/verify` and the field's you need `id` and `name` in the response. 

https://www.fastly.com/documentation/reference/api/account/customer/#get-logged-in-customer

phamann (changes requested):
LGTM other than a couple of minor nits.

pkg/commands/sso/root.go
```diff
@@ -188,6 +223,55 @@ func (c *RootCommand) identifyProfileAndFlow() (profileName string, flow Profile
...
+	})
+	if err != nil {
+		c.Globals.ErrLog.Add(err)
+		return fmt.Errorf("error executing verify API request: %w", err)
+	}
+
+	var response CurrentCustomerResponse
+	if err := json.Unmarshal(data, &response); err != nil {
+		c.Globals.ErrLog.Add(err)
+		return fmt.Errorf("error decoding verify API response: %w", err)
```
phamann, line 260: Nit: the error message is no longer correct:

```suggestion
		return fmt.Errorf("error executing current_customer API request: %w", err)
	}

	var response CurrentCustomerResponse
	if err := json.Unmarshal(data, &response); err != nil {
		c.Globals.ErrLog.Add(err)
		return fmt.Errorf("error decoding current_customer API response: %w", err)
```

phamann (approved):
LGTM, other than very minor error message nit.

PR #1230, opened 2024-06-14
fix(sso): correct the behaviour for direct sso invocation

Integralist:


pkg/commands/sso/root.go
```diff
@@ -108,17 +110,36 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
...
 			c.Globals.AuthServer.SetParam("account_hint", c.ProfileSwitchCustomerID)
 		}
+	default:
+		// Handle `fastly sso` being invoked directly.
+		p := profile.Get(profileName, c.Globals.Config.Profiles)
+		if p == nil {
+			err := fmt.Errorf(profile.DoesNotExist, profileName)
+			c.Globals.ErrLog.Add(err)
+			return err
+		}
```
phamann, line 131: This seems a bit harsh, couple of options to consider:

- Easiest: update the `profile.DoesNotExist` error message to also include a note on how they create one, currently it just says `the profile '%s' does not exist`, i.e. can we add `to create one, run: fastly profile create`
- Instead of erroring here, could we write a messaging saying that a profile doesn't exist and that we'll create one for them?
- Best: the above, but with a `Y/n` prompt

pkg/commands/sso/root.go
```diff
@@ -108,17 +110,36 @@ func (c *RootCommand) Exec(in io.Reader, out io.Writer) error {
...
+		c.Globals.AuthServer.SetParam("prompt", "login")
+		c.Globals.AuthServer.SetParam("login_hint", p.Email)
+		c.Globals.AuthServer.SetParam("account_hint", p.CustomerID)
+		// IMPORTANT: We must make the specified profile the default.
+		// If we don't, then the next command run will fail to check the token.
+		// Because the current profile will not be the active session.
+		// And we don't want to have to force the user to have re-auth again.
+		// Really, `fastly sso` should be a hidden command.
+		// And all users should use the `fastly profile ...` subcommands.
+		c.ProfileDefault = true
```
phamann, line 141: Whilst I understand **_why_** we need to do this, it may be confusing to the user if we don't explicitly tell them we're doing this. Could we leave a message eitehr before or after the command saying that `Profile <foo> is now the default profile for all commands`.

Or we update the help text for this root command saying it is used to manage your default SSO profile?

PR #1232, opened 2024-06-14
fix(sso): update output to reflect default profile behaviour

Integralist:
> **NOTE:** Remediation errors that explain how to create a profile were added in a separate PR (already merged).

This PR ensures we error early if `fastly sso` is called directly with an invalid profile name.

It also updates the output to reflect that the profile specified will become the new default profile.


phamann (approved):
Much cleaner 👌🏻

PR #1239, opened 2024-07-04
feat: add debug-mode around all network requests

Integralist:
There was a bug recently which we were only able to identify and fix, by dumping a related HTTP request/response.

We have `--debug-mode` but currently it only passes that information to go-fastly which dumps the req/resp.

This PR adds the same logic as found in go-fastly and ensures all network requests are wrapped in conditional branches checking for debug mode being enabled.


kailan (approved):
This is so helpful – thank you!

jsocol:
We also got a failure in `pkg/config/TestUseStatic` reading [`config-legacy.toml`](https://github.com/fastly/cli/blob/main/pkg/config/config_test.go#L208) which hasn't been touched in 2 years. Some kind of change to the MacOS runners recently?

kpfleming:
This was the `tomlq` change; I'll update this branch and the tests should pass.

kpfleming:
Tests are now passing, I'll merge this.

PR #1244, opened 2024-07-09
fix(update): Ensure that the CLI binary will be executable after an update

### LINKED ISSUE #1168: `fastly update` does not make the new binary executable on Linux
**Version**
(this is the updated version:)
Fastly CLI version v10.8.8 (6fc2fc1a)
Built with go version go1.22.1 linux/amd64 (2024-03-19)
Viceroy version: viceroy 0.9.4

**What happened**

I saw:
```
A new version of the Fastly CLI is available.
Current version: 10.8.6
Latest version: 10.8.8
Run `fastly update` to get the latest version.
```
and ran:
```
% fastly update
✓ Updating versioning information

Current version: 10.8.6
Latest version: 10.8.8

✓ Fetching latest release
✓ Replacing binary

SUCCESS: Updated /var/home/ian/bin/fastly to 10.8.8.
```
but then when I tried using the `fastly` command again I got:
```
zsh: permission denied: fastly
```
because the newly downloaded and installed `~/bin/fastly` didn't have execute permissions. Somewhere in `cli/pkg/commands/update/root.go` you probably need to add an `os.Chmod()`.

kpfleming:
When the downloaded CLI binary cannot be 'moved' into position (i.e. when it is located on a different filesystem from the target location), it is copied, but the permissions need to be set after the copy so that it will be executable.

Closes #1168.

### DIFF
--- a/pkg/commands/update/root.go
+++ b/pkg/commands/update/root.go
@@ -130,6 +130,15 @@ func (c *RootCommand) Exec(_ io.Reader, out io.Writer) error {
 				})
 				return fmt.Errorf("error 'copying' latest binary in place: %w (following an error 'moving': %w)", err, renameErr)
 			}
+
+			// G302 (CWE-276): Expect file permissions to be 0600 or less
+			// gosec flagged this:
+			// Disabling as the file was not executable without it and we need all users
+			// to be able to execute the binary.
+			/* #nosec */
+			if err := os.Chmod(currentBin, 0o755); err != nil {
+				return fmt.Errorf("failed to modify permissions on after 'copying' latest binary: %w", err)
+			}
 		}
 		return nil
 	})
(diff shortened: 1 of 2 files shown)


pkg/commands/update/root.go
```diff
@@ -130,6 +130,15 @@ func (c *RootCommand) Exec(_ io.Reader, out io.Writer) error {
...
 				return fmt.Errorf("error 'copying' latest binary in place: %w (following an error 'moving': %w)", err, renameErr)
 			}
+
+			// G302 (CWE-276): Expect file permissions to be 0600 or less
+			// gosec flagged this:
+			// Disabling as the file was not executable without it and we need all users
+			// to be able to execute the binary.
+			/* #nosec */
+			if err := os.Chmod(currentBin, 0o755); err != nil {
+				return fmt.Errorf("failed to modify permissions on after 'copying' latest binary: %w", err)
```
Integralist, line 140:
```suggestion
				return fmt.Errorf("failed to modify permissions after 'copying' latest binary: %w", err)
```

Integralist (approved):
One minor comment, otherwise LGTM

PR #1247, opened 2024-07-12
refactor: Simplify declaration and execution of test scenarios

kpfleming:
Adds an 'Arg' field to the TestScenario structure, to hold a simple string of arguments instead of a slice of individual arguments.

Adds a 'RunScenarios' function to the testutil package which is able to execute all of the tests in a slice of TestScenario structs. This function accepts an additional slice (of strings) which indicate the command name(s) to be prepended to the arguments supplied in each scenario, so that they do not need to be repeated in the scenario declarations. Finally, it splits the 'Arg' field into individual arguments so that the scenario definition can supply a simple string.

Converts serviceversion_test.go to use the new features.

Converts .tmpl/test.go to use the new features, and also fixes some problems with the test scenarios declared there.

The remaining tests will be converted to the new features once this PR has been merged.

### DIFF
--- a/.tmpl/test.go
+++ b/.tmpl/test.go
@@ -17,7 +17,6 @@ func TestCreate(t *testing.T) {
 	scenarios := []testutil.TestScenario{
 		{
 			Name:      "validate missing --version flag",
-			Args:      "",
 			WantError: "error parsing arguments: required flag --version not provided",
 		},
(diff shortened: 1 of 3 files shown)


Integralist (approved):
Other than one minor comment, looks fine to me.

PR #1254, opened 2024-07-26
logtail: add --timestamps flag

joeshaw:
This will print the timestamp of the request start time with each log
message.

### DIFF
--- a/pkg/commands/logtail/root.go
+++ b/pkg/commands/logtail/root.go
@@ -448,10 +448,10 @@ func (c *RootCommand) printLogs(out io.Writer, logs []Log) {
 
 		for _, l := range filtered {
 			if c.cfg.printTimestamps {
-				fmt.Fprintf(out, "%s | %s\n", l.RequestStartFromRaw().UTC().Format(time.RFC3339), l.String())
-			} else {
-				fmt.Fprintln(out, l.String())
+				fmt.Fprint(out, l.RequestStartFromRaw().UTC().Format(time.RFC3339))
+				fmt.Fprint(out, " | ")
 			}
+			fmt.Fprintln(out, l.String())
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/commands/logtail/root.go
```diff
@@ -446,7 +447,11 @@ func (c *RootCommand) printLogs(out io.Writer, logs []Log) {
 		filtered := filterStream(c.cfg.stream, logs)
 
 		for _, l := range filtered {
-			fmt.Fprintln(out, l.String())
+			if c.cfg.printTimestamps {
+				fmt.Fprintf(out, "%s | %s\n", l.RequestStartFromRaw().UTC().Format(time.RFC3339), l.String())
+			} else {
+				fmt.Fprintln(out, l.String())
+			}
```
joeshaw, line 454: I wasn't sure if it was better to format the string here, or add a `StringWithTimestamp` method on the `Log`.  Happy to do whichever you think is best.

kpfleming:
I was going to suggest enhancing the unit tests for `logtail` to test this new parameter... but there are no command-line unit tests for `logtail` :-(

pkg/commands/logtail/root.go
kpfleming: Since `Log` is only used in this module, and the output format is not really part of that structure, I have no objections to it being done here. If Go had a ternary operator this would be a good place to use it, but...

Maybe one small improvement would be to just declare a string variable that is either empty or contains the formatted timestamp prefix (with separator) and then call Fprintf only once, including both that string and `l.String()`.

Integralist:
CI failed...

```
FAIL: TestInit/with_default_name_inferred_from_directory
```

pkg/commands/logtail/root.go
joeshaw: Variation on this which should limit formatting & allocations done in 681638d

joeshaw:
> CI failed...

I suspect this was a flake.  It's unrelated to the change in any case.  The additional commit has kicked off a rebuild, hopefully it'll clear up.

PR #1264, opened 2024-08-01
fix(all commands): --service-name flag should have priority. 

### LINKED ISSUE #1261: [BUG] accidental service deletion when PR is cleaned up for failed deployment
**Version**

```
Fastly CLI version v10.12.3 (f0fe32f4)
Built with go version go1.22.4 linux/amd64 (2024-07-30)
```

**What happened**

**Command Ran:**
```bash
fastly service delete --quiet --service-name @resf/rocky-linux-docs-73 --force --token ***
```

**Expected:**
I expected the service identified by the `--service-name` argument to be deleted.

**Actual:**
Despite providing `--service-name`, the `fastly service delete` command deleted the service ID specified in the `fastly.toml`. This led to the accidental deletion of the main service instead of the intended preview service.

**Possible Fixes:**
1. Modify the Fastly CLI to verify the named service existence before attempting deletion.
2. Modify the Fastly CLI to raise an error if inconsistent instructions are given (i.e., `--service-name` should never fall back to a stored value).

ref: https://github.com/fastly/compute-actions/issues/52

kpfleming:
When the --service-name flag was added it was given the lowest
priority for obtaining a service ID. If a user supplied this flag but
also supplied --service-id, or had a fastly.toml with a service ID in
it, or had the FASTLY_SERVICE_ID environment variable set, the
--service-name flag would be ignored. This resulted in a user
accidentally deleting a service because the flag was ignored.

This commit changes the priority so that --service-name and
--service-id have equal (top) priority, and also ensures that
supplying both will result in an error. It also adds a unit test for
the argparser.ServiceID() function which implements this logic.

Closes #1261.

### DIFF
--- a/.tmpl/create.go
+++ b/.tmpl/create.go
@@ -42,7 +42,7 @@ func NewCreateCommand(parent argparser.Registerer, globals *config.Data, data ma
 	c.RegisterFlag(argparser.StringFlagOpts{
 		Action:      c.serviceName.Set,
 		Name:        argparser.FlagServiceName,
-		Description: argparser.FlagServiceDesc,
+		Description: argparser.FlagServiceNameDesc,
 		Dst:         &c.serviceName.Value,
 	})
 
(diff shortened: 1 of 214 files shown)


jsocol (approved):
lgtm! appreciate the clarity of `--service-id` vs `--service-name`

PR #1274, opened 2024-08-05
More test case refactoring, new features in TestScenario

kpfleming:


pkg/commands/compute/metadata_test.go
```diff
@@ -2,182 +2,174 @@ package compute_test
...
+
+				var testFile config.File
+				unmarshalErr := toml.Unmarshal(data, &testFile)
+				if unmarshalErr != nil {
+					t.Error(unmarshalErr)
+				}
+
+				testutil.AssertString(t, "enable", testFile.WasmMetadata.BuildInfo)
+				testutil.AssertString(t, "enable", testFile.WasmMetadata.MachineInfo)
+				testutil.AssertString(t, "enable", testFile.WasmMetadata.PackageInfo)
```
kpfleming, line 64: @Integralist This code (and the other three copies of it) previously used `opts.Config.Read` to read the modified TOML file, and then validate that it had the desired configuration.

However, after refactoring the code into this new structure, that did not work: the values in the `opts.Config.WasmMetadata` struct were wrong for three of the four test cases. I added logging to display the actual contents of the file, and also directly unmarshaled it into another `config.File` object, and both showed the correct (expected) contents.

In the end I went with direct unmarshaling, since this test is really testing the `compute manifest` command's behavior and not the behavior of the `config` package.

Integralist: SGTM 👍🏻

pkg/commands/acl/acl_test.go
```diff
@@ -10,8 +10,6 @@ import (
 	"github.com/fastly/cli/pkg/testutil"
 )
 
-const ()
```
Integralist, line 13: lol wat!

kpfleming: Yeah, that was a leftover from a `sed` script which removed the only line left in the block but didn't remove the block!

pkg/testutil/scenarios.go
```diff
@@ -19,19 +19,34 @@ import (
...
-	WantOutputs     []string
+	// scenario name (appears in output when tests are executed)
+	Name string
+	// if this is supplied, NewEnv will be used to create a temporary environment
+	// for the scenario
+	NewEnv          *NewEnvConfig
 	PathContentFlag *PathContentFlag
-	SetEnv          *map[string]string
+	// set one (or more) environment variables during the scenario's execution
+	SetEnv *map[string]string
```
Integralist, line 39: The `map` type is already passed-by-pointer so you don't need to take the address of a map type when passing it around as it's already a pointer.

Integralist:
I'll leave me review comments any way so you can circle back around to them at some point.

pkg/testutil/scenarios.go
```diff
@@ -42,14 +57,23 @@ type PathContentFlag struct {
 	Content func() string
 }
 
+// NewEnvConfig provides the details required to setup a temporary test
```
Integralist, line 60: I'm not a fan of the `New` prefix on the struct name as `New` is typically used in go to indicate a constructor function. This struct should just be `EnvConfig` as that's what it represents/contains details of (i.e. an environment config).

pkg/testutil/scenarios.go
```diff
@@ -42,14 +57,23 @@ type PathContentFlag struct {
...
+// environment, and optionally a function to run which accepts the
+// environment directory and can modify fields in the TestScenario
+type NewEnvConfig struct {
+	EnvOpts      *EnvOpts
+	EditScenario func(*TestScenario, string)
+}
+
 // RunScenario executes a TestScenario struct.
-// The Args field of the scenario is prepended with the content of the 'command'
+// The Arg field of the scenario is prepended with the content of the 'command'
```
Integralist, line 69: I'm not a fan of `Arg` (singular) as it suggests a _single_ argument, rather than potentially numerous flags/args within the string itself (e.g. `Arg:  "-s 123 --version 1 --package pkg/package.tar.gz --autoclone",`).

I understand it was named this way because `Args` was already used. But I would suggest we switch from `Arg` to `Args` once the latter is no longer used within the CLI code base.

pkg/testutil/scenarios.go
```diff
@@ -19,19 +19,34 @@ import (
...
-	Name            string
-	Stdin           []string
-	WantError       string
-	WantOutput      string
-	WantOutputs     []string
+	// scenario name (appears in output when tests are executed)
+	Name string
+	// if this is supplied, NewEnv will be used to create a temporary environment
+	// for the scenario
+	NewEnv          *NewEnvConfig
```
Integralist, line 36: As per an earlier comment, I think this should just be...

```go
EnvConfig *EnvConfig
```

...or even just...

```go
Env *Env
```

If you use the latter `Env`, then you might want to change `SetEnv` to `EnvVars` (although changing it to `EnvVars` regardless might be better as a field name (as `Set` is a verb in this context and thus suggests a function).

pkg/commands/compute/metadata_test.go
```diff
@@ -2,182 +2,174 @@ package compute_test
...
+							Dst: "config.toml",
+						},
+					},
+				},
+				EditScenario: func(scenario *testutil.TestScenario, rootdir string) {
+					scenario.ConfigPath = filepath.Join(rootdir, "config.toml")
+				},
+			},
+			WantOutput: "SUCCESS: configuration updated",
+			Validator: func(t *testing.T, _ *testutil.TestScenario, opts *global.Data, _ error, _ bytes.Buffer) {
```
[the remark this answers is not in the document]
kpfleming: `RunScenario` is already validating that the `err` is empty by virtue of the fact that `WantError` is empty in this scenario. If `err` is non-empty, the `Validator` won't even be called... which really means there's not really any need to have the `error` as an argument. I'll remove it.

pkg/testutil/scenarios.go
kpfleming: Great minds think alike... I had planned to do just that, and also contemplated using something a bit 'random' for the temporary field name, but decided not to because it wasn't going to be replaced within a day or two.

pkg/testutil/scenarios.go
```diff
@@ -19,19 +19,34 @@ import (
 type TestScenario struct {
 	API mock.API
 	Arg string
-	// will be removed once all users are using RunScenarios()
-	Args            []string
-	ConfigPath      string
-	ConfigFile      *config.File
-	DontWantOutput  string
+	// FIXME: remove once all users are using RunScenario() or RunScenarios()
+	Args []string
```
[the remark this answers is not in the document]
kpfleming: I'll mark all of these as 'resolved' as the requested changes are in #1276.

PR #1275, opened 2024-08-06
Support configuring connection keepalive parameters

ulyssa:
This is a draft PR that adds support for creating and updating backends with the different keepalive parameters:

- `keepalive_time`, which is how long HTTP keepalive connections will be kept when they are idle; I've made the flag for this `--http-ka-time` to make it easier to differentiate with the `--tcp-ka-*` flags.
- `tcp_keepalive_enable`, which determines whether TCP keepalives are turned on on connections.
- `tcp_keepalive_interval` which determines how long to wait in between sending each TCP keepalive probe.
- `tcp_keepalive_probes` which determines up to how many probes to send before considering the connection dead.
- `tcp_keepalive_time` which determines how long to wait after sending the last data packet before starting to send TCP keepalive probes.


ulyssa:
This will depend on pulling in fastly/go-fastly#537 once it has a tag we can put in `go.mod`.

Integralist:
@kpfleming this PR requires a new go-fastly release that contains https://github.com/fastly/go-fastly/pull/537

kpfleming:
This will need tests added in `backend_test.go` as well.

ulyssa:
I added tests in f1fda72, which help me realized that I missed accounting for the new fields in `describe`/`list`, which I've now done in 62f5074.

Integralist:
The build is failing...

```
Error: pkg/text/backend.go:41:7: b.TCPKeepAliveEnable undefined (type *fastly.Backend has no field or method TCPKeepAliveEnable)
Error: pkg/text/backend.go:44:68: b.TCPKeepAliveEnable undefined (type *fastly.Backend has no field or method TCPKeepAliveEnable)
Error: pkg/text/backend.go:46:68: b.TCPKeepAliveIntvl undefined (type *fastly.Backend has no field or method TCPKeepAliveIntvl)
Error: pkg/text/backend.go:47:66: b.TCPKeepAliveProbes undefined (type *fastly.Backend has no field or method TCPKeepAliveProbes)
Error: pkg/text/backend.go:48:67: b.TCPKeepAliveTime undefined (type *fastly.Backend has no field or method TCPKeepAliveTime)
```

kpfleming:
> The build is failing...
> 
> ```
> Error: pkg/text/backend.go:41:7: b.TCPKeepAliveEnable undefined (type *fastly.Backend has no field or method TCPKeepAliveEnable)
> Error: pkg/text/backend.go:44:68: b.TCPKeepAliveEnable undefined (type *fastly.Backend has no field or method TCPKeepAliveEnable)
> Error: pkg/text/backend.go:46:68: b.TCPKeepAliveIntvl undefined (type *fastly.Backend has no field or method TCPKeepAliveIntvl)
> Error: pkg/text/backend.go:47:66: b.TCPKeepAliveProbes undefined (type *fastly.Backend has no field or method TCPKeepAliveProbes)
> Error: pkg/text/backend.go:48:67: b.TCPKeepAliveTime undefined (type *fastly.Backend has no field or method TCPKeepAliveTime)
> ```

I haven't made a go-fastly release yet, so unless someone else did this will continue. That's on my agenda for today if nobody else does it first.

kpfleming:
go-fastly 9.8.0 has been released.

pkg/commands/backend/describe.go
```diff
@@ -127,6 +127,15 @@ func (c *DescribeCommand) print(out io.Writer, b *fastly.Backend) error {
 	fmt.Fprintf(out, "Min TLS version: %v\n", fastly.ToValue(b.MinTLSVersion))
 	fmt.Fprintf(out, "Max TLS version: %v\n", fastly.ToValue(b.MaxTLSVersion))
 	fmt.Fprintf(out, "SSL ciphers: %v\n", fastly.ToValue(b.SSLCiphers))
+	fmt.Fprintf(out, "HTTP KeepAlive Timeout: %v\n", fastly.ToValue(b.KeepAliveTime))
+	if b.TCPKeepAliveEnable == nil {
+		fmt.Fprintf(out, "TCP KeepAlive Enabled:\n")
+	} else {
+		fmt.Fprintf(out, "TCP KeepAlive Enabled: %v\n", fastly.ToValue(b.TCPKeepAliveEnable))
+	}
```
kpfleming, line 135: I think this is likely to be confusing for users, as "TCP KeepAlive Enabled:" with no value is going to be interpreted as "enabled".

If it is necessary for the field to always appear even if the value is `nil`, then the label should be "TCP KeepAlive" and the value should be "unset" or something like that. If it's OK for the field to not appear at all if the value is `nil`, then that would be preferable.

Whichever change is made, the same change should be made in `pkg/text/backend.go`.

pkg/commands/backend/create.go
```diff
@@ -203,6 +214,34 @@ func (c *CreateCommand) Exec(_ io.Reader, out io.Writer) error {
...
+
+		switch c.tcpKaEnable.Value {
+		case "true":
+			tcpKaEnable = true
+		case "false":
+			tcpKaEnable = false
+		default:
+			err := errors.New("'tcp-ka-enable' flag must be one of the following [true, false]")
+			c.Globals.ErrLog.Add(err)
+			return err
```
kpfleming, line 228: The unit test should have a scenario for this, to ensure that invalid values are rejected; it's OK to add three scenarios to cover all three possibilities.

pkg/commands/backend/update.go
```diff
@@ -236,6 +247,36 @@ func (c *UpdateCommand) Exec(_ io.Reader, out io.Writer) error {
...
+
+		switch c.TCPKaEnable.Value {
+		case "true":
+			tcpKaEnable = true
+		case "false":
+			tcpKaEnable = false
+		default:
+			err := errors.New("'tcp-ka-enable' flag must be one of the following [true, false]")
+			c.Globals.ErrLog.Add(err)
+			return err
```
kpfleming, line 261: Same comment as for `describe`; the unit test should exercise all three paths in this switch statement.

kpfleming (changes requested):
Very very close :-)

pkg/commands/backend/create.go
ulyssa: Done in 208e61da518bc6cd44cf0a7563eb0449b1aff2fb!

pkg/commands/backend/describe.go
ulyssa: I've made it `unset` in b899fb9bc41cec92c81cdf66609ddb4627d8b8a9 so that it's always present in case someone greps for it or is scanning output in a support ticket.

pkg/commands/backend/update.go
ulyssa: Done in 208e61da518bc6cd44cf0a7563eb0449b1aff2fb!

kpfleming (approved):

PR #1276, opened 2024-08-06
refactor: Apply review feedback from PR 1274.

kpfleming:


pkg/testutil/scenarios.go
```diff
@@ -95,8 +114,8 @@ func RunScenario(t *testing.T, command []string, scenario TestScenario) {
 			}
 
 			// Create test environment
-			scenario.NewEnv.EnvOpts.T = t
-			rootdir = NewEnv(*scenario.NewEnv.EnvOpts)
+			scenario.Env.EnvOpts.T = t
```
Integralist, line 117: I can see we have some stutter here, let's update `EnvOpts` to be `Opts` so we get a nicer `scenario.Env.Opts`.

ulyssa (approved):

PR #1284, opened 2024-08-13
refactor: Rename various parts of testutil package.

kpfleming:
* Restore 'Args' name in scenario struct.

* Rename TestScenario to CLIScenario to clarify that it is only applicable to CLI-style tests.

* Rename RunScenarios and RunScenario for the same reason.

* Rename Args function to SplitArgs to better describe its purpose.

### DIFF
--- a/pkg/testutil/args.go
+++ b/pkg/testutil/args.go
@@ -19,7 +19,7 @@ import (
 
 var argsPattern = regexp.MustCompile("`.+`")
 
-// Args is a simple wrapper function designed to accept a CLI command
+// SplitArgs is a simple wrapper function designed to accept a CLI command
 // (including flags) and return it as a slice for consumption by app.Run().
 //
 // NOTE: One test file (TestBigQueryCreate) passes RSA content inline into the
@@ -35,7 +35,7 @@ var argsPattern = regexp.MustCompile("`.+`")
 // surrounded by backticks with --- and then splits the resulting string by
 // spaces. Afterwards if there was a backtick matched, then we re-insert the
 // backticked content into the slice where --- is found.
-func Args(args string) []string {
+func SplitArgs(args string) []string {
 	var backtickMatch []string
 
 	if strings.Contains(args, "`") {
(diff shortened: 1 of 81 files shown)


pkg/testutil/args.go
```diff
@@ -35,7 +35,7 @@ var argsPattern = regexp.MustCompile("`.+`")
 // surrounded by backticks with --- and then splits the resulting string by
 // spaces. Afterwards if there was a backtick matched, then we re-insert the
 // backticked content into the slice where --- is found.
-func Args(args string) []string {
+func SplitArgs(args string) []string {
```
cee-dub, line 38: Drive-by review, feel free to ignore: this logic seems extraneous, straining to make this convenience available where simply using a slice literal would be simpler.

If this were my codebase, I might even go as far as having the function error when the passed string contained a quote or backtick, with the error message asking the programmer to use `[] string` to pass complex arguments.

pkg/testutil/scenarios.go
```diff
@@ -15,22 +15,18 @@ import (
...
 	API mock.API
-	// Arg is the input arguments for the command to execute (not
+	// Args is the input arguments for the command to execute (not
 	// including the command names themselves).
-	Arg string
-	// Args is the input arguments for the command to execute.
-	// DEPRECATED: Args will be removed once all users are using
-	// RunScenario() or RunScenarios().
-	Args []string
+	Args string
```
cee-dub, line 29: Part 2 🚗: This might have to change back to `[]string` to agree with `os.Cmd`, and that probably means using `SplitArgs()` in more places if you were to use my feedback.

pkg/testutil/args.go
kpfleming: I found it a bit odd too, but it predates my involvement with the codebase and isn't an important thing to change (I've been focusing on making test scenarios easier to write and maintain). For all I know the scenario which needs this special handling may no longer be present!

cee-dub: Fair point! Thanks for taking on this maintenance and polishing this important codebase.

Integralist (approved):

PR #1285, opened 2024-08-13
chore: Use full Go version in go.mod.

kpfleming:
As of Go 1.21 this is expected to be a full 1.N.P version string, not just 1.N (and CodeQL is complaining about it).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/fastly/cli
 
-go 1.22
+go 1.22.0
 
 require (
 	github.com/Masterminds/semver/v3 v3.2.1


kpfleming:
> It is probably better in this case to use 1.22.6, as that's the latest released patch

I was curious about that, and assumed that since we are declaring the language feature set that we expect/have tested against that the .0 version would be fine since patch releases wouldn't change the feature set.



> We could update our `go.mod` to become compatible by adding `/v10` like so: `module github.com/fastly/cli/v10`

Interesting... I would be quite surprised if anything is importing this as a module into another project, so that should be non-disruptive, unless all of the internal references to `github.com/fastly/cli/...` would also need `v10` injected into them.

cee-dub:
> all of the internal references to `github.com/fastly/cli/...` would also need `v10` injected into them

They do, e.g. `import "github.com/fastly/cli/pkg/app"` -> `import "github.com/fastly/cli/v10/pkg/app"`

As an alternative idea, we could begin using a vanity URL and 0.X version at the same time.

kpfleming:
OK, since it's unlikely anyone is importing this as a module today (given the versioning problem it not even work), let's create an issue in this repo for improving the module naming/versioning as a separate task, so we can have a more complete discussion of it.

PR #1327, opened 2024-10-18
fix(compute build): Emit warning if Rust 1.82.0 (or higher) are used for the build.

kpfleming:
Rust 1.82.0 introduced a change to how 'reference types' are handled in WebAssembly and the Compute platform is not yet compatible with that change.

### DIFF
--- a/.fastly/config.toml
+++ b/.fastly/config.toml
@@ -1,4 +1,4 @@
-config_version = 6
+config_version = 7
 
 [fastly]
 account_endpoint = "https://accounts.fastly.com"
--- a/.fastly/config.toml
+++ b/.fastly/config.toml
@@ -18,7 +18,7 @@ toolchain_constraint = ">= 1.21"           # Go toolchain constraint for use wit
 toolchain_constraint_tinygo = ">= 1.18"    # Go toolchain constraint for use with TinyGo.
 
 [language.rust]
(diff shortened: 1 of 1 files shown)


Integralist (commented):
I think we need to bump the `config_version` so as to ensure existing users (who already have a local config) will have this changed merged into their existing config.

See https://github.com/fastly/cli/blob/main/DEVELOPMENT.md#fastlyconfigtoml

> If the config has changed in any way, then you (the CLI developer) should ensure the config_version number is bumped before publishing a new CLI release. This is because when the user updates to that new CLI version and the invoke the CLI, the CLI will identify a mismatch between the user's local config version and the embedded config version. This will cause the embedded config to be merged with the local config and consequently the user's config will be updated to include the new fields.

Integralist (approved):

kpfleming:
In my local testing, without bumping `config_version`, the `config.toml` file was updated to have the new `toolchain_constraint` value. I'm not able to track down the code path which caused this at the moment as I have to head to the airport, but I'd prefer not to bump `config_version` if it's not necessary and it does not seem to be necessary.

kpfleming:
I'm closing this PR as it will not be necessary; the Compute fleet is being updated to be compatible with WASM packages produced by Rust 1.82 and that should be complete within the next few days.

PR #1338, opened 2024-11-07
breaking(products): Remove support for NGWAF product.

kpfleming:
The support added for the NGWAF product in version 10.15.0 did not work, and is being removed so that it can be redesigned. Since the feature was non-operational, no users of the CLI could have relied on it, so the major version number is not being incremented in spite of the breaking nature of the change.


philippschulte (approved):
Thanks

PR #1339, opened 2024-11-07
fix(profile/token): 'profile token' command must check the validity of the stored token.

kpfleming:
The 'fastly profile token' command did not check the validity (expiration) of the stored token, which meant that it would emit an invalid token if the stored session token (and refresh token) had expired. This PR changes the behavior so that the validity of the token is checked as it is for all commands which actually make use of the token.

Because this command is most often used in a non-interactive situation (in a script or some other automation), expired and soon-to-expire tokens just return errors, they don't prompt for re-authentication.

### DIFF
--- a/pkg/commands/profile/token.go
+++ b/pkg/commands/profile/token.go
@@ -3,12 +3,15 @@ package profile
 import (
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/config"
 	fsterr "github.com/fastly/cli/pkg/errors"
 	"github.com/fastly/cli/pkg/global"
 	"github.com/fastly/cli/pkg/profile"
 	"github.com/fastly/cli/pkg/text"
+	fsttime "github.com/fastly/cli/pkg/time"
 )
 
 // TokenCommand represents a Kingpin command.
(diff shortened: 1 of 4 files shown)


pkg/app/run.go
```diff
@@ -345,8 +346,11 @@ func configureKingpin(data *global.Data) *kingpin.Application {
 //
 // Finally, we check if there is a profile override in place (e.g. set via the
 // --profile flag or using the `profile` field in the fastly.toml manifest).
-func processToken(cmds []argparser.Command, data *global.Data) (token string, tokenSource lookup.Source, err error) {
-	token, tokenSource = data.Token()
+func processToken(commandName string, cmds []argparser.Command, data *global.Data) (token string, tokenSource lookup.Source, err error) {
+	// if the command is 'profile token', then the token can only
```
kpfleming, line 350: This is necessary because if a token is found on the command line or in the environment, no validity-checking will be performed... and then the `profile token` command will output the possibly-expired token from the profile.

Verified in my testing.

kpfleming:
I'm not 100% sold on this implementation, and will be working on an alternative.

pkg/commands/profile/token.go
```diff
@@ -26,32 +29,42 @@ func NewTokenCommand(parent argparser.Registerer, g *global.Data) *TokenCommand
 	return &c
 }
 
+// By default tokens must be valid for at least 5 minutes to be
+// considered valid.
+const defaultTokenTTL int64 = 5 * 60
```
phamann, line 34: The more golang idomatic thing to do here would be to store a `time.Duration`:

```suggestion
const defaultTokenTTL time.Duration = 5 * time.Minute
```

This also make its intention and value more clear to readers.

You can then change the signature of `checkTokenValidity()` to accept a `Duration`

pkg/commands/profile/token.go
```diff
@@ -60,3 +73,21 @@ func (c *TokenCommand) Exec(_ io.Reader, out io.Writer) (err error) {
 		Remediation: fsterr.ProfileRemediation,
 	}
 }
+
+func checkTokenValidity(profileName string, p *config.Profile, ttl int64) (err error) {
```
phamann, line 77: Per my message above, this can/should accept a `time.Duration` for the `ttl`.

```suggestion
func checkTokenValidity(profileName string, p *config.Profile, ttl time.Duration) (err error) {
```

phamann (changes requested):
LGTM other than minor nit around the `ttl`

pkg/commands/profile/token.go
kpfleming: Thanks! Great suggestion.

phamann (approved):
LGTM

PR #1340, opened 2024-11-08
feat(profile/token): Allow user to specify how long token must be valid.

kpfleming:
The default token validity period for 'fastly profile token' is 5 minutes, but users may want to ensure that the token is valid for a longer period (or only a shorter period). This PR adds 'fastly profile token --ttl=&lt;period&gt;', allowing the user to specify the validity period in seconds, minutes, or hours (or some combination).


philippschulte (approved):
lgtm, thank you!

PR #1349, opened 2024-11-18
fix(compute/init): Init no longer fails if directory of same name as starter kit exists in current directory

### LINKED ISSUE #1255: init command with `-p` fails if file (or directory) with same name as the starter kit already exists in current directory
**Version**

```
Fastly CLI version v10.12.3 (f0fe32f4)
Built with go version go1.22.4 linux/amd64 (2024-07-28)
Viceroy version: viceroy 0.10.2
```

**What happened**

```
% cd ./my-app
% fastly compute init -p ./my-app
Creating a new Compute project.

Press ^C at any time to quit.

  ... (skipping lines of output) ...

Starter kit:
[1] Default starter for JavaScript
    A basic starter kit that demonstrates routing, simple synthetic responses and overriding caching rules.
    https://github.com/fastly/compute-starter-kit-javascript-default
[2] Empty starter for JavaScript
    An empty application template for the Fastly Compute environment which returns a 200 OK response.
    https://github.com/fastly/compute-starter-kit-javascript-empty
[3] Default starter for TypeScript
    A basic starter kit that demonstrates routing, simple synthetic responses and overriding caching rules.
    https://github.com/fastly/compute-starter-kit-typescript

INFO: For a complete list of Starter Kits:
    https://www.fastly.com/documentation/solutions/starters

Choose option or paste git URL: [1] 1

✗ Fetching package template

ERROR: failed to create local compute-starter-kit-javascript-default archive: open compute-starter-kit-javascript-default: is a directory.

%
```

This occurs when a directory called `compute-starter-kit-javascript-default` already exists in the current directory.

The problem seems to arise from  https://github.com/fastly/cli/blob/ad4b939e13d877410b1d8fd49f1abbd2279a8840/pkg/commands/compute/init.go#L947-L955 where the file is created in the current directory, rather than using something like `os.TempDir()`.

My current workaround is to forego the `-p` argument and switch to `my-app` directory first before running the command.

harmony7:
Resolves #1255.

This PR resolves the issue by placing the temporary file created during the download performed by `fastly compute init --from` in a temporary directory rather than the current directory.

Notes:

* This PR adds a `setupSteps` member to the test table in _init_test_, and runs those steps as part of the initialization of each test.

* This PR adds a few new test cases to _init_test_ that uses the above `setupSteps` mechanism to create a directory with the same name as the starter kit before calling `fastly compute init`.

* The existing `defer` cleanup done to remove the temporary file has been removed, because the cleanup code for the temporary directory will clean it up.

### DIFF
--- a/pkg/commands/compute/init.go
+++ b/pkg/commands/compute/init.go
@@ -944,7 +944,23 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
 		return err
 	}
 
-	filename := filepath.Base(c.CloneFrom)
+	tempdir, err := tempDir("package-init-download")
+	if err != nil {
+		err = fmt.Errorf("error creating temporary path for package template download: %w", err)
+		c.Globals.ErrLog.Add(err)
+		spinner.StopFailMessage(msg)
+		spinErr := spinner.StopFail()
+		if spinErr != nil {
+			return fmt.Errorf(text.SpinnerErrWrapper, spinErr, err)
+		}
+		return err
+	}
+	defer os.RemoveAll(tempdir)
+
+	filename := filepath.Join(
+		tempdir,
+		filepath.Base(c.CloneFrom),
+	)
 	ext := filepath.Ext(filename)
 
 	// gosec flagged this:
@@ -963,16 +979,6 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
 		}
 		return err
 	}
-	defer func() {
-		// NOTE: Later on we rename the file to include an extension and the
-		// following call to os.Remove works still because the `filename` variable
-		// that is still in scope is also updated to include the extension.
-		err := os.Remove(filename)
-		if err != nil {
-			c.Globals.ErrLog.Add(err)
-			text.Info(out, "We were unable to clean-up the local %s file (it can be safely removed)", filename)
-		}
-	}()
 
 	_, err = io.Copy(f, res.Body)
 	if err != nil {
--- a/pkg/commands/compute/init.go
+++ b/pkg/commands/compute/init.go
@@ -944,13 +944,11 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
 		return err
 	}
 
-	tempdir, err := tempDir("package-init-download")
-	if err != nil {
+	if tempdir, err := tempDir("package-init-download"); err != nil {
 		err = fmt.Errorf("error creating temporary path for package template download: %w", err)
 		c.Globals.ErrLog.Add(err)
 		spinner.StopFailMessage(msg)
-		spinErr := spinner.StopFail()
-		if spinErr != nil {
+		if spinErr := spinner.StopFail(); spinErr != nil {
 			return fmt.Errorf(text.SpinnerErrWrapper, spinErr, err)
 		}
 		return err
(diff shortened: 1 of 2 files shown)


pkg/commands/compute/init_test.go
```diff
@@ -375,6 +446,14 @@ func TestInit(t *testing.T) {
 				_ = os.Chdir(pwd)
 			}()
 
+			// Before running the test, run some steps to initialize the environment.
+			if testcase.setupSteps != nil {
+				err := testcase.setupSteps()
+				if err != nil {
+					panic(err)
```
kpfleming, line 453:
```suggestion
				if err := testcase.setupSteps(); err != nil {
					t.Fatal(err)
```

pkg/commands/compute/init_test.go
```diff
@@ -187,6 +232,32 @@ func TestInit(t *testing.T) {
...
+				},
+			},
+			wantOutput: []string{
+				"Fetching package template",
+				"Reading fastly.toml",
+				"SUCCESS: Initialized package",
+			},
+			setupSteps: func() error {
+				file, err := os.Create("main.zip")
+				if file != nil {
```
kpfleming, line 255:
```suggestion
				if file, err := os.Create("main.zip"); file != nil {
```

pkg/commands/compute/init.go
```diff
@@ -944,7 +944,23 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
 		return err
 	}
 
-	filename := filepath.Base(c.CloneFrom)
+	tempdir, err := tempDir("package-init-download")
+	if err != nil {
```
kpfleming, line 948:
```suggestion
	if tempdir, err := tempDir("package-init-download"); err != nil {
```

pkg/commands/compute/init.go
```diff
@@ -944,7 +944,23 @@ func (c *InitCommand) FetchPackageTemplate(branch, tag string, archives []file.A
...
+		err = fmt.Errorf("error creating temporary path for package template download: %w", err)
+		c.Globals.ErrLog.Add(err)
+		spinner.StopFailMessage(msg)
+		spinErr := spinner.StopFail()
+		if spinErr != nil {
```
kpfleming, line 953:
```suggestion
		if spinErr := spinner.StopFail(); spinErr != nil {
```

kpfleming (changes requested):
Some minor Go style changes, but otherwise this looks great!

harmony7:
Thanks for your review! I've applied the changes from your suggestions, please review again.  Go ahead and merge it if you approve it as I will be away for the next few days.

pkg/commands/compute/init.go
harmony7: I had to revert this one because `tempdir` is used later (outside the scope of this `if` block), so it has to stay in scope.

harmony7:
This pattern occurred many more times in this file so I applied it there too:

```diff
-spinErr := spinner.StopFail()
-if spinErr != nil {
+if spinErr := spinner.StopFail(); spinErr != nil {
```

(We may want to consider factoring this out)

pkg/commands/compute/init_test.go
harmony7: I had to revert this one because `err` was needed outside the scope of the `if` block.

harmony7:
I had to revert the inlined assign/compare pattern for a couple of the cases because the variables that were assigned needed to be available later, outside the `if` block. I don't know if there's a Go-style best practice for this kind of pattern.

kpfleming:
There is... use `var <name> <type>` before the `if` block, so that the declaration becomes independent of the block. It's not a big deal though.

kpfleming (approved):

PR #1362, opened 2024-12-17
feat(product): Add 'product' commands for each Fastly product.

kpfleming:
Adds 'enable', 'disable', and 'status' commands for each product previously supported by the 'products' command (which is now deprecated).

Also adds 'configure' commands for products which support configuration (DDoS Protection and Next-Gen WAF).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,10 +38,11 @@ require (
 )
 
 require (
+	github.com/dnaeon/go-vcr v1.2.0 // indirect
 	github.com/go-jose/go-jose/v4 v4.0.4 // indirect
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/rogpeppe/go-internal v1.11.0 // indirect
-	github.com/stretchr/testify v1.9.0 // indirect
+	github.com/stretchr/testify v1.10.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 )
 
@@ -77,3 +78,5 @@ require (
 )
 
 require 4d63.com/optional v0.2.0
+
+replace github.com/fastly/go-fastly/v9 => github.com/kpfleming/go-fastly/v9 v9.12.1-0.20241217164724-8f6ebe66851a
--- a/pkg/commands/product/bot_management/disable.go
+++ b/pkg/commands/product/bot_management/disable.go
@@ -1,0 +1,77 @@
+package bot_management
+
+import (
+	"io"
+
+	"github.com/fastly/go-fastly/v9/fastly"
+	"github.com/fastly/go-fastly/v9/fastly/products/bot_management"
+
+	"github.com/fastly/cli/pkg/api"
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/manifest"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// DisableFn is a dependency-injection point for unit tests to provide
+// a mock implementation of the API operation.
+var DisableFn = func(client api.Interface, serviceID string) error {
+	return bot_management.Disable(client.(*fastly.Client), serviceID)
+}
+
+// DisableCommand calls the Fastly API to disable the product.
+type DisableCommand struct {
+	argparser.Base
+	Manifest manifest.Data
+
+	serviceName argparser.OptionalServiceNameID
+}
+
+// NewDisableCommand returns a usable command registered under the parent.
+func NewDisableCommand(parent argparser.Registerer, g *global.Data) *DisableCommand {
+	c := DisableCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("disable", "Disable the "+bot_management.ProductName+" product")
+
+	// Optional.
+	c.RegisterFlag(argparser.StringFlagOpts{
+		Name:        argparser.FlagServiceIDName,
+		Description: argparser.FlagServiceIDDesc,
+		Dst:         &g.Manifest.Flag.ServiceID,
+		Short:       's',
+	})
+	c.RegisterFlag(argparser.StringFlagOpts{
+		Action:      c.serviceName.Set,
+		Name:        argparser.FlagServiceName,
+		Description: argparser.FlagServiceNameDesc,
+		Dst:         &c.serviceName.Value,
+	})
+	return &c
+}
+
+// Exec invokes the application logic for the command.
+func (c *DisableCommand) Exec(_ io.Reader, out io.Writer) error {
+	serviceID, source, flag, err := argparser.ServiceID(c.serviceName, *c.Globals.Manifest, c.Globals.APIClient, c.Globals.ErrLog)
+	if err != nil {
+		c.Globals.ErrLog.Add(err)
+		return err
+	}
+
+	if c.Globals.Verbose() {
+		argparser.DisplayServiceID(serviceID, flag, source, out)
+	}
+
+	err = DisableFn(c.Globals.APIClient, serviceID)
+	if err != nil {
+		c.Globals.ErrLog.Add(err)
+		return err
+	}
+
+	text.Success(out,
+		"Disabled "+bot_management.ProductName+" on service %s", serviceID)
+
+	return nil
+}
--- a/pkg/commands/product/bot_management/product_test.go
+++ b/pkg/commands/product/bot_management/product_test.go
@@ -1,0 +1,125 @@
+package bot_management_test
+
+import (
+	"testing"
+
+	"github.com/fastly/go-fastly/v9/fastly"
+	"github.com/fastly/go-fastly/v9/fastly/products/bot_management"
+
+	"github.com/fastly/cli/pkg/api"
+	root "github.com/fastly/cli/pkg/commands/product"
+	sub "github.com/fastly/cli/pkg/commands/product/bot_management"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/testutil"
+)
+
+func TestProductEnablement(t *testing.T) {
+	scenarios := []testutil.CLIScenario{
+		{
+			Name:      "validate missing Service ID: enable",
+			Args:      "enable",
+			WantError: "error reading service: no service ID found",
+		},
+		{
+			Name:      "validate missing Service ID: disable",
+			Args:      "enable",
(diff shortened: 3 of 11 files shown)


pkg/commands/product/bot_management/status.go
```diff
@@ -0,0 +1,112 @@
...
+	if s.Enabled {
+		msg = "enabled"
+	} else {
+		msg = "disabled"
+	}
+
+	text.Info(out,
+		bot_management.ProductName+" is %s on service %s", msg, serviceID)
+
+	return nil
```
cee-dub, line 111: This logic is a bit hard to follow. One idea...
```suggestion
        state := "disabled"
        _, err = GetFn(c.Globals.APIClient, serviceID)
	if err != nil {
                // The API returns a 'Bad Request' error when the product has
		// not been enabled on the service so we ignore it; any other
		// error should be reported
                var herr *fastly.HTTPError
		if !errors.As(err, &herr) || !herr.IsBadRequest() {
			c.Globals.ErrLog.Add(err)
			return err
		}
        } else {
                s.Enabled = true
                state = "enabled"
        }

        if ok, err := c.WriteJSON(out, s); ok {
		return err
	}

        text.Info(out, bot_management.ProductName+" is %s on service %s", state, serviceID)
	return nil
```

pkg/commands/product/bot_management/status.go
```diff
@@ -0,0 +1,112 @@
+package bot_management
```
cee-dub, line 1: The linter is agreeing with me that package names shouldn't have underscores.

pkg/commands/product/bot_management/product_test.go
```diff
@@ -0,0 +1,125 @@
...
+			Name:      "validate invalid json/verbose flag combo: status",
+			Args:      "status --service-id 123 --json --verbose",
+			WantError: "invalid flag combination, --verbose and --json",
+		},
+		{
+			Name: "validate success for enabling product",
+			Setup: func(t *testing.T, scenario *testutil.CLIScenario, opts *global.Data) {
+				sub.EnableFn = func(_ api.Interface, _ string) (*bot_management.EnableOutput, error) {
+					return nil, nil
+				}
```
cee-dub, line 43: Cool!

pkg/commands/product/bot_management/disable.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// NewDisableCommand returns a usable command registered under the parent.
+func NewDisableCommand(parent argparser.Registerer, g *global.Data) *DisableCommand {
+	c := DisableCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("disable", "Disable the "+bot_management.ProductName+" product")
```
cee-dub, line 37: At first, I was rather confused until I realized that `bot_management.ProductName` refers to `github.com/fastly/go-fastly/v9/fastly/products/bot_management.ProductName` and not to something in this package. Same for the rest of the `bot_management.Thing`s in `pkg/commands/product/bot_management/*.go` files.

cee-dub (commented):
Love the test mocking. Package names could be more idiomatic.

pkg/commands/product/bot_management/disable.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/manifest"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// DisableFn is a dependency-injection point for unit tests to provide
+// a mock implementation of the API operation.
+var DisableFn = func(client api.Interface, serviceID string) error {
+	return bot_management.Disable(client.(*fastly.Client), serviceID)
+}
```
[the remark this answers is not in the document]
kpfleming: Yeah, it's fairly hard to understand.

pkg/commands/product/bot_management/status.go
kpfleming: Ahh, yes, that's definitely better. I'll roll this into some other changes in the branch.

pkg/commands/product/bot_management/status.go
kpfleming: Yeah... I've been trying to stick with using our actual product IDs for ease of understanding by future maintainers, but the Go community has a different opinion (except for test packages, where underscores are allowed).

pkg/commands/product/bot_management/disable.go
kpfleming: I was confused by that when I started working in Go; if you're in package A, and you've also imported a package named A, then qualified references like `A.foo` *always* refer to the imported package and never to the 'local' package. Now that I've gotten used to it I'm not bothered by it, but this sort of thing is what prompted the desire for import aliases in the other project :-)

pkg/commands/product/bot_management/status.go
cee-dub: Sorry, I didn't use my normal editor and looks like I have a syntax error in there.

pkg/commands/product/bot_management/status.go
cee-dub: In that case it is representing a build tag. The same works for other specific words that match OS targets or architectures, like `_linux` or `_arm64`.

cee-dub: There also the only exception to two packages in one directory, where `p` and `p_test` can coexist to allow for blackbox testing of `p`.

pkg/commands/product/bot_management/disable.go
cee-dub: Might I suggest an alias is probably appropriate here? `bot` or `botapi` could work

pkg/commands/product/bot_management/disable.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/fastly/cli/pkg/api"
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/manifest"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// DisableFn is a dependency-injection point for unit tests to provide
+// a mock implementation of the API operation.
+var DisableFn = func(client api.Interface, serviceID string) error {
```
cee-dub, line 18: Probably a Go idiom or just a style preference of my own from familiarity with the standard library, but i suggest `DisableFunc` and friends.

pkg/commands/product/bot_management/disable.go
cee-dub: Try it and see what feels best, I guess!

pkg/commands/product/bot_management/disable.go
kpfleming: Resolved as part of some refactoring.

kpfleming:
@cee-dub If you're interested, this could use a second 'structural' review; much like go-fastly, this now has an `internal/productcore` package which uses generics to provide the base implementations for all of the commands in `fastly product <name> ...`. It's complete for all products except NGWAF, which requires an extra enablement parameter, and then it will get extended to support the configuration commands too.

pkg/commands/product/botmanagement/common.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/fastly/go-fastly/v9/fastly"
+	"github.com/fastly/go-fastly/v9/fastly/products/botmanagement"
+
+	"github.com/fastly/cli/internal/productcore"
+	"github.com/fastly/cli/pkg/api"
+)
+
+// EnablementHooks is a structure of dependency-injection points used
+// by unit tests to provide mock behaviors
+var EnablementHooks = productcore.EnablementHookFuncs[*botmanagement.EnableOutput]{
```
cee-dub, line 13: The `*botmanagement.EnableOutput` here confused me at first, and I was wondering why there were all these different types when everything here is nicely generic, but then I realized it is the `go-fastly` API response. Not certain this calls for an import alias, but I would find it reasonable to use one here.

internal/productcore/disable.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// Disable is a base type for all 'disable' commands.
+type Disable[O any] struct {
+	Base
+	hooks *EnablementHookFuncs[O]
```
cee-dub, line 16: I'm not sure of any reason that `hooks` needs to be of pointer type. Should be fine to use the struct's value here and in the other generic action implementations.

pkg/commands/product/botmanagement/common.go
```diff
@@ -0,0 +1,23 @@
+package botmanagement
```
cee-dub, line 1: I think this package and each of the other new enable/disable/status packages under `pkg/commands/product` should be reduced to fewer files. When they each contained full implementations, separate files made sense. Now that the generic core does most of the work, I'd prefer to see something like `commands.go` and `commands_test.go` until a product gains unique functionality.

cee-dub (commented):
Nice use of generics to deal with consuming the different API type that are essentially the same.

go.mod
```diff
@@ -78,3 +79,5 @@ require (
 )
 
 require 4d63.com/optional v0.2.0
+
+replace github.com/fastly/go-fastly/v9 => github.com/kpfleming/go-fastly/v9 v9.12.1-0.20250113130142-b6a3af55bc44
```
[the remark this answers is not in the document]
kpfleming: https://github.com/fastly/go-fastly/pull/572

That will be released before this is merged.

pkg/commands/product/botmanagement/common.go
kpfleming: Yeah... I thought about using an import alias here, as I had to do search-and-replace in 9 sets of source files :-)

internal/productcore/disable.go
kpfleming: During testing the 'global' EnablementHookFuncs structure in each package gets modified by the tests, and so if a copy was made during the instantiation of this structure (which happens during the CLI startup) the mock functions would not be called.

pkg/commands/product/botmanagement/common.go
kpfleming: Very good point... I'll do that now, and also use import aliases to remove unnecessary differences between the source files.

internal/productcore/disable.go
kpfleming: I'll add a comment to the source during the next round of edits, so the next person with this question will have an answer!

PR #1368, opened 2025-01-13
fix(build): Pin to Rust 1.83.0.

kpfleming:
Some crates used in this repository won't compile with Rust 1.84.0.

Also switch to using 'tq' binary instead of 'tomlq', as noted by the tomlq crate's warning message.


.github/workflows/pr_test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: "Checkout code"
         uses: actions/checkout@v4
       - name: "Install Rust"
-        uses: dtolnay/rust-toolchain@stable # to install tomlq via `make config`
+        uses: dtolnay/rust-toolchain@1.83.0 # to install tq via `make config`
```
Integralist, line 14: Do we not also need to update https://github.com/fastly/cli/blob/7b4df0e7dfdfb113193072553ee60d20dc4305f5/scripts/config.sh#L9

kpfleming: The crate name did not change yet, as far as I can tell.

Integralist: or does the same install create two binaries?

UPDATE: I see now it's mentioned in the CHANGELOG that it now installs a tq binary

Integralist (approved):

PR #1371, opened 2025-01-16
fix(kvstoreentry/create): rework --dir

fgsch:
Simplify code and exclude directory when inserting keys into the KV store. 

Before this change `create --dir=foo`, will create:

```
foo/abc
foo/def
```

After this change:

```
abc
def
```

### DIFF
--- a/pkg/commands/kvstoreentry/create.go
+++ b/pkg/commands/kvstoreentry/create.go
@@ -206,7 +206,6 @@ func (c *CreateCommand) ProcessDir(in io.Reader, out io.Writer) error {
 	msg := "%s %d of %d files"
 	spinner.Message(fmt.Sprintf(msg, "Processing", 0, filesTotal) + "...")
 
-	base := filepath.Base(path)
 	processed := make(chan struct{}, c.dirConcurrency)
 	sem := make(chan struct{}, c.dirConcurrency)
 	filesVerboseOutput := make(chan string, filesTotal)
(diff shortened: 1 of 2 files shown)


kpfleming:
Ouch... the existing behavior was definitely broken. Thanks for fixing it! Thankfully the other 'entry' commands (config store, secret store) don't offer the `--dir` option so they don't have this problem.

Would you be willing to tackle improving the tests for this command to ensure that the behavior is correct? A proper test would have caught this originally, it would be great to have a way to avoid regressions.

fgsch:
> Would you be willing to tackle improving the tests for this command to ensure that the behavior is correct? A proper test would have caught this originally, it would be great to have a way to avoid regressions.

Sure, I will have a look.

fgsch:
> > Would you be willing to tackle improving the tests for this command to ensure that the behavior is correct? A proper test would have caught this originally, it would be great to have a way to avoid regressions.
> 
> Sure, I will have a look.

Done. Not sure why the tests are failing. I need to have another look, but at first glance these looked unrelated.

Integralist:
> Not sure why the tests are failing.

FWIW the full test suite (running from this branch) passes for me locally.

So yeah, the fact CI is failing is weird.

fgsch:
I opened https://github.com/fastly/cli/pull/1372 to test (it changes whitespace in a single file) and I'm getting the same errors so it looks like something is up with the CI / tests.

kpfleming:
I've never seen that test failure before, and the only recent change I can think of is that `ubuntu-latest` became Ubuntu 24.04 last week. I'll open a PR to use `ubuntu-22.04` instead and see if the CI passes.

kpfleming:
Same failure with Ubuntu 22.04. This will require some investigation next week.

kpfleming:
Test suite in 'main' passes locally using Go 1.23.4 and Rust 1.83.0. Very strange.

fgsch:
I've updated this branch and the tests are passing now. PTAL.

kpfleming (approved):

PR #1377, opened 2025-01-22
Updated dependencies

SahibYar:
Updated dependencies

### Testing

I executed `make all` after updating dependencies, and output is same as `main` branch

[output.txt](https://github.com/user-attachments/files/18501173/output.txt)


kpfleming:
This looks fine to me, thanks for the contribution! Would you be willing to include a `.github/dependabot.yml` file in this PR to enable 'version' update PRs from Dependabot? That would keep us from getting out of date again, as we will get automatic PRs for new dependency versions.

SahibYar:
The Dependabot configuration is already in place, as seen here: [Dependabot.yml in the Fastly CLI repository](https://github.com/fastly/cli/blob/main/.github/dependabot.yml#L4).

kpfleming:
Indeed it is, my apologies. The problem appears to be that the configuration does not allow Dependabot to update 'indirect' dependencies, and the updates you've offered in this PR are primarily indirect dependencies. This can be cured by adding this block to the Dependabot configuration for the `gomod` package-ecosystem:
```
    allow:
      # Allow both direct and indirect updates for all packages.
      - dependency-type: "all"
```

PR #1378, opened 2025-01-23
fix(build): Correct generation of static configuration file.

kpfleming:
The behavior of 'tq' has recently changed, and just like 'jq' it wraps output in double-quotes unless it is told to generate 'raw' output.


philippschulte (approved):
Thank you!

PR #1386, opened 2025-01-30
build(pr_test.yml): tuning up the build process

anthony-gomez-fastly:
All Submissions:

### DIFF
--- a/.github/workflows/merge_to_main.yml
+++ b/.github/workflows/merge_to_main.yml
@@ -1,0 +1,35 @@
+on:
+  pull_request:
+    branches: [ "main" ]
+  merge_group:
+    types: [checks_requested]
+jobs:
+  build:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v4
+      - name: "Install Node"
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+      - name: "Install Rust"
+        uses: dtolnay/rust-toolchain@1.83.0 # to install tq via `make config`
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22.x
+      - name: "Install dependencies"
+        run: make dependencies
+        shell: bash
+      - name: "Create Build"
+        run: make build
+        shell: bash
+      - name: "Upload Build"
+        uses: actions/upload-artifact@v4
+        with:
+          name: fastly-cli-build-${{ matrix.platform }}-${{ github.sha }}
+          path: fastly
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -1,4 +1,10 @@
-on: pull_request
+on: 
+  pull_request:
+    # The specific activity types are listed here to include "labeled" and "unlabeled"
+    # (which are not included by default for the "pull_request" trigger).
+    # This is needed to allow skipping enforcement of the changelog in PRs with specific labels,
+    # as defined in the (optional) "skipLabels" property.
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
 name: Test
 # Stop any in-flight CI jobs when a new commit is pushed.
 concurrency:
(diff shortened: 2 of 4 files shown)


.github/workflows/pr_test.yml
```diff
@@ -43,6 +49,8 @@ jobs:
           echo "gobin=$(go env GOPATH)/bin" >> $GITHUB_OUTPUT # speed up dependency installs
           echo "gobuild=$(go env GOCACHE)" >> $GITHUB_OUTPUT  # speed up `go test` runs
           echo "gomod=$(go env GOMODCACHE)" >> $GITHUB_OUTPUT # speed up use of third-party modules
+      - name: "Fix cache directory perms"
```
anthony-gomez-fastly, line 52: this fixes the annoying cache errors in the build

.github/workflows/merge_to_main.yml
```diff
@@ -0,0 +1,32 @@
+on:
+  pull_request:
+    branches: [ "main" ]
+  merge_group:
+    types: [checks_requested]
+jobs:
+    build:
```
anthony-gomez-fastly, line 7: this creates an ubuntu cli build on merge to `main`, I figured it would be useful for people either wanting to test out the latest and greatest

.github/workflows/pr_test.yml
```diff
@@ -167,3 +179,29 @@ jobs:
       - name: Build docker images
         run: |
           for dockerFile in Dockerfile*; do docker build -f $dockerFile . ; done
+  build:
```
anthony-gomez-fastly, line 182: creates an ubuntu build

.github/workflows/merge_to_main.yml
philippschulte: I think this is a great idea! I guess most people who would take advantage of this are running macOS. I would recommend uploading binaries for all 3 operation systems and add the `retention-days` field to the workflow.

.github/workflows/merge_to_main.yml
```diff
@@ -0,0 +1,32 @@
...
+          uses: actions/setup-go@v5
+          with:
+            go-version: 1.22.x
+        - name: "Install dependencies"
+          run: make dependencies
+          shell: bash
+        - name: "Create Build"
+          run: make build
+          shell: bash
+        - name: "Upload Build"
```
philippschulte, line 28: I would recommend that we add a script to the repo which installs Node, Go, Rust, and the dependencies.  That script could then be invoked in all necessary workflows. With such a script we would have only one place to manage the necessary versions.

.github/workflows/merge_to_main.yml
```diff
@@ -0,0 +1,32 @@
...
+          uses: actions/upload-artifact@v4
+          with:
+            name: fastly-cli-build-ubuntu-${{ github.sha }}
+            path: fastly
```
philippschulte, line 32: Please remove the extra line.

.github/workflows/pr_test.yml
```diff
@@ -159,6 +167,10 @@ jobs:
           TEST_COMPUTE_INIT: true
           TEST_COMPUTE_BUILD: true
           TEST_COMPUTE_DEPLOY: true
+  changelog:
```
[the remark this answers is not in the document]
philippschulte: I like the idea!

.github/workflows/pr_test.yml
```diff
@@ -167,3 +179,29 @@ jobs:
...
+        run: make dependencies
+        shell: bash
+      - name: "Create Build"
+        run: make build
+        shell: bash
+      - name: "Upload Build"
+        uses: actions/upload-artifact@v4
+        with:
+          name: fastly-cli-build-ubuntu-${{ github.sha }}
+          path: fastly
```
philippschulte, line 207: Maybe it is sufficient to just have artifacts for the PR changes? If we decide to do it as proposed in this PR then then script installing the dependencies makes even more sense in my opinion.

CONTRIBUTING.md
```diff
@@ -15,5 +15,6 @@ $ git remote add upstream git@github.com:fastly/cli.git
 ```bash
 $ make all
 ```
+1. Add your changes to `CHANGELOG.MD` in [Commitizen](https://commitizen-tools.github.io/commitizen/) style message
```
philippschulte, line 18: Please change it to `CHANGELOG.md`.

philippschulte (changes requested):
Thanks for doing this. Only some nit pick requests and questions. Thanks

.github/workflows/merge_to_main.yml
anthony-gomez-fastly: I will add the mac build. I talked to CI-support and they said that they do 30 day retention by default so i was going to leave it at that unless we want to make it more restrictive?

.github/workflows/merge_to_main.yml
anthony-gomez-fastly: Probably what i'll do in addition is build an image that can then be used to both run the tests and then build, good callout

.github/workflows/pr_test.yml
anthony-gomez-fastly: what do you mean just artifacts for the changes? can you only use parts of the cli?

philippschulte: When I made the comment the pr uploaded the cli artifact twice (`.github/workflows/pr_test.yml` and in `.github/workflows/merge_to_main.yml`). In your latest changes you have removed it from `.github/workflows/pr_test.yml`. I was just wondering if it makes sense to upload the cli artifacts twice. I think it makes sense to just have it for new releases (`.github/workflows/merge_to_main.yml`) as it is right now.

philippschulte (dismissed):

.github/workflows/pr_test.yml
```diff
@@ -43,6 +49,8 @@ jobs:
           echo "gobin=$(go env GOPATH)/bin" >> $GITHUB_OUTPUT # speed up dependency installs
           echo "gobuild=$(go env GOCACHE)" >> $GITHUB_OUTPUT  # speed up `go test` runs
           echo "gomod=$(go env GOMODCACHE)" >> $GITHUB_OUTPUT # speed up use of third-party modules
+      - name: "Fix cache directory perms"
+        run: chmod -R 0755  ~/go/pkg/mod
```
kpfleming, line 53: Explicit caching may no longer be required; when I was reading the release notes for setup-go I saw [this](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) which seems to indicate that it will handle caching itself. If that's the case then removing our own cache handling will simplify the jobs.

kpfleming (dismissed):

.github/workflows/pr_test.yml
anthony-gomez-fastly: oh nice, i'll look into that!

.github/workflows/pr_test.yml
anthony-gomez-fastly: yeah i had that to make sure that the build/upload worked, removed it for the final merge

kpfleming (approved):

PR #1388, opened 2025-02-07
feat(compute): add support for acls in compute

philippschulte:
The compute ACL docs can be found [here](https://www.fastly.com/documentation/reference/api/acls/acls/#compute-acl-lookup-acls).
 
 Commands:
 
```
philipp-XPS-15-9500: fastly compute acl create

USAGE
  fastly compute acl create --name=NAME [<flags>]

Create a compute ACL

REQUIRED FLAGS
  --name=NAME  Name of the compute ACL

OPTIONAL FLAGS
  -j, --json  Render output as JSON
```

```
philipp-XPS-15-9500: fastly compute acl list-acls

USAGE
  fastly compute acl list-acls [<flags>]

List all compute ACLs

OPTIONAL FLAGS
  -j, --json  Render output as JSON
```

```
philipp-XPS-15-9500: fastly compute acl describe

USAGE
  fastly compute acl describe --acl-id=ACL-ID [<flags>]

Describe a compute ACL

REQUIRED FLAGS
  --acl-id=ACL-ID  Compute ACL ID

OPTIONAL FLAGS
  -j, --json  Render output as JSON
```

```
philipp-XPS-15-9500: fastly compute acl update

USAGE
  fastly compute acl update --acl-id=ACL-ID [<flags>]

Update a compute ACL

REQUIRED FLAGS
  --acl-id=ACL-ID  Alphanumeric string identifying a compute ACL

OPTIONAL FLAGS
  --file=FILE            Batch update json passed as file path or content, e.g.
                         $(< batch.json)
  --operation=OPERATION  Indicating that this entry is to be added to/updated in
                         the ACL
  --prefix=PREFIX        An IP prefix defined in Classless Inter-Domain Routing
                         (CIDR) format, i.e. a valid IP address (v4 or v6)
                         followed by a forward slash (/) and a prefix length
                         (0-32 or 0-128, depending on address family)
  --action=ACTION        The action taken on the IP address
```

```
philipp-XPS-15-9500: fastly compute acl lookup

USAGE
  fastly compute acl lookup --acl-id=ACL-ID --ip=IP [<flags>]

Find a matching ACL entry for an IP address

REQUIRED FLAGS
  --acl-id=ACL-ID  Compute ACL ID
  --ip=IP          Valid IPv4 or IPv6 address

OPTIONAL FLAGS
  -j, --json  Render output as JSON
  ```
  
  ```
  philipp-XPS-15-9500: fastly compute acl delete

USAGE
  fastly compute acl delete --acl-id=ACL-ID [<flags>]

Delete a compute ACL

REQUIRED FLAGS
  --acl-id=ACL-ID  Compute ACL ID

OPTIONAL FLAGS
  -j, --json  Render output as JSON
  ```
  
```
philipp-XPS-15-9500: fastly compute acl list-entries

USAGE
  fastly compute acl list-entries --acl-id=ACL-ID [<flags>]

List all entries of a compute ACL

REQUIRED FLAGS
  --acl-id=ACL-ID  Compute ACL ID

OPTIONAL FLAGS
  -c, --cursor=CURSOR  Pagination cursor (Use 'next_cursor' value from list
                       output)
  -l, --limit=50       Maximum number of items to list
  -j, --json           Render output as JSON
```

All Submissions:

### New Feature Submissions:

### CODE REVIEW

**Review by deg4uss3r — APPROVED** [review]
This looks exactly like I expected, great work!

**Review by awilliams-fastly — CHANGES_REQUESTED** [review]
I built a version of the CLI from your branch, and ran the following commands:

```
$ ./fastly compute acl lookup --acl-id=6nbj10rEjAf63NaxmBJhmW --ip 1.2.3.4

Prefix: 1.2.3.4/32
```
Here, the formatting doesn't seem right, and it's missing the `Action`. I believe the issue is that the PrefixWriter being used needs to be flushed. Looks like this may be happening in other cases of PrefixWriter as well.

```
$ ./fastly compute acl lookup --acl-id=6nbj10rEjAf63NaxmBJhmW --ip 2.2.3.4 --json

ERROR: failed to decode json response: EOF.

```
I don't think this should be output as an error. It's expected that the API return a `204 No Content` in the case where there's no matches to a lookup.

**Comment by philippschulte** [comment]
@awilliams-fastly thanks for reviewing and pointing out the issue with the `lookup` command if there are no matches. Please have a look at Ricky's comment: https://github.com/fastly/cli/pull/1388#issuecomment-2653721462.

The plan is to cut a new `go-fastly` release and update this PR with the latest version to resolve this issue.

File: pkg/text/computeacl.go [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+	}
+	tbl.Print()
+}
+
+// PrintComputeACLEntry displays a compute ACL entry.
+func PrintComputeACLEntry(out io.Writer, prefix string, entry *computeacls.ComputeACLEntry) {
+	out = textio.NewPrefixWriter(out, prefix)
+
+	fmt.Fprintf(out, "\nPrefix: %s\n", entry.Prefix)
+	fmt.Fprintf(out, "Action: %s", entry.Action)
```
kpfleming, line 42: Since this string does not contain '\n', `PrefixWriter` won't output it unless `out.Flush()` is called (see https://github.com/segmentio/textio#indenting). If this string can't include a '\n' because it needs to support continuation, then an explicit `Flush()` in this function would be a suitable fix.

pkg/commands/compute/computeacl/computeacl_test.go
```diff
@@ -0,0 +1,611 @@
...
+	var b bytes.Buffer
+	text.PrintComputeACLEntry(&b, "", entry)
+	return b.String()
+}
+
+func fmtComputeACLEntries(entries []computeacls.ComputeACLEntry) string {
+	var b bytes.Buffer
+	text.PrintComputeACLEntriesTbl(&b, entries)
+	return b.String()
+}
```
kpfleming, line 611: I'm a little concerned about these functions, as they are using the same formatters as the code-under-test uses, so that means the code-under-test is not being treated as a 'black box'.

In specific, as noted in my other comment the 'Action' is not being included in the output from `PrintComputeACLEntry` when a single entry is output (due to the lack of flushing the `PrefixWriter`), but the test is not failing because the test uses the same formatter.

In other words: the tests should probably be using their own formatters, or have hardcoded 'golden' strings to compare against.

kpfleming (changes requested):

kpfleming:
> Here, the formatting doesn't seem right, and it's missing the `Action`. I believe the issue is that the PrefixWriter being used needs to be flushed. Looks like this may be happening in other cases of PrefixWriter as well.

Confirmed, and the unit tests did not catch this because they use the same formatting functions as the code-under-test uses. I've left a comment asking for that to be changed, as the tests should not make use of internal functionality of the code being tested.

awilliams-fastly:
> @awilliams-fastly thanks for reviewing and pointing out the issue with the lookup command if there are no matches. Please have a look at Ricky's comment: https://github.com/fastly/cli/pull/1388#issuecomment-2653721462.

Thanks for catching me up on those discussions, and sorry that I didn't see them earlier.

I built a version of the CLI using this branch **+ the updated `go-fastly` package**, and ran a similar test as before:

```shell
$ ./fastly compute acl lookup --acl-id=6nbj10rEjAf63NaxmBJhmW --ip 2.2.3.4 --json

ERROR: 204 - No Content.

```

I believe the `fastly compute acl lookup` command would be more user-friendly and cause less confusion if it didn't show an error whenever an IP isn't found, since it's not really an error, rather a lack of results.

I think there's a way to address this by having `computeacls.Lookup` ([here](https://github.com/fastly/cli/pull/1388/files#diff-95a1655c1bf603aa725c4c21f87fe7ed8e911f0541e6aea48fc57f29414bf43aR57)) somehow distinguish between an error, such as a network timeout, and a `204 No Content` response. The latter being the case in the output above.

I've described one way to do this in a [comment on the `go-fastly` PR](https://github.com/fastly/go-fastly/pull/595/files#r1953017040). Am happy to go over other ideas with you. Again, sorry for not seeing this earlier.

pkg/text/computeacl.go
awilliams-fastly: Agree, and please note that the same applies to all other uses of `PrefixWriter`.

pkg/commands/compute/computeacl/computeacl_test.go
kpfleming: An alternative solution is to leave these functions as they are, but add unit tests for the `PrintComputeACL...` functions themselves. Those units tests would have hardcoded input and output for comparison, and they would ensure that all of the provided input appears in the output and in the correct format.

pkg/text/computeacl.go
philippschulte: I just missed to add the `\n`; addressed in c766bc2af3bc2b7a61f556ce2fac779309c5a6dd.

pkg/commands/compute/computeacl/computeacl_test.go
philippschulte: Addressed in e7fe55a0ba6586ae4ddae940b05ed79d65d3ccc4.

philippschulte:
@awilliams-fastly thanks for the review!

> I believe the `fastly compute acl lookup` command would be more user-friendly and cause less confusion if it didn't show an error whenever an IP isn't found, since it's not really an error, rather a lack of results.

Addressed in 4ab0ab1e2a36d16b5cfd9c4b3efc444747ac4bf1.

kpfleming (changes requested):
Just a few minor changes, otherwise this looks really good!

pkg/commands/compute/computeacl/lookup.go
```diff
@@ -0,0 +1,78 @@
...
+		ComputeACLIP: &c.ip,
+	})
+	if err != nil {
+		c.Globals.ErrLog.Add(err)
+		return err
+	}
+
+	// Status 204 - No Content
+	if entry == nil {
+		text.Warning(out, "Compute ACL (%s) has no entry with IP (%s)", c.id, c.ip)
```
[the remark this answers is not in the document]
philippschulte: Thanks for pointing this out! Instead of returning an empty object I am using Go's default behavior for `nil` which translates in `json` to `null` which of course is valid `json`. I hope that works for you.

Addressed in d4b7803c58f5c06f68e10035f640bf4fe748bba8.

pkg/commands/compute/computeacl/update.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+	}
+
+	c.CmdClause = parent.Command("update", "Update a compute ACL")
+
+	// Required.
+	c.CmdClause.Flag("acl-id", "Alphanumeric string identifying a compute ACL").Required().StringVar(&c.computeACLID)
+
+	// Optional.
+	c.CmdClause.Flag("file", "Batch update json passed as file path or content, e.g. $(< batch.json)").Action(c.file.Set).StringVar(&c.file.Value)
```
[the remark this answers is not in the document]
philippschulte: Addressed in a4f8c8f3296bf9cbdd4ca79c2271f78d80ff5755.

pkg/commands/compute/computeacl/computeacl_test.go
```diff
@@ -0,0 +1,618 @@
...
+    							"status": 400
+							}
+						`))),
+					},
+				},
+			},
+			WantError: "400 - Bad Request",
+		},
+		{
+			Name: "validate API status 204 - no content",
```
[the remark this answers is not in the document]
philippschulte: Addressed in d4b7803c58f5c06f68e10035f640bf4fe748bba8.

kpfleming (approved):

awilliams-fastly (approved):
Thanks for making those changes.

I re-ran my tests from previously:
```
./fastly compute acl lookup --acl-id=6nbj10rEjAf63NaxmBJhmW --ip 1.2.3.4

Prefix: 1.2.3.4/32
Action: ALLOW

./fastly compute acl lookup --acl-id=6nbj10rEjAf63NaxmBJhmW --ip 2.2.3.4
INFO: Compute ACL (6nbj10rEjAf63NaxmBJhmW) has no entry with IP (2.2.3.4)
```

I'm not sure if there's meant to be an extra newline before the Prefix line? Otherwise, looks good to me.

deg4uss3r (approved):

philippschulte:
> I'm not sure if there's meant to be an extra newline before the Prefix line? Otherwise, looks good to me.

I took it from here: https://github.com/fastly/cli/blob/40aee817d419365fbe0354ab80ffc38609942e4b/pkg/text/kvstore.go#L19 but it looks like that this is the only feature we are adding an extra newline at the top.
Therefore, I removed it in 74ac0d510cae641b08110161a54156bcbafb222c.

philippschulte:
> > I'm not sure if there's meant to be an extra newline before the Prefix line? Otherwise, looks good to me.
> 
> I took it from here:
> 
> https://github.com/fastly/cli/blob/40aee817d419365fbe0354ab80ffc38609942e4b/pkg/text/kvstore.go#L19
> but it looks like that this is the only feature we are adding an extra newline at the top.
> Therefore, I removed it in [74ac0d5](https://github.com/fastly/cli/commit/74ac0d510cae641b08110161a54156bcbafb222c).

awilliams-fastly (approved):

PR #1428, opened 2025-03-05
feat(accesskeys): add support for access keys

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Users will be able to manage their access keys through the cli

### DIFF
--- a/pkg/commands/objectstorage/accesskeys/accesskeys_test.go
+++ b/pkg/commands/objectstorage/accesskeys/accesskeys_test.go
@@ -20,6 +20,8 @@ const (
 	akSecret      = "accessKeySecret"
 	akDescription = "accessKeyDescription"
 	akPermission  = "read-only-objects"
+	akBucket1     = "bucket1"
+	akBucket2     = "bucket2"
 )
 
 var ak = accesskeys.AccessKey{
@@ -27,6 +29,7 @@ var ak = accesskeys.AccessKey{
 	SecretKey:   akSecret,
 	Description: akDescription,
 	Permission:  akPermission,
+	Buckets:     []string{akBucket1, akBucket2},
 	CreatedAt:   testutil.Date,
 }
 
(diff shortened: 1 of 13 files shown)


pkg/commands/objectstorage/accesskeys/accesskeys_test.go
```diff
@@ -0,0 +1,310 @@
...
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusOK,
+						Status:     http.StatusText(http.StatusOK),
+						Body:       io.NopCloser(bytes.NewReader(testutil.GenJSON(ak))),
+					},
+				},
+			},
+			WantOutput: fstfmt.EncodeJSON(ak),
+		},
```
philippschulte, line 86: We should have a test which specifies the optional `buckets` flag. Also to see if the data type `string` for `buckets` is the best option `kingpin` has to offer. Maybe there is something inside `kingpin` better suited for list types?

pkg/commands/objectstorage/accesskeys/get.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/fastly/cli/pkg/text"
+)
+
+// GetCommand calls the Fastly API to get an access key.
+type GetCommand struct {
+	argparser.Base
+	argparser.JSONOutput
+
+	// Required.
+	id string
```
philippschulte, line 22: I would recommend to replace `id` with `key` or `access-key` everywhere in the implementation since `id` is not used anywhere in our public docs as well.

philippschulte (dismissed):
This looks great! Just some nit picks and linting checks you need to address.

pkg/commands/objectstorage/accesskeys/accesskeys_test.go
anthony-gomez-fastly: added a test, and followed the convention for multiple strings by allowing the flag to be specified multiple times

kpfleming (dismissed):
This does indeed look great. There are two linter checks showing up here that probably don't show up locally, but they are easy fixes.

kpfleming (approved):

PR #1446, opened 2025-03-31
Support new file/format model shape for local_server.kv_store and secret_store (as well as metadata in kv_store)

### LINKED ISSUE #1319: CLI doesn't handle new KV store (and upcoming secret store) local server config
**Version**

Fastly CLI version v10.14.1 (b3926881)
Built with go version go1.22.7 linux/amd64 (2024-09-17)
Viceroy version: viceroy 0.12.1

**What happened**

Viceroy recently added support for loading KV store data from a file in https://github.com/fastly/Viceroy/pull/365, and equivalent secret store support in https://github.com/fastly/Viceroy/pull/428.

Previously in TOML you could specify entries like:

```toml
[local_server]
kv_stores.store_name = [{ key = "a_key", data = "hello world" }]
```

ie, an array.  But now you can also define it as an object like so:

```toml
[local_server]
kv_stores.store_name = {file = "entries.json", format = "json"}
```

ie, an object.  Secret store's syntax will be the same.

The CLI doesn't like this and throws an error:

```
✗ Verifying fastly.toml

ERROR: (0, 0): Can't convert file = "entries.json"
format = "json"
(*toml.Tree) to a tree.
```

This is probably because the struct definition for `local_server` tries to deserialize into a `map[string][]LocalKVStore` here: https://github.com/fastly/cli/blob/b3926881fe8d74598c72978a65ffa5d6c90a82b4/pkg/manifest/local_server.go.  This probably needs to be changed to accept either a struct or a slice.

The workaround is to move these KV/secret store definitions to a separate toml file and run viceroy with them directly.

harmony7:
Viceroy's KV Store can now be configured using external JSON files specified using fastly.toml (https://github.com/fastly/Viceroy/pull/365 and https://github.com/fastly/Viceroy/pull/428).

However, Fastly CLI's fastly.toml validation would choke and it was not possible to use the new syntax.

```
✗ Verifying fastly.toml

ERROR: (0, 0): Can't convert file = "./foo.json"
format = "json"
(*toml.Tree) to a tree.
```

This PR updates the TOML marshaling/unmarshaling mechanism to enable the new syntax in fastly.toml.

Fixes #1319

### Changes to Core Features:

This should be backwards compatible.

### DIFF
--- a/pkg/manifest/file.go
+++ b/pkg/manifest/file.go
@@ -81,6 +81,9 @@ func (f *File) MarshalTOML() ([]byte, error) {
 					if e.Data != "" {
 						obj["data"] = e.Data
 					}
+					if e.Metadata != "" {
+						obj["metadata"] = e.Metadata
+					}
 					items = append(items, obj)
 				}
 				kvStores[key] = items
--- a/pkg/manifest/file.go
+++ b/pkg/manifest/file.go
@@ -2,6 +2,7 @@ package manifest
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -48,6 +49,107 @@ type File struct {
 	exists    bool
 	output    io.Writer
 	readError error
+}
+
+// MarshalTOML performs custom marshalling to TOML for objects of File type.
+func (f *File) MarshalTOML() ([]byte, error) {
+	localServer := make(map[string]interface{})
+
+	if f.LocalServer.Backends != nil {
+		localServer["backends"] = f.LocalServer.Backends
+	}
+
+	if f.LocalServer.ConfigStores != nil {
+		localServer["config_stores"] = f.LocalServer.ConfigStores
+	}
+
+	if f.LocalServer.KVStores != nil {
+		kvStores := make(map[string]interface{})
+		for key, entry := range f.LocalServer.KVStores {
+			if entry.External != nil {
+				kvStores[key] = map[string]interface{}{
+					"file":   entry.External.File,
+					"format": entry.External.Format,
+				}
+			} else {
+				items := make([]map[string]interface{}, 0, len(entry.Array))
+				for _, e := range entry.Array {
+					obj := map[string]interface{}{"key": e.Key}
+					if e.File != "" {
+						obj["file"] = e.File
+					}
+					if e.Data != "" {
+						obj["data"] = e.Data
+					}
+					items = append(items, obj)
+				}
+				kvStores[key] = items
+			}
+		}
+		localServer["kv_stores"] = kvStores
+	}
+
+	if f.LocalServer.SecretStores != nil {
+		secretStores := make(map[string]interface{})
+		for key, entry := range f.LocalServer.SecretStores {
+			if entry.External != nil {
+				secretStores[key] = map[string]interface{}{
+					"file":   entry.External.File,
+					"format": entry.External.Format,
+				}
+			} else {
+				items := make([]map[string]interface{}, 0, len(entry.Array))
+				for _, e := range entry.Array {
+					obj := map[string]interface{}{"key": e.Key}
+					if e.File != "" {
+						obj["file"] = e.File
+					}
+					if e.Data != "" {
+						obj["data"] = e.Data
+					}
+					items = append(items, obj)
+				}
+				secretStores[key] = items
+			}
+		}
+		localServer["secret_stores"] = secretStores
+	}
+
+	if f.LocalServer.ViceroyVersion != "" {
+		localServer["viceroy_version"] = f.LocalServer.ViceroyVersion
+	}
+
+	type outputFile struct {
+		Authors         []string    `toml:"authors"`
+		ClonedFrom      string      `toml:"cloned_from,omitempty"`
+		Description     string      `toml:"description"`
+		Language        string      `toml:"language"`
+		Profile         string      `toml:"profile,omitempty"`
+		LocalServer     interface{} `toml:"local_server"` // override this field
+		ManifestVersion Version     `toml:"manifest_version"`
+		Name            string      `toml:"name"`
+		Scripts         Scripts     `toml:"scripts,omitempty"`
+		ServiceID       string      `toml:"service_id"`
+		Setup           Setup       `toml:"setup,omitempty"`
+	}
+
+	out := outputFile{
+		Authors:         f.Authors,
+		ClonedFrom:      f.ClonedFrom,
+		Description:     f.Description,
+		Language:        f.Language,
+		Profile:         f.Profile,
+		LocalServer:     localServer,
+		ManifestVersion: f.ManifestVersion,
+		Name:            f.Name,
+		Scripts:         f.Scripts,
+		ServiceID:       f.ServiceID,
+		Setup:           f.Setup,
+	}
+
+	var buf bytes.Buffer
+	err := toml.NewEncoder(&buf).Encode(out)
+	return buf.Bytes(), err
 }
 
 // Exists yields whether the manifest exists.
--- a/pkg/manifest/file.go
+++ b/pkg/manifest/file.go
@@ -53,7 +53,7 @@ type File struct {
 
 // MarshalTOML performs custom marshalling to TOML for objects of File type.
 func (f *File) MarshalTOML() ([]byte, error) {
-	localServer := make(map[string]interface{})
+	localServer := make(map[string]any)
 
 	if f.LocalServer.Backends != nil {
 		localServer["backends"] = f.LocalServer.Backends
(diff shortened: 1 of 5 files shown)


harmony7:
Working with Marshal and UnmarshalJSON have been pretty crazy, there may be cleaner ways to do this. I'm open to any suggestions.

kpfleming:
No need to request a review from @Integralist for this, the DevTools team can handle it now.

philippschulte (changes requested):
I have found only some nit picks otherwise this looks good to me!

Withe the following config:

```
philipp@philipp-XPS-15-9500:~/src/fastly/cli/ps$ cat fastly.toml 
# This file describes a Fastly Compute package. To learn more visit:
# https://www.fastly.com/documentation/reference/compute/fastly-toml

authors = ["ps@gmail.com"]
cloned_from = "https://github.com/fastly/compute-starter-kit-rust-default"
description = "ps"
language = "rust"
manifest_version = 3
name = "ps"
service_id = ""

[local_server]
kv_stores.store_name = {file = "entries.json", format = "json"}

[scripts]
  build = "cargo build --profile release"
```

everything builds:

```
philipp@philipp-XPS-15-9500:~/src/fastly/cli/ps$ ./fastly compute build
✓ Verifying fastly.toml
✓ Identifying package name
✓ Identifying toolchain
✓ Running [scripts.build]
✓ Creating package archive

SUCCESS: Built package (pkg/ps.tar.gz)
```

pkg/manifest/testdata/fastly-viceroy-update.toml
```diff
@@ -40,6 +40,7 @@ store_one = [
   { key = "first", data = "This is some data" },
   { key = "second", file = "strings.json" },
 ]
+store_three = { file = "path/to/kv.json", format = "json" }
```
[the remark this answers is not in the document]
harmony7: I added this to make sure we had a test case for this

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,246 @@
...
+			decoder := toml.NewDecoder(strings.NewReader(tt.inputTOML))
+			err := decoder.Decode(&m)
+
+			buf := new(bytes.Buffer)
+			encoder := toml.NewEncoder(buf)
+
+			encodeErr := encoder.Encode(m)
+			if encodeErr != nil {
+				log.Fatalf("TOML encode failed: %v", encodeErr)
+			}
```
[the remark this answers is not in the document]
harmony7: good catch, this was left over debug code. I'll remove

harmony7:
Thanks for your review.
I have applied your suggestions, and seen that the tests pass again.

Please review once more =)

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+import (
+	"bytes"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/pelletier/go-toml"
+)
+
+type KVWrapper struct {
```
philippschulte, line 12: We don't need this.

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+	KVStores LocalKVStoreMap `toml:"kv_stores"`
+}
+
+func (w KVWrapper) MarshalTOML() ([]byte, error) {
```
philippschulte, line 16: Same here.

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	obj["kv_stores"] = kv
+
+	buf := new(bytes.Buffer)
+	err := toml.NewEncoder(buf).Encode(obj)
+	return buf.Bytes(), err
+}
+
+type SecretStoreWrapper struct {
```
philippschulte, line 38: Same here.

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+	SecretStores LocalSecretStoreMap `toml:"secret_stores"`
+}
+
+func (w SecretStoreWrapper) MarshalTOML() ([]byte, error) {
```
philippschulte, line 42: Same here.

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+[kv_stores]
+my-kv = "not-a-valid-entry"
+`,
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var m KVWrapper
```
philippschulte, line 114: This can be replaced with `LocalServer`.

pkg/manifest/local_server_test.go
```diff
@@ -0,0 +1,214 @@
...
+[secret_stores]
+my-secret-store = "not-a-valid-entry"
+`,
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var m SecretStoreWrapper
```
philippschulte, line 190: This can be replaced with `LocalServer`.

philippschulte (changes requested):
Thanks for the changes. I believe we don't need those wrapper types in the test suite. Please have a look and let me know. Thanks.

pkg/manifest/file.go
```diff
@@ -50,6 +51,105 @@ type File struct {
...
+					}
+					if e.Data != "" {
+						obj["data"] = e.Data
+					}
+					items = append(items, obj)
+				}
+				secretStores[key] = items
+			}
+		}
+		localServer["secret_stores"] = secretStores
```
kpfleming, line 115: This is minor and non-blocking, but this loop block is nearly identical to the one above and could be refactored into a function used by both. Since we've had bugs in the past where behavior was changed for one type of store and not the other, reducing duplication will be beneficial.

pkg/manifest/local_server.go
```diff
@@ -24,16 +31,171 @@ type LocalConfigStore struct {
 	Contents map[string]string `toml:"contents,omitempty"`
 }
 
-// LocalKVStore represents an kv_store to be mocked by the local testing server.
-type LocalKVStore struct {
+// KVStoreArrayEntry represents an array-based key/value store entries.
+// It expects a key plus either a data or file field.
+type KVStoreArrayEntry struct {
```
kpfleming, line 36: Does this need to be extended to support 'Metadata', as was done in https://github.com/fastly/Viceroy/pull/459?

kpfleming (approved):
A couple of questions, only one of which is potentially blocking.

pkg/manifest/local_server_test.go
harmony7: It turns out MarshalTOML wasn't part of some leftover code I had during development, so I removed it.

pkg/manifest/local_server_test.go
harmony7: I've turned it into an anonymous struct in the test.

pkg/manifest/local_server_test.go
harmony7: same

pkg/manifest/local_server_test.go
harmony7: same

pkg/manifest/local_server_test.go
harmony7: same

pkg/manifest/local_server_test.go
harmony7: I used an anonymous struct that looks like this:
```go
			var m struct {
				KVStores LocalKVStoreMap `toml:"kv_stores"`
			}
```

Since this test is supposed to test LocalKVStoreMap UnmarshalTOML only, this struct is beneficial for having just that layer.

pkg/manifest/local_server_test.go
harmony7: same

pkg/manifest/local_server.go
harmony7: Yes, I've pushed a commit to support it, see my newest push.

pkg/manifest/file.go
harmony7: KV Store and Secret Store do not have the same structure anymore so I'm going to leave this alone for now, please look at my newest commit.

harmony7:
Thanks for your comments, @philippschulte 
I've pushed a commit to address the test, please take a look.

harmony7:
Thanks for your comments, @kpfleming 
I tried adding `metadata` to `fastly-viceroy-update.toml` and saw that the test broke.
So I've pushed a commit that adds support for Metadata and fix the test case.

philippschulte (approved):
LGTM. Thanks for addressing all of my comments.

kpfleming (approved):
Looks great!

Once this is merged we'll need to update the [fastly.toml docs](https://www.fastly.com/documentation/reference/compute/fastly-toml/) too.

joeshaw:
Thank you for tackling this! 👍

harmony7:
Thanks everyone!

> Once this is merged we'll need to update the [fastly.toml docs](https://www.fastly.com/documentation/reference/compute/fastly-toml/) too.

Agreed, I'll work on that too.

PR #1479, opened 2025-05-12
ci(go.mod): replace go tools linter and formatter with golangci and move tools into their own mod file

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:


.github/workflows/pr_test.yml
```diff
@@ -45,28 +45,10 @@ jobs:
...
-      - name: "Static analysis check"
-        run: make staticcheck
-        shell: bash
-      - name: "Security audit"
-        run: make gosec
-        shell: bash
-      - name: "Run go imports"
-        run: make imports
-        shell: bash
+      - name: golangci-lint
```
anthony-gomez-fastly, line 48: replace individual linters with golangci action

Makefile
```diff
@@ -42,7 +42,7 @@ endif
 # EXAMPLE:
 # make release GORELEASER_ARGS="--clean --skip=post-hooks --skip=validate"
 release: dependencies $(GO_FILES) ## Build executables using goreleaser
-	goreleaser build ${GORELEASER_ARGS}
+	$(GO_BIN) tool -modfile=tools.mod goreleaser build ${GORELEASER_ARGS}
```
anthony-gomez-fastly, line 45: use the new `tools.mod`

kpfleming: As a followup, I'd like to switch to using binary builds of `goreleaser` too, but I'm happy to work on that so you can move on to more interesting work!

pkg/commands/acl/describe.go
```diff
@@ -59,7 +59,7 @@ type DescribeCommand struct {
 
 // Exec invokes the application logic for the command.
 func (c *DescribeCommand) Exec(_ io.Reader, out io.Writer) error {
-	if c.Globals.Verbose() && c.JSONOutput.Enabled {
+	if c.Globals.Verbose() && c.Enabled {
```
kpfleming, line 62: I understand why the linter is suggesting this change, but I dislike it, as it loses useful information for the reader of the code. In some cases the containing structure may have its own `Enabled` field which shadows the one from the embedded `JSONOutput` structure, in which case we'd have to use the more explicit syntax anyway.

If this lint can be disabled so that we can retain the explicit syntax without generating a warning, I'd prefer that.

kpfleming (changes requested):
One change request, the rest looks great!

pkg/commands/acl/describe.go
kpfleming: An alternative, if you prefer: the field name in the `JSONOutput` struct could be changed to `JSONOutputEnabled`. That would be repetitive within the structure, but since most usage of it (possibly all) is as an embedded structure inside another structure, that would be fine and the resulting code in all the commands would be `c.JSONOutputEnabled` instead of `c.Enabled`.

.github/workflows/pr_test.yml
```diff
@@ -45,28 +46,16 @@ jobs:
...
-        run: make imports
-        shell: bash
+      - name: "Config Artifact"
+        uses: actions/download-artifact@v4
+        with:
+          name: config-artifact-${{ github.sha }}
+      - name: "Move Config"
+        run: mv config.toml pkg/config/config.toml
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v8
```
Integralist, line 56: Might be worth adding https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#only-new-issues

pkg/commands/acl/describe.go
Integralist: I'd prefer not renaming the field and ignoring this particular lint issue.

I'm not sure which linter is reporting an error, but if you find the linter github repo then there's a chance it might support either "skip" or "ignore" functionality (I usually grep the linter repos for those terms).

anthony-gomez-fastly: fixed in `f75db468` by ignoring the `QF1008` rule in staticcheck

.github/workflows/pr_test.yml
anthony-gomez-fastly: good shout, adding

kpfleming (approved):

Makefile
anthony-gomez-fastly: so interestingly we actually use the [goreleaser github action](https://github.com/fastly/cli/blob/main/.github/workflows/tag_to_draft_release.yml#L35) on CI so we would only need to use the binaries locally? also I can't find any calls to the `go-callvis` tool outside of the makefile so i'm not sure when to use that

Integralist (approved):

Makefile
kpfleming: Yep, only for pre-release testing and for testing changes to the goreleaser configuration.

PR #1485, opened 2025-05-29
Increase max config store value length

awilliams-fastly:
The maximum allowed length for a config store entry's value is now configurable. The default is still 8k characters, but may be larger. This increases the limit in the CLI to 64k. If a user attempts to create an entry larger than their allowed limit, the API will return an error. 64k was chosen as a limit to prevent reading from a never-ending input stream.

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Users with a limit larger than 8k will be able to create config store entries larger than 8k.

### DIFF
--- a/pkg/commands/configstoreentry/errors.go
+++ b/pkg/commands/configstoreentry/errors.go
@@ -9,9 +9,9 @@ import (
 
 const (
 	maxKeyLen = 256
-	// maxValueLen is the maximum length of config store entry's value. It's set to 64k,
-	// even though users may have a smaller limit. The API will reject requests if the
-	// value is larger than the user's limit.
+	// maxValueLen is the maximum length of Config Store entry's value. It's set to 64k,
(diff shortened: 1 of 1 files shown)


pkg/commands/configstoreentry/errors.go
```diff
@@ -8,8 +8,11 @@ import (
 )
 
 const (
-	maxKeyLen   = 256
-	maxValueLen = 8000
+	maxKeyLen = 256
+	// maxValueLen is the maximum length of config store entry's value. It's set to 64k,
+	// even though users may have a smaller limit. The API will reject requests if the
```
kpfleming, line 13: By 'user' here do you mean 'customer' (as in `CID`)? If so, the comment should probably be specific about that, so there's no confusion.

kpfleming (approved):

PR #1509, opened 2025-07-19
feat: Add experimental "enable Pushpin" mode

harmony7:
This PR adds code to the CLI to enable `fastly compute serve` with "enable Pushpin", a new mode that locates and runs Pushpin before starting Viceroy to run guest code.

This is a companion PR to https://github.com/fastly/Viceroy/pull/497, and will only function after that feature has landed.

## How to enable it:

It can be invoked in one of two ways:

1. set a new flag `--experimental-enable-pushpin`
2. or, set a new flag in `fastly.toml`:

```toml
[local_server.pushpin]
enable = true
```

## What it does:

* `fastly compute serve` starts Pushpin before starting Viceroy, and detects its startup. It also sets up to shut it down when the CLI process ends.
    * `pushpin` must be started at this time, because routes are set based on the backends defined in the `fastly.toml` manifest file.
    * "kill process" is done differently between Windows and Unix, so build flags are used to include the right implementation of `killProcess`.

* Pushpin routes are set up based on the `fastly.toml` manifest file.
    * A route is set up per backend.
    * The name of each backend is set as an `id=` condition. For example, a backend named `origin` will set `id=origin`. Viceroy handles Fanout handoff by forwarding a request for this backend to Pushpin and set the header `Pushpin-Route: origin`, allowing Pushpin to match the route.
    * If the backend definition's URL has a path segment, then it is set as a `replace_beg` configuration on the route. For example, if a backend is configured for `http://localhost:3000/realtime`, then this is set as `replace_beg=realtime`. Pushpin will forward a request for `/api/user` to this backend as `http://localhost:3000/realtime/api/user`.
    * `over_http` is set for every route target, as to always enable WebSocket-over-HTTP.
    * If the backend URL is HTTPS, then `ssl` is added.
    * If the backend definition sets `override_host`, then it will be set as the `host` value of the route target.

* When starting Viceroy, `--local-pushpin-proxy-port=<port>` is appended as a command-line argument, which starts Viceroy in "enable Pushpin" mode.

* Pushpin proxy runs by default on port `7677`, but this can be overridden:
   * using `proxy_port=<port>` under `local_server.pushpin` section of the `fastly.toml` file, or
   * using `--pushpin-proxy-port=<port>`

* Pushpin's publishing endpoint runs by default on port `5561`, but this can be overridden:
   * using `publish_port=<port>` under `local_server.pushpin` section of the `fastly.toml` file, or
   * using `--pushpin-publish-port=<port>`

* The `pushpin` binary is searched on the system path, but can be specified:
   * using `pushpin_path=/path/to/pushpin` under the `local_server.pushpin` section of the `fastly.toml` file, or
   * using `--pushpin-path=/path/to/pushpin`.

* The CLI writes the config and routes to temporary files in the OS temporary directory.  This is cleaned up as part of shutdown.

* Pushpin runner places some runtime files into a temporary directory specified in pushpin.conf.  We set this to an OS temporary directory, so that multiple instances can run if necessary.  This is cleaned up as part of shutdown.

* Pushpin runner places logs into a directory specified in pushpin.conf.  This is placed in the ./pushpin-logs directory under the current project

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

The new feature is opt-in using the `--experimental-enable-pushpin` flag, and does nothing if the new flag is not set.


harmony7:
Reverting to draft, as I am going to add another feature.

harmony7:
I've added the following updates to the way this works:

We write a routes file to disk to the OS temporary directory, and reference it from a fresh copy of the pushpin.conf file (which I've copied out from the Pushpin repo and made some modifications to), also written to the temporary directory.
This conf file is configured with the publish and proxy ports, as well as a temporary directory for the runtime files, and settings logs output to pushpin-logs under the compute directory.

pkg/commands/compute/pushpin.conf.template
```diff
@@ -0,0 +1,176 @@
...
+# maximum number of client connections
+client_maxconn=50000
+
+# whether connections can use compression
+allow_compression=false
+
+# paths
+mongrel2_bin=mongrel2
+m2sh_bin=m2sh
+zurl_bin=zurl
```
kpfleming: Are these additional binaries that Pushpin might try to execute? If so that could end up being a problem as the user might have specified an explicit path to `pushpin` (meaning it is not locatable using `PATH`) and these binaries won't be locatable.

pkg/commands/compute/serve.go
```diff
@@ -92,6 +106,10 @@ func NewServeCommand(parent argparser.Registerer, g *global.Data, build *BuildCo
 	c.CmdClause.Flag("metadata-filter-envvars", "Redact specified environment variables from [scripts.env_vars] using comma-separated list").Action(c.metadataFilterEnvVars.Set).StringVar(&c.metadataFilterEnvVars.Value)
 	c.CmdClause.Flag("metadata-show", "Inspect the Wasm binary metadata").Action(c.metadataShow.Set).BoolVar(&c.metadataShow.Value)
 	c.CmdClause.Flag("package-name", "Package name").Action(c.packageName.Set).StringVar(&c.packageName.Value)
+	c.CmdClause.Flag("experimental-enable-pushpin", "Enable experimental Pushpin support for local testing of Fanout and WebSockets").BoolVar(&c.enablePushpin)
+	c.CmdClause.Flag("pushpin-path", "The path to a user installed version of the Pushpin runner binary").StringVar(&c.pushpinRunnerBinPath)
+	c.CmdClause.Flag("pushpin-proxy-port", "The port to run the Pushpin runner on. Defaults to 7677.").StringVar(&c.pushpinProxyPort)
+	c.CmdClause.Flag("pushpin-publish-port", "The port to run the Pushpin publish handler on. Defaults to 5611.").StringVar(&c.pushpinPublishPort)
```
kpfleming: Are these defaults in Pushpin itself? Normally if there is a default we'd specify it as part of the `Flag` definition, which allows Kingpin to display the default (and then it doesn't need to be included in the help text). If a user has built their own copy of Pushpin and changed the defaults, this help text will be incorrect.

I think it would be better to explicitly include the defaults here even if they are expected to match the ones in the Pushpin binary, especially now that the CLI writes its own `pushpin.conf` and the default port numbers are under control of the CLI.

pkg/commands/compute/serve.go
```diff
@@ -197,6 +215,217 @@ func (c *ServeCommand) Exec(in io.Reader, out io.Writer) (err error) {
 		return err
 	}
 
+	enablePushpin := c.enablePushpin || c.Globals.Manifest.File.LocalServer.Pushpin.EnablePushpin
+	var pushpinProxyPort uint16
+	if enablePushpin {
```
kpfleming, line 220: If there's a way to move this entire block into its own function that would be preferable.

pkg/commands/compute/pushpin.conf.template
harmony7: These binaries are not run by default (they are only run under certain configurations of Pushpin, which won't be used in this integration). They are safe to leave as is but it's also safe to remove them so I'm going to do that.

pkg/commands/compute/serve.go
harmony7: These need to be able to be left blank, and in that case they "fall back" to what's specified in fastly.toml.
If I put a default value using `.Default()` then I won't be able to differentiate the case when the flag is not provided.

So if the value read from this is "", then it tries to get the value from fastly.toml, and if not specified, then it specifically assigns these default values in code.

I've added clarification to the help text however.

See `GetPushpinProxyPort()` and `GetPushpinPublishPort()`.

harmony7:
@kpfleming thanks for the review! I've made fixes, please re-review =)

pkg/commands/compute/serve.go
harmony7: have done so, please take a look

PR #1511, opened 2025-08-07
docs: improve object-storage access-keys create flag docs

joeshaw:
Specify the valid permissions (read-write-admin, read-only-admin,
read-write-objects, read-only-objects) and note that if you don't
provide the `--bucket` flag, all buckets are accessible by the key.

 All Submissions:

### New Feature Submissions:

### User Impact

Better help output.


joeshaw:
I wasn't sure: do I create a changelog entry now that I know what the PR number is?

kpfleming:
No changelog entry is needed for this type of change. I've added the label to skip the check and will re-run CI once it has completed the first run.

PR #1514, opened 2025-08-08
Upgrade to go-fastly version 11.

kpfleming:
go-fastly version 11 requires a context.Context object to be passed as the first argument in all of the API functions it exports. This commit modifies all of the code in the CLI to pass context.TODO() in that case, since the CLI does not use context objects for any other purpose.

The CLI also has an internal 'interface' which maps to the complete set of API functions in the 'fastly' namespace exported by go-fastly, and hundreds of 'mock' functions used in the CLI test suite. Those have also been updated, and along the way many inconsistencies in those function signatures were corrected.

go-fastly version 11 removed a deprecated set of product enablement API functions, so the `fastly products` command has been updated to use the new API functions. The tests for the `fastly products` command no longer mock the API calls as they no longer exist, and the new `fastly product` command which will appear in the next release will provide full test coverage of the CLI's use of the new product enablement API functions.

There are no functional changes in this commit, and the entire test suite still passes.

At least 80% of the changes in this commit were produced using [comby](https://comby.dev/).

 All Submissions:

### Changes to Core Features:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?

### DIFF
--- a/pkg/argparser/flags.go
+++ b/pkg/argparser/flags.go
@@ -1,6 +1,7 @@
 package argparser
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,7 +13,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/fastly/go-fastly/v10/fastly"
+	"github.com/fastly/go-fastly/v11/fastly"
+
 	"github.com/fastly/kingpin"
 
 	"github.com/fastly/cli/pkg/api"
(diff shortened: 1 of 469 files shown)


pkg/argparser/flags.go
```diff
@@ -114,7 +116,7 @@ type OptionalServiceVersion struct {
 
 // Parse returns a service version based on the given user input.
 func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fastly.Version, error) {
-	vs, err := client.ListVersions(&fastly.ListVersionsInput{
+	vs, err := client.ListVersions(context.TODO(), &fastly.ListVersionsInput{
```
Integralist, line 119: Is this PR draft because of these TODO contexts? What's the situation with these?

kpfleming: It was draft because I forgot to click the 'ready for review' button after the CI checks all reported success.

`context.TODO()` is the recommended method for creating an empty `context.Context` object, from the module documentation.

Integralist: Ok no problem. I always thought it was more for during _developing_ a feature. But sounds like the intention is to figure out the right context some further point down the line.

PR #1517, opened 2025-08-15
refactor(domainv1,tools): use updated go-fastly domainmanagement imports and types

Integralist:
All Submissions:

### Changes to Core Features:

### User Impact

There is no impact to the user. The changes as described in this PR are internal only. No package paths have changed, nor have any package names changed, nor have any CLI commands/subcommands changed. The only change has been to the go-fastly calls to match the latest go-fastly interface.

I have an internal conversation going to ascertain if this PR should be doing more (e.g. renaming command from domains-v1 to domain-management, as well as renaming the package path from domainv1 to domainmanagement). But even if we do want to go ahead with those changes we'll likely want to do that in a follow-up PR.


rcaril (approved):
Looks good to me. Further validated locally.

PR #1520, opened 2025-08-19
Followup fixes for Pushpin integration

harmony7:
This PR is a followup to #1509 that includes the following fixes:

* Fix to Pushpin context that inadvertently left out the path to the routes file and the cleanup function when building the context
* Adds an output message indicating when "Experimental Pushpin mode" is enable
* Makes Pushpin-related logs slightly less noisy

### DIFF
--- a/pkg/commands/compute/serve.go
+++ b/pkg/commands/compute/serve.go
@@ -600,7 +600,7 @@ func (c *ServeCommand) GetPushpinProxyPort(out io.Writer) (uint16, error) {
 		}
 		pushpinProxyPort = uint16(pushpinProxyPortInt)
 		if c.Globals.Verbose() {
-			text.Info(out, "Using Pushpin proxy port from --pushpin-proxy-port flag: %d\n\n", pushpinProxyPort)
+			text.Output(out, "DEBUG: Using Pushpin proxy port from --pushpin-proxy-port flag: %d", pushpinProxyPort)
 		}
 		return pushpinProxyPort, nil
 	}
@@ -608,14 +608,14 @@ func (c *ServeCommand) GetPushpinProxyPort(out io.Writer) (uint16, error) {
 	pushpinProxyPort = c.Globals.Manifest.File.LocalServer.Pushpin.PushpinProxyPort
 	if pushpinProxyPort != 0 {
 		if c.Globals.Verbose() {
-			text.Info(out, "Using Pushpin proxy port via `local_server.pushpin.proxy_port` setting: %d\n\n", pushpinProxyPort)
+			text.Output(out, "DEBUG: Using Pushpin proxy port via `local_server.pushpin.proxy_port` setting: %d", pushpinProxyPort)
 		}
 		return pushpinProxyPort, nil
 	}
 
 	pushpinProxyPort = 7677
 	if c.Globals.Verbose() {
-		text.Info(out, "Using default Pushpin proxy port %d\n\n", pushpinProxyPort)
+		text.Output(out, "DEBUG: Using default Pushpin proxy port %d", pushpinProxyPort)
 	}
 	return pushpinProxyPort, nil
 }
@@ -640,7 +640,7 @@ func (c *ServeCommand) GetPushpinPublishPort(out io.Writer) (uint16, error) {
 		}
 		pushpinPublishPort = uint16(pushpinPublishPortInt)
 		if c.Globals.Verbose() {
-			text.Info(out, "Using Pushpin publish handler port from --pushpin-publish-port flag: %d\n\n", pushpinPublishPort)
+			text.Output(out, "DEBUG: Using Pushpin publish handler port from --pushpin-publish-port flag: %d", pushpinPublishPort)
 		}
 		return pushpinPublishPort, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/commands/compute/serve.go
```diff
@@ -600,22 +600,22 @@ func (c *ServeCommand) GetPushpinProxyPort(out io.Writer) (uint16, error) {
 		}
 		pushpinProxyPort = uint16(pushpinProxyPortInt)
 		if c.Globals.Verbose() {
-			text.Info(out, "Using Pushpin proxy port from --pushpin-proxy-port flag: %d\n\n", pushpinProxyPort)
+			text.Output(out, "DEBUG: Using Pushpin proxy port from --pushpin-proxy-port flag: %d", pushpinProxyPort)
```
kpfleming: If this is 'debug' output, it should be conditioned on `c.Globals.Flags.Debug`, not `c.Globals.Verbose`. If it's something that most/all users should see, then `Verbose` is fine but `text.Info` is the proper way to emit it.

pkg/commands/compute/serve.go
```diff
@@ -810,43 +812,76 @@ func (c *pushpinContext) buildPushpinConf() string {
 // command line and/or fastly.toml. The cleanup function on the returned pushpinContext
 // needs to eventually be called by the caller to shut down Pushpin.
 func (c *ServeCommand) startPushpin(spinner text.Spinner, out io.Writer) (pushpinContext, error) {
+	text.Output(out, "%s: %s", text.BoldYellow("EXPERIMENTAL"), "Enabling Pushpin support for local testing of Fanout.")
```
kpfleming: Use `text.Important` instead of duplicating its functionality.

pkg/commands/compute/serve.go
```diff
@@ -974,8 +957,13 @@ func (c *ServeCommand) startPushpin(spinner text.Spinner, out io.Writer) (pushpi
...
 	// Start Pushpin
 	if c.Globals.Verbose() {
-		text.Info(out, "Starting Pushpin runner in the background...")
 		text.Output(out, "%s: %s", text.BoldYellow("Pushpin command"), strings.Join(pushpinCmd.Args, " "))
+		text.Output(out, "%s: %d", text.BoldYellow("Pushpin proxy port"), pushpinCtx.proxyPort)
+		text.Output(out, "%s: %d", text.BoldYellow("Pushpin publisher port"), pushpinCtx.publishPort)
+		text.Output(out, "%s: %d - %d", text.BoldYellow("Pushpin other reserved ports"), pushpinCtx.publishPort+1, pushpinCtx.publishPort+3)
+		text.Output(out, "%s: %s", text.BoldYellow("Pushpin temporary runtime directory"), pushpinCtx.pushpinRunDir)
+		text.Output(out, "%s: %s", text.BoldYellow("Pushpin conf file"), pushpinCtx.confFilePath)
+		text.Output(out, "%s: %s", text.BoldYellow("Pushpin routes file"), pushpinCtx.routesFilePath)
```
kpfleming, line 966: `text.Info` is the proper way to emit these.

pkg/commands/compute/serve.go
```diff
@@ -810,43 +812,76 @@ func (c *pushpinContext) buildPushpinConf() string {
...
+	)
+
+	text.Break(out)
+
+	err = spinner.Start()
 	if err != nil {
 		return pushpinCtx, err
 	}
-	pushpinCtx.publishPort, err = c.GetPushpinPublishPort(out)
+	msg := "Running local Pushpin"
```
kpfleming:
```suggestion
	msg := "Starting Pushpin"
```

pkg/commands/compute/serve.go
```diff
@@ -1015,14 +1003,12 @@ func (c *ServeCommand) startPushpin(spinner text.Spinner, out io.Writer) (pushpi
...
-			Remediation: fmt.Sprintf("A process may already be running on port %d.", pushpinCtx.proxyPort),
+			Remediation: fmt.Sprintf("Check that your disk isn't full and that a process isn't already running on ports %d or %d - %d.", pushpinCtx.proxyPort, pushpinCtx.publishPort, pushpinCtx.publishPort+3),
 		}
 	}
 
-	if c.Globals.Verbose() {
-		text.Info(out, "Local Pushpin started on port %d.", pushpinCtx.proxyPort)
-		text.Break(out)
-	}
+	text.Success(out, "Local Pushpin started.")
```
kpfleming:
```suggestion
	text.Success(out, "Pushpin started.")
```

kpfleming (changes requested):
A few changes, mostly just using the functions from the `text` package for consistency.

pkg/commands/compute/serve.go
harmony7: For these, I did them this way to mirror how these lines were outputting similar headings:

https://github.com/fastly/cli/blob/914f007b804834682ca962d20371ff81c3099ab5/pkg/commands/compute/serve.go#L1084-L1087

pkg/commands/compute/serve.go
harmony7: Globals.Flags.Debug seems to be specifically (at least according to the help text) to display the requests/responses that go over the wire.

I'll use `c.Globals.Verbose()` and revert to `Info`, though because of the colored headings that made (in my opinion) the output a little noisy.

pkg/commands/compute/serve.go
harmony7: text.Important puts the heading IMPORTANT but I wanted the heading EXPERIMENTAL.

For now I'll switch it to `Info`

harmony7:
I've applied changes from review

For reference, here's what the output looks like now:

Without verbose:
image

With verbose:
image

kpfleming:
That looks great!

kpfleming (approved):

PR #1529, opened 2025-08-28
feat(kvstoreentry): Support for missing CRUD attributes

rcaril:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

- New `get` command that only provides the `value` of a `key`
- Updated the `describe` to now only provide the attributes of a key, rather than the value of a key itself

The following features are now supported:

### Create:
- add
- append
- prepend
- metadata
- if_generation_match
- background_fetch

### Delete
- if_generation_match
- force

### Describe
- if_generation_match
- metadata

### Get 
- if_generation_match

## Tests:
```
 make test TEST_ARGS="-run TestCreateCommand ./pkg/commands/kvstoreentry/"
ok      github.com/fastly/cli/pkg/commands/kvstoreentry 1.105s

make test TEST_ARGS="-run TestGetCommand ./pkg/commands/kvstoreentry/"
ok      github.com/fastly/cli/pkg/commands/kvstoreentry 1.083s

 make test TEST_ARGS="-run TestDescribeCommand ./pkg/commands/kvstoreentry/"
ok      github.com/fastly/cli/pkg/commands/kvstoreentry 1.069s

make test TEST_ARGS="-run TestDeleteCommand ./pkg/commands/kvstoreentry/"
ok      github.com/fastly/cli/pkg/commands/kvstoreentry 1.103s
```

### DIFF
--- a/pkg/commands/kvstoreentry/get.go
+++ b/pkg/commands/kvstoreentry/get.go
@@ -58,21 +58,26 @@ func (c *GetCommand) Exec(_ io.Reader, out io.Writer) error {
 		return fsterr.ErrInvalidVerboseJSONCombo
 	}
 
+	// Validate generation value before making API call
+	var inputGeneration uint64
+	if c.Generation != "" {
+		var err error
+		inputGeneration, err = strconv.ParseUint(c.Generation, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid generation value: %s", c.Generation)
+		}
+	}
+
 	result, err := c.Globals.APIClient.GetKVStoreItem(context.TODO(), &c.Input)
 	if err != nil {
 		c.Globals.ErrLog.Add(err)
 		return err
 	}
 
-	// Check if the generation marker provided matches the API.
+	// Check if the generation marker matches the API result
 	if c.Generation != "" {
-		inputGeneration, err := strconv.ParseUint(c.Generation, 10, 64)
 		if inputGeneration != result.Generation {
 			return fmt.Errorf("generation value does not match: expected %d, got %d", result.Generation, inputGeneration)
-		}
-
-		if err != nil {
-			return fmt.Errorf("invalid generation value: %s", c.Generation)
 		}
 	}
 


pkg/commands/kvstoreentry/describe.go
```diff
@@ -46,23 +46,29 @@ func (c *DescribeCommand) Exec(_ io.Reader, out io.Writer) error {
...
-	value, err := c.Globals.APIClient.GetKVStoreKey(context.TODO(), &c.Input)
+	item, err := c.Globals.APIClient.GetKVStoreItem(context.TODO(), &c.Input)
 	if err != nil {
 		c.Globals.ErrLog.Add(err)
 		return err
 	}
 
 	if c.JSONOutput.Enabled {
-		text.Output(out, `{"%s": "%s"}`, c.Input.Key, value)
+		text.Output(out, `{"key": "%s", "generation": "%d", "metadata": "%s"}`, c.Input.Key, item.Generation, item.Metadata)
```
philippschulte: There are a couple of issues with how you print JSON to std out:
- this prints generation as a string (quoted) not a number
- no escaping on metadata – quotes/newlines in metadata will break the JSON

Have a look at other resources how we print JSON to stdout. I think this is the common pattern we use:

```
	if ok, err := c.WriteJSON(out, o); ok {
		return err
	}
```

pkg/commands/kvstoreentry/create.go
```diff
@@ -87,6 +100,23 @@ func (c *CreateCommand) Exec(in io.Reader, out io.Writer) error {
...
+	// Parse generation match if provided.
+	if c.ifGenMatch != "" {
+		inputGeneration, err := strconv.ParseUint(c.ifGenMatch, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid generation value: %s", c.ifGenMatch)
+		}
+		c.Input.IfGenerationMatch = inputGeneration
+	}
+	
+	c.Input.Metadata = &c.metadata
```
philippschulte: This sends an empty-string pointer when the flag isn’t provided which can be different from “not set”. I would also recommend to check the byte limit:

```
if c.metadata != "" {
    if len(c.metadata) > 2000 {
        return <proper error message>
    }
    c.Input.Metadata = &c.metadata
}
```

pkg/commands/kvstoreentry/create.go
```diff
@@ -35,10 +36,16 @@ func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateComman
...
 	c.CmdClause.Flag("dir", "Path to a directory containing individual files where the filename is the key and the file contents is the value").StringVar(&c.dirPath)
 	c.CmdClause.Flag("dir-allow-hidden", "Allow hidden files (e.g. dot files) to be included (skipped by default)").BoolVar(&c.dirAllowHidden)
 	c.CmdClause.Flag("dir-concurrency", "Limit the number of concurrent network resources allocated").Default("50").IntVar(&c.dirConcurrency)
 	c.CmdClause.Flag("file", `Path to a file containing individual JSON objects (e.g., {"key":"...","value":"base64_encoded_value"}) separated by new-line delimiter`).StringVar(&c.filePath)
+	c.CmdClause.Flag("if-generation-match", "Value which must match the current generation marker in an item for an update operation to proceed").StringVar(&c.ifGenMatch)
+	c.CmdClause.Flag("metadata", "An arbitrary data field which can contain up to 2000 bytes of data").StringVar(&c.metadata)
+	c.CmdClause.Flag("prepend", "If an item with the specified key exists, the value provided in the operation is prepended to the existing value instead of replacing it (Default: false)").Default("false").BoolVar(&c.prepend)
```
philippschulte: `bool` flags default `false` by default. Please remove them all because this is redundant.

pkg/commands/kvstoreentry/delete.go
```diff
@@ -92,9 +96,38 @@ func (c *DeleteCommand) Exec(in io.Reader, out io.Writer) error {
...
 	}
 
+	// Check if the generation marker provided matches the API.
+	if c.IfGenerationMatch != "" {
+		getInput := fastly.GetKVStoreItemInput{
+			StoreID: c.StoreID,
+			Key:     c.key.Value,
+		}
+
+		result, err := c.Globals.APIClient.GetKVStoreItem(context.TODO(), &getInput)
```
philippschulte: Did you check if the API provides a server-side conditional for this field? If it's not provided then it should be.

pkg/commands/kvstoreentry/delete.go
```diff
@@ -52,6 +54,8 @@ func NewDeleteCommand(parent argparser.Registerer, g *global.Data) *DeleteComman
 	// Optional.
 	c.CmdClause.Flag("all", "Delete all entries within the store").Short('a').BoolVar(&c.DeleteAll)
 	c.CmdClause.Flag("concurrency", "The thread pool size (ignored when set without the --all flag)").Default(strconv.Itoa(DeleteKeysPoolSize)).Short('r').IntVar(&c.PoolSize)
+	c.CmdClause.Flag("force", "Return a successful result from a 'delete' operation even if the specified key was not found").BoolVar(&c.Force)
+	c.CmdClause.Flag("if-generation-match", "Value which must match the current generation marker in an item for an update operation to proceed").StringVar(&c.IfGenerationMatch)
```
philippschulte: I guess this should be updated to delete?

pkg/commands/kvstoreentry/get.go
```diff
@@ -0,0 +1,106 @@
...
+		text.Output(out, `{"%s": "%s"}`, c.Input.Key, encodedValue)
+		return nil
+	}
+
+	if c.Globals.Flags.Verbose {
+		// We are encoding the value of the key here to ensure safe
+		// output for binary content along with other outputs.
+		encodedValue := base64.StdEncoding.EncodeToString([]byte(value))
+		text.PrintKVStoreKeyValue(out, "", c.Input.Key, encodedValue)
+		return nil
```
philippschulte: Maybe it would be useful to also display the generation in verbose mode?

philippschulte (changes requested):
Overall this is a solid implementation. Please keep in mind that this introduces a breaking change for anyone who previously used `fastly kv-store-entry describe -k foo -s bar` to get the value. Please make sure to state it properly in the changelog. Please add the link to the docs in the PR description. Thank you!

pkg/commands/kvstoreentry/delete.go
rcaril: Honestly, the purpose of this flag is not clear. I'll check with the team on this - assuming you are referring to `force`.

pkg/commands/kvstoreentry/get.go
rcaril: We chatted about this internally and decided that just the attributes should be the new `describe` command, I understand the thinking here however. If we add generation here - perhaps that renders the `describe` command redundant? 

@kpfleming - thoughts on this one?

pkg/commands/kvstoreentry/describe.go
rcaril: Generation was requested to be a string here:
https://fastly.slack.com/archives/C0294PCAS87/p1756236475872529?thread_ts=1756231143.658389&cid=C0294PCAS87

Thanks for the suggestion on the metadata escaping, i'll fix that.

pkg/commands/kvstoreentry/get.go
kpfleming: I don't think this needs a 'verbose' mode at all, since the entire purpose of the `get` command is to get the stored value. There is no 'verbose' method of displaying the stored value.

rcaril: Sounds good, i'll remove this.

rcaril: @kpfleming  - verbose is part of Kingpin, isn't it? So should we just provide a text reply that verbose mode is not available for this command, or just perform no additional action if provided?

pkg/commands/kvstoreentry/delete.go
rcaril: We are doing this check to perform a more friendly response. By default, the API throws a `412` with minimal feedback when a argument fails. This extra check tells the end user what the current generation is without having them need to perform another command to find out. 

https://www.fastly.com/documentation/reference/api/services/resources/kv-store-item/#kv-store-delete-item
-> See **Example 3**

kpfleming: This is unsafe as it creates a race condition. Do an Internet search for "TOCTOU" and you'll learn more than you ever wanted to know about it :-)

If the API responds with 412 and there are multiple reasons why it may have failed but the response does not indicate which one, that will have to be fixed at the API level, not in the CLI.

pkg/commands/kvstoreentry/get.go
kpfleming: It's not part of Kingpin, but it is a global flag for all commands in the CLI. Yes, the code should check to see if verbose mode was requested and emit an error indicating that `get` does not support verbose mode :-)

rcaril: Since throwing an error still resulted in some verbose lines outputted, I added a global bool argument to suppress this information: 

```
kv-store-entry get -s <id> --key testG --verbose
Fastly API token provided via FASTLY_API_TOKEN


ERROR: the 'get' command does not support the --verbose flag.
```
After ->
```
kv-store-entry get -s <id> --key testG --verbose

ERROR: the 'get' command does not support the --verbose flag.

```

rcaril: See [53c9197](https://github.com/fastly/cli/pull/1529/commits/53c919744fc8e12c385d4b641043da8f8d017465) for ref.

pkg/commands/kvstoreentry/delete.go
rcaril: Let's chat more about this next week.

pkg/commands/kvstoreentry/get.go
rcaril: After reviewing the code base as per @philippschulte's suggestion, I did not see any other feature that captures this same suppression for the global `verbose` flag already. 

Kingpin doesn't have any type of Global feature as it is, so we are covered there. When looking at our code for similar features, we do have a `quiet` flag, but this is provided by the end user, not the code base to manipulate the output / suppress a given flag. 

Having the additional granularity of a per command way to disable the verbose flag allows for more flexibility in the future with a simple implementation pattern per command. 

IE:
```
Base: argparser.Base{
			Globals: g,
			// This argument suppresses the 'Fastly API' output from the global verbose command.
			SuppressVerbose: true,
...
```

pkg/commands/kvstoreentry/create.go
philippschulte: I leave this up to you but I would check the the byte length.

pkg/commands/kvstoreentry/delete.go
philippschulte: My only complaint is the use of the word "update" in the context of a `delete` command.

pkg/commands/kvstoreentry/get.go
philippschulte: This looks like a robust and idiomatic solution to me.

philippschulte (approved):

pkg/commands/kvstoreentry/delete.go
rcaril: Ah, yes I see what you mean. I copied this from our docs (https://www.fastly.com/documentation/reference/api/services/resources/kv-store-item/#kv-store-delete-item) which seem to be the source of the typo here. I'll update this in the code base and follow up with the docs team on this correction.

rcaril: Corrected in [142da1f](https://github.com/fastly/cli/pull/1529/commits/142da1fc86ab2c49a7479fc0ab8e38295aa2fdf8)

CHANGELOG.md
```diff
@@ -3,10 +3,15 @@
...
 ### Breaking:
+- breaking(kvstoreentry): The 'describe' command now returns only key attributes (ie: generation, metadata) instead of a given key's value ([#1529](https://github.com/fastly/cli/pull/1529))
 
 ### Enhancements:
 - feat(logging): Add support for 'CompressionCodec' and 'GzipLevel' attribute to the HTTPS endpoint.
 - feat(kvstoreentry): Add support for the 'prefix' parameter for List operations ([#1526](https://github.com/fastly/cli/pull/1526))
+- feat(kvstoreentry): Add support for the add, append, prepend, metadata, if_generation_match, and background_fetch 'create' command operations ([#1529](https://github.com/fastly/cli/pull/1529))
+- feat(kvstoreentry): Add support for the if_generation_match and metadata 'describe' command operations ([#1529](https://github.com/fastly/cli/pull/1529))
+- feat(kvstoreentry): Add support for the if_generation_match and force 'delete' command operations ([#1529](https://github.com/fastly/cli/pull/1529))
+- feat(kvstoreentry): Add support for the if_generation_match 'get' command operation ([#1529](https://github.com/fastly/cli/pull/1529))
```
kpfleming: This PR adds the entire `get` operation, so this line should probably be different.
```suggestion
- feat(kvstoreentry): Add the 'get' command operation which obtains the value of the item ([#1529](https://github.com/fastly/cli/pull/1529))
```

pkg/argparser/cmd.go
```diff
@@ -65,6 +66,11 @@ func (b Base) Name() string {
 	return b.CmdClause.FullCommand()
 }
 
+// SuppressesVerbose returns true if this command should suppress verbose output.
+func (b Base) SuppressesVerbose() bool {
+	return b.SuppressVerbose
```
kpfleming, line 71: Excellent!

pkg/commands/kvstoreentry/get.go
rcaril: @kpfleming - when you have time, this is really the item I wanted another review on.

pkg/commands/kvstoreentry/describe.go
```diff
@@ -46,23 +45,36 @@ func (c *DescribeCommand) Exec(_ io.Reader, out io.Writer) error {
...
+		}
 		return nil
 	}
 
 	if c.Globals.Flags.Verbose {
-		text.PrintKVStoreKeyValue(out, "", c.Input.Key, value)
+		// Print the key attributes.
+		fmt.Fprintf(out, "Key: %s\n", c.Input.Key)
+		fmt.Fprintf(out, "Generation: %d\n", item.Generation)
+		fmt.Fprintf(out, "Metadata: %s\n", item.Metadata)
```
kpfleming: This doesn't seem to be necessary now as it generates the same output as the non-verbose code below; with these changes 'verbose' no longer has any effect for the `describe` command.

pkg/commands/kvstoreentry/get.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid generation value: %s", c.Generation)
+		}
+
+		if inputGeneration != result.Generation {
+			return fmt.Errorf("generation value does not match: expected %d, got %d", result.Generation, inputGeneration)
+		}
+	}
+
+	// Ensure we close the value reader.
```
kpfleming, line 84: Move this above the generation marker check, otherwise the value reader will not be closed if the generation marker check fails. Realistically that doesn't matter because the entire program will exit, but it's good practice anyway.

pkg/commands/kvstoreentry/get.go
```diff
@@ -0,0 +1,106 @@
...
+		valueBytes, err := io.ReadAll(result.Value)
+		if err != nil {
+			c.Globals.ErrLog.Add(err)
+			return err
+		}
+		value = string(valueBytes)
+	}
+
+	if c.JSONOutput.Enabled {
+		// We are encoding the value of the key here to ensure safe
```
kpfleming:
```suggestion
		// We are encoding the value of the item here to ensure safe
```

pkg/commands/kvstoreentry/get.go
kpfleming: Just submitted my review, and the verbose-suppression feature looks like a great solution.

pkg/commands/kvstoreentry/describe.go
rcaril: @kpfleming what would you recommend here then? Provide a error for verbose here as well?

rcaril: If so, I have added the same verbose suppression here as well in [171e85d](https://github.com/fastly/cli/pull/1529/commits/171e85d8eb59f63840142206ba32cf80be2e1b29).

kpfleming (approved):

pkg/commands/kvstoreentry/get.go
rcaril: I misunderstood what you asked here prior, corrected this in[acf2105](https://github.com/fastly/cli/pull/1529/commits/acf210555fff871a1261f85439131585227a4a9d)

PR #1535, opened 2025-09-09
Ensure [local_server.pushpin] section persists during fastly.toml manifest file update

### LINKED ISSUE #1532: Updating fastly.toml loses local_server.pushpin section
**Version**

```
% npx fastly version
Fastly CLI version v11.5.0 (617c3243)
Built with go version go1.24.6 darwin/arm64 (2025-09-08)
Viceroy version: viceroy 0.14.2
```

**What happened**

Invoking something that causes the Fastly CLI to update fastly.toml will wipe out the `local_server.pushpin` section.

harmony7:
### Change summary

This PR updates `MarshalTOML()` for `manifest/File` such that the `[local_server.pushpin]` section and its fields persist when the file is written.

* The `pushpin` section as well as its individual fields are now optional, such that it will be able to persist the file back as it was before starting.
* `[local_server.pushpin]` has been added to `fastly-viceroy-update.toml`, which is used by the test `TestManifestPersistsLocalServerSection`, ensuring that the section is written out and reads back in to the same value.

This fixes #1532.

### Changes to Core Features:

### User Impact

Users can now safely run CLI commands that modify fastly.toml without losing their `[local_server.pushpin]` section.

Since this overwrite happens right after initializing an app from the starter kit, this would be nice to get out asap.
Would be great if this can be released as a quick patch release such as 11.5.1.


rcaril (approved):
The changes looks good to me.

PR #1566, opened 2025-10-27
[CDTOOL-1197] Use bulk purge for 'purge --key' requests

rcaril:
### Change summary

The PR modifies the 'purge --key' command to use the bulk purge API behind the scenes. This was a necessary change due to purge requests with spaces and other characters being serialized due to the nature of the `POST /service/{service_id}/purge/{surrogate_key}` endpoint. 

Internal discussion on this can be found here:
https://fastly.slack.com/archives/C01E7FV8P5H/p1761241147473289 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

This change will not apply any noticeable behavior to the end user. The commands responses will be identical. 

Purge test file has been update to use variables instead of hard coding all test items.

### DIFF
--- a/pkg/commands/purge/root.go
+++ b/pkg/commands/purge/root.go
@@ -166,13 +166,13 @@ func (c *RootCommand) purgeKey(serviceID string, out io.Writer) error {
 		})
 		return err
 	}
-	// The bulk purge endpoint doesn't return a 'Status' field like the single-key
-	// endpoint did. To avoid a breaking change in the CLI output, we hardcode
-	// 'Status: ok' in the success message to maintain consistent behavior.
 	purgeID, ok := m[c.key]
 	if !ok {
 		return fmt.Errorf("no purge ID returned for key: %s", c.key)
 	}
+	// The bulk purge endpoint doesn't return a 'Status' field like the single-key
+	// endpoint did. To avoid a breaking change in the CLI output, we hardcode
+	// 'Status: ok' in the success message to maintain consistent behavior.
 	text.Success(out, "Purged key: %s (soft: %t). Status: ok, ID: %s", c.key, c.soft, purgeID)
 	return nil
 }


pkg/commands/purge/root.go
```diff
@@ -149,6 +149,30 @@ func (c *RootCommand) purgeAll(serviceID string, out io.Writer) error {
...
+			"Service ID": serviceID,
+			"Key":        c.key,
+			"Soft":       c.soft,
+		})
+		return err
+	}
+	// As we are now using the bulk purge endpoint instead of the single purge by key,
+	// 'Status' is no longer returned. To avoid a breaking change, we are emulating this
+	// response by hardcoding 'Status: ok' to a successful resonses.
+	text.Success(out, "Purged key: %s (soft: %t). Status: ok, ID: %s", c.key, c.soft, m[c.key])
```
philippschulte: This assumes `m[c.key]` always exists but maps in Go don’t panic on missing keys — they return the zero value (in this case, an empty string ""), silently. I would recommend to use Go's two-value map lookup to explicitly check:

```
purgeID, ok := m[c.key]
if !ok {
	return fmt.Errorf("no purge ID returned for key: %s", c.key)
}
text.Success(out, "Purged key: %s (soft: %t). Status: ok, ID: %s", c.key, c.soft, purgeID)
```

philippschulte (commented):
Thanks for the fix — I appreciate the switch to the bulk purge endpoint to address key serialization issues. This is definitely the right approach for problematic keys (e.g., keys with spaces or slashes).

However, I’m concerned about one trade-off introduced here: since the bulk purge endpoint doesn’t return a status, we’re now hardcoding "Status: ok" in the CLI output. While this maintains compatibility with prior output, it may mislead users into thinking a purge succeeded when the key wasn’t actually purged.

I’m wondering if we could take a more dynamic approach here:

Detect whether the key is "safe" for the single purge endpoint by checking if key == url.PathEscape(key), and if so, use the original single purge API (/purge/<key>) which returns real status.

Fall back to the bulk purge endpoint for keys that would be altered by encoding.

This would preserve the benefits of the existing implementation while retaining meaningful status for keys that don’t require escaping.

That said, I realize this adds complexity and potentially inconsistent behavior across keys, so I’m deferring the final decision to @kpfleming . If @kpfleming agrees that hardcoding `Status: ok`  is acceptable then it should be clearly documented as assumed, not returned in the CHANGELOG.

pkg/commands/purge/root.go
rcaril: Fixed in [d0b135e](https://github.com/fastly/cli/pull/1566/commits/d0b135e567ac9e9a7d4e3e86ce220a2675e71168).

pkg/commands/purge/root.go
```diff
@@ -149,6 +149,34 @@ func (c *RootCommand) purgeAll(serviceID string, out io.Writer) error {
...
+		c.Globals.ErrLog.AddWithContext(err, map[string]any{
+			"Service ID": serviceID,
+			"Key":        c.key,
+			"Soft":       c.soft,
+		})
+		return err
+	}
+	// As we are now using the bulk purge endpoint instead of the single purge by key,
+	// 'Status' is no longer returned. To avoid a breaking change, we are emulating this
+	// response by hardcoding 'Status: ok' to a successful resonses.
```
kpfleming: There are spelling/grammar errors here, but I'm not providing a suggestion as I'm not quite sure what the intent was...

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 
 ### Bug fixes:
 - fix(compute): clarify fastly.toml error message when file not found ([#1556](https://github.com/fastly/cli/pull/1556))
+- fix(purge/key): single purge requests now use the bulk endpoint to avoid serialization scenarios. ([#1566](https://github.com/fastly/cli/pull/1566))
```
kpfleming: This is a description of the change which was made, but not how it benefits the user. The change description can be found in the commit message, so I'd prefer if this said something like "Ensure that single-key purges will work even if the key contains URL-unsafe characters".

rcaril: Updated in [e88e783](https://github.com/fastly/cli/pull/1566/commits/e88e7838eb3b304c8e4eec9e9b65df882dd9e288).

pkg/commands/purge/root.go
rcaril: Updated to be more clear in [e88e783](https://github.com/fastly/cli/pull/1566/commits/e88e7838eb3b304c8e4eec9e9b65df882dd9e288).

rcaril: Also moved the comment down a few lines in [ed29a1a](https://github.com/fastly/cli/pull/1566/commits/ed29a1a4b493bd871e8e420269e95d911d6cbb80) to ensure that it's obvious as to what the comment is referring to.

PR #1569, opened 2025-10-30
feat(compute/serve): set sig_iss and sig_key to allow client code to test Grip-Sig signing

harmony7:
### Change summary

This PR updates the `pushpin.conf` generated during Pushpin integration to include the `sig_iss` and `sig_key` values:

```ini
[proxy]
sig_iss = viceroy
sig_key = viceroy_dev_signing_key_dont_use_in_production
```

Until now, when running a Compute app locally using the Viceroy / Pushpin bridge, the backend couldn't verify requests signed via the `Grip-Sig` header because the signature was being signed by Pushpin defaults which we didn't document.

This forced developers to maintain separate code paths for local vs. production (`isSigned` would be false in local, true in prod).

By assigning an issuer (`viceroy`) and a deterministic symmetric key, Pushpin now produces a valid `Grip-Sig` JWT locally, allowing the same verification logic to run in both environments.

> [!NOTE]
> Security Note
> The key `viceroy_dev_signing_key_dont_use_in_production`:
> * Is dev-only, used exclusively for local Pushpin -> backend signing.
> * Is deliberately low-entropy and self-descriptive so any accidental usage in production is immediately visible.
> * Does not grant access to any Fastly or customer resources.

(Examples below in JavaScript but origin code can be in any language)

Local test code can now:

```javascript
import { Publisher } from '@fanoutio/grip';

const publisher = new Publisher({
  control_uri: 'http://127.0.0.1:5561/',
  key: undefined,
  verify_iss: 'viceroy',
  verify_key: 'viceroy_dev_signing_key_dont_use_in_production',
});
```

Production would still be:

```javascript
import { Publisher } from '@fanoutio/grip';
import { PUBLIC_KEY_FASTLY_FANOUT_JWK } from '@fanoutio/grip/fastly-fanout`;

const publisher = new Publisher({
  control_uri: `https://api.fastly.com/service/${serviceId}`,
  key: FASTLY_API_TOKEN,
  verify_iss: `fastly:${serviceId}`,
  verify_key: JSON.stringify(PUBLIC_KEY_FASTLY_FANOUT_JWK),
});
```

These values can all be switched out between environments using configuration.

Validation code path:

```javascript
  const { isProxied, isSigned } = await publisher.validateGripSig(gripSig ?? '');
  // isProxied and isSigned are both true now, both in local dev and prod
```

Publish code path:

```javascript
  await publisher.publishHttpStream('<channel>', 'Test Publish!');
```

(note that for local publish, `key` is not used, so it's set to undefined on the publisher config)

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

* [ ] What is the user impact of this change?

None, this enables new functionality.

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,18 @@
 ## [Unreleased]
 
 ### Breaking:
+
+### Enhancements:
+
+### Bug fixes:
+
+### Dependencies:
+
+## [v13.0.0](https://github.com/fastly/cli/releases/tag/v13.0.0) (2025-10-30)
+
+### Breaking:
 - breaking(tls-custom): correct 'tls-custom activation enable' command parameters to reflect expected input from API ([#1562](https://github.com/fastly/cli/pull/1562))
+- breaking(compute/build): Block version 1.91.0 of Rust as it produces broken WASM packages. ([#1571](https://github.com/fastly/cli/pull/1571))
 
 ### Enhancements:
 - feat(compute/serve): set sig_iss and sig_key to allow client code to test Grip-Sig signing ([#1569](https://github.com/fastly/cli/pull/1569))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - breaking(tls-custom): correct 'tls-custom activation enable' command parameters to reflect expected input from API ([#1562](https://github.com/fastly/cli/pull/1562))
 
 ### Enhancements:
+- feat(compute/serve): set sig_iss and sig_key to allow client code to test Grip-Sig signing
```
philippschulte: Please add the link of the PR to the end same as all other entries. Thanks

harmony7: ok! I've added an addition push that does this.

PR #1579, opened 2025-11-17
[CDTOOL-1220] Add Support for NGWAF Virtual Patch CRUD Commands

rcaril:
### Change summary

This PR provides support for the NGWAF Virtual Patch CLI commands. 

 All Submissions:

### New Feature Submissions:

```
make test TEST_ARGS="-run TestVirtualPatchRetrieve ./pkg/commands/ngwaf/virtualpatch"
ok      github.com/fastly/cli/pkg/commands/ngwaf/virtualpatch   1.066s

make test TEST_ARGS="-run TestVirtualPatchList ./pkg/commands/ngwaf/virtualpatch"
ok      github.com/fastly/cli/pkg/commands/ngwaf/virtualpatch   1.059s

 make test TEST_ARGS="-run TestVirtualPatchUpdate ./pkg/commands/ngwaf/virtualpatch"
ok      github.com/fastly/cli/pkg/commands/ngwaf/virtualpatch   1.077s
```

### Changes to Core Features:

### DIFF
--- a/pkg/commands/ngwaf/virtualpatch/retrieve.go
+++ b/pkg/commands/ngwaf/virtualpatch/retrieve.go
@@ -35,7 +35,7 @@ func NewRetrieveCommand(parent argparser.Registerer, g *global.Data) *GetCommand
 	c.CmdClause = parent.Command("retrieve", "Retrieve a vitual patch")
 
 	// Required.
-	c.CmdClause.Flag("virtualpatch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
+	c.CmdClause.Flag("virtual-patch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
 	c.CmdClause.Flag("workspace-id", "Workspace ID").Required().StringVar(&c.workspaceID)
 
 	// Optional.
--- a/pkg/commands/ngwaf/virtualpatch/update.go
+++ b/pkg/commands/ngwaf/virtualpatch/update.go
@@ -38,7 +38,7 @@ func NewUpdateCommand(parent argparser.Registerer, g *global.Data) *UpdateComman
 	c.CmdClause = parent.Command("update", "Update a virtual patch")
 
 	// Required.
-	c.CmdClause.Flag("virtualpatch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
+	c.CmdClause.Flag("virtual-patch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
 	c.CmdClause.Flag("workspace-id", "Workspace ID").Required().StringVar(&c.workspaceID)
 
 	// Optional.
--- a/pkg/commands/ngwaf/virtualpatch/virtualpatch_test.go
+++ b/pkg/commands/ngwaf/virtualpatch/virtualpatch_test.go
@@ -34,17 +34,17 @@ func TestVirtualPatchRetrieve(t *testing.T) {
 	scenarios := []testutil.CLIScenario{
 		{
 			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--virtualpatch-id %s", virtualpatchID),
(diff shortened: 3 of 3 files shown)


pkg/commands/ngwaf/virtualpatch/retrieve.go
```diff
@@ -0,0 +1,73 @@
...
+	c := GetCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+
+	c.CmdClause = parent.Command("retrieve", "Retrieve a vitual patch")
+
+	// Required.
+	c.CmdClause.Flag("virtualpatch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
```
anthony-gomez-fastly: should this be `virtual-patch-id`? genuinely unsure

rcaril: Hmm, I think you are right. Internally we use `virtualPatchId` - so I think the word separation would make sense here. I'll update, good catch!

rcaril: Fixed in [d612932](https://github.com/fastly/cli/pull/1579/commits/d612932382c5ef3402e185305eda70d14b6ea222).

pkg/commands/ngwaf/virtualpatch/update.go
```diff
@@ -0,0 +1,96 @@
...
+	c.CmdClause.Flag("workspace-id", "Workspace ID").Required().StringVar(&c.workspaceID)
+
+	// Optional.
+	c.RegisterFlag(argparser.StringFlagOpts{
+		Name:        "enabled",
+		Description: "Specify the the toggle status indicator of the VirtualPatch.",
+		Action:      c.enabled.Set,
+		Dst:         &c.enabled.Value,
+	})
+	c.CmdClause.Flag("mode", "Specify the action to take when a signal for virtual patch is detected.").Action(c.mode.Set).StringVar(&c.mode.Value)
```
anthony-gomez-fastly, line 51: do we know the valid set of modes? i guess we don't check in go-fastly either so not worth adding extra validation :/

pkg/commands/ngwaf/virtualpatch/update.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+	}
+	c.CmdClause = parent.Command("update", "Update a virtual patch")
+
+	// Required.
+	c.CmdClause.Flag("virtualpatch-id", "Virtual Patch ID").Required().StringVar(&c.virtualpatchID)
+	c.CmdClause.Flag("workspace-id", "Workspace ID").Required().StringVar(&c.workspaceID)
+
+	// Optional.
+	c.RegisterFlag(argparser.StringFlagOpts{
```
anthony-gomez-fastly, line 45: do we want to use `c.RegisterFlagBool` instead? example in `pkg/commands/secretstoreentry/create.go`

pkg/commands/ngwaf/virtualpatch/update.go
rcaril: It's `log` or `block`. From the patterns in CLI, it seems standard to let the API do the validation here. This most likely won't change, so we can add validation in the CLI if you think we should.

pkg/commands/ngwaf/virtualpatch/update.go
rcaril: We use this pattern to ensure that the flag can be set to true or false, other patterns in the CLI allow for a enable or disable via flag provided or not, but this actually disallows for a flag to be altered. 

See internal chat on this here for the last time this came up :) 
https://fastly.slack.com/archives/C08ESC78PGQ/p1756405737545179

pkg/commands/ngwaf/virtualpatch/update.go
anthony-gomez-fastly: makes sense, we're just really inconsistent :/

PR #1582, opened 2025-11-18
Add CRUD functionality for NGWAF lists

anthony-gomez-fastly:
### Change summary

 
Allow users to perform CRUD operations

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/pkg/commands/ngwaf/workspace/countrylist/countrylist_test.go
+++ b/pkg/commands/ngwaf/workspace/countrylist/countrylist_test.go
@@ -65,9 +65,17 @@ func TestCountryListCreate(t *testing.T) {
 			WantError: "error parsing arguments: required flag --name not provided",
 		},
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--entries %s --name %s", listEntries, listName),
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: fmt.Sprintf("--entries %s --name %s", listEntries, listName),
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusNoContent,
+						Status:     http.StatusText(http.StatusNoContent),
+					},
+				},
+			},
+			WantError: "error reading workspace ID: no workspace ID found",
 		},
 		{
 			Name: "validate internal server error",
@@ -123,9 +131,17 @@ func TestCountryListDelete(t *testing.T) {
 			WantError: "error parsing arguments: required flag --list-id not provided",
 		},
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--list-id %s", listID),
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: fmt.Sprintf("--list-id %s", listID),
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusNoContent,
+						Status:     http.StatusText(http.StatusNoContent),
+					},
+				},
+			},
+			WantError: "error reading workspace ID: no workspace ID found",
 		},
 		{
 			Name: "validate bad request",
@@ -185,9 +201,17 @@ func TestCountryListGet(t *testing.T) {
 			WantError: "error parsing arguments: required flag --list-id not provided",
 		},
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--list-id %s", listID),
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: fmt.Sprintf("--list-id %s", listID),
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusNoContent,
+						Status:     http.StatusText(http.StatusNoContent),
+					},
+				},
+			},
+			WantError: "error reading workspace ID: no workspace ID found",
 		},
 		{
 			Name: "validate bad request",
@@ -252,9 +276,17 @@ func TestCountryListList(t *testing.T) {
 
 	scenarios := []testutil.CLIScenario{
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      "",
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: "",
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusNoContent,
+						Status:     http.StatusText(http.StatusNoContent),
+					},
+				},
+			},
+			WantError: "error reading workspace ID: no workspace ID found",
 		},
 		{
 			Name: "validate internal server error",
@@ -339,9 +371,17 @@ func TestCountryListUpdate(t *testing.T) {
 			WantError: "error parsing arguments: required flag --list-id not provided",
 		},
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--list-id %s", listID),
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: fmt.Sprintf("--list-id %s", listID),
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
+						StatusCode: http.StatusNoContent,
+						Status:     http.StatusText(http.StatusNoContent),
+					},
+				},
+			},
+			WantError: "error reading workspace ID: no workspace ID found",
 		},
 		{
 			Name: "validate API success",
--- a/pkg/commands/ngwaf/workspace/iplist/iplist_test.go
+++ b/pkg/commands/ngwaf/workspace/iplist/iplist_test.go
@@ -65,9 +65,17 @@ func TestIPListCreate(t *testing.T) {
 			WantError: "error parsing arguments: required flag --name not provided",
 		},
 		{
-			Name:      "validate missing --workspace-id flag",
-			Args:      fmt.Sprintf("--entries %s --name %s", listEntries, listName),
-			WantError: "error parsing arguments: required flag --workspace-id not provided",
+			Name: "validate missing --workspace-id flag",
+			Args: fmt.Sprintf("--entries %s --name %s", listEntries, listName),
+			Client: &http.Client{
+				Transport: &testutil.MockRoundTripper{
+					Response: &http.Response{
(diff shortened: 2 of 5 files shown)


pkg/commands/ngwaf/accountcountrylist/create.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/fastly/go-fastly/v12/fastly/ngwaf/v1/scope"
+
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/commands/ngwaf/ngwaflist"
+	fsterr "github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// CreateCommand calls the Fastly API to create account level country lists.
```
rcaril: I wonder if we want to update this text to `account-level` to match that of our API verbiage? 

https://www.fastly.com/documentation/reference/api/ngwaf/lists/#ngwafCreateList

pkg/commands/ngwaf/accountcountrylist/create.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
+	c := CreateCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create an account level country list").Alias("add")
```
rcaril: Same nit here.

pkg/commands/ngwaf/ngwaflist/api.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/commands/ngwaf"
+	fsterr "github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/go-fastly/v12/fastly"
+	"github.com/fastly/go-fastly/v12/fastly/ngwaf/v1/lists"
+	"github.com/fastly/go-fastly/v12/fastly/ngwaf/v1/scope"
+)
+
+type ListCreateInput struct {
```
anthony-gomez-fastly, line 15: this file is the main place where logic lives, the setup is that each of these api functions take an input that lets these functions perform the API operation and pass the result back

pkg/commands/ngwaf/accountcountrylist/create.go
anthony-gomez-fastly: fixed in `c0a29a19`

pkg/text/list.go
```diff
@@ -0,0 +1,45 @@
...
+	fmt.Fprintf(out, "Type: %s\n", listToPrint.Type)
+	fmt.Fprintf(out, "Entries: %s\n", strings.Join(listToPrint.Entries, ", "))
+	fmt.Fprintf(out, "Scope: %s\n", listToPrint.Scope.Type)
+	fmt.Fprintf(out, "Updated (UTC): %s\n", listToPrint.UpdatedAt.UTC().Format(time.Format))
+}
+
+// PrintWorkspaceTbl displays workspaces in a table format.
+func PrintListTbl(out io.Writer, listsToPrint []lists.List) {
+	tbl := NewTable(out)
+	tbl.AddHeader("ID", "Name", "Description", "Type", "Scope", "Entries", "Created At")
```
rcaril: Does it make sense to also add the `Updated At` field here? This may be too long of a table if we do this however. I was comparing the output to what we provide in the API / UI, and it seems we provide both the Created and Updated timestamp. 

If we had to chose between the two, perhaps Updated At may be more helpful?

pkg/text/list.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// PrintList displays an NGWAF list.
+func PrintList(out io.Writer, listToPrint *lists.List) {
+	fmt.Fprintf(out, "ID: %s\n", listToPrint.ListID)
+	fmt.Fprintf(out, "Name: %s\n", listToPrint.Name)
+	fmt.Fprintf(out, "Description: %s\n", listToPrint.Description)
+	fmt.Fprintf(out, "Type: %s\n", listToPrint.Type)
+	fmt.Fprintf(out, "Entries: %s\n", strings.Join(listToPrint.Entries, ", "))
+	fmt.Fprintf(out, "Scope: %s\n", listToPrint.Scope.Type)
```
rcaril, line 19: The `scope` field here seems redundant, as the command is already prefixed with `account` or `workspace`.

rcaril: Since we are moving around the command structure, this is probably fine to leave as is now.

pkg/commands/ngwaf/workspace/countrylist/list.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+
+	// Optional.
+	c.RegisterFlagBool(c.JSONFlag())
+
+	return &c
+}
+
+// Exec invokes the application logic for the command.
+func (c *ListCommand) Exec(_ io.Reader, out io.Writer) error {
```
rcaril, line 52: Don't we need to use your parse command here to ensure that the `workspaceID` can be passed in via env var?

```
	if err := c.workspaceID.Parse(); err != nil {
		return err
```
IE:

https://github.com/fastly/cli/blob/34586f7de193afd72df1a42fbc18f15c6e63dfe8/pkg/commands/ngwaf/redaction/create.go#L56

This is why I wound up not using the `Required:    true,` field here:
https://github.com/fastly/cli/blob/34586f7de193afd72df1a42fbc18f15c6e63dfe8/pkg/commands/ngwaf/redaction/create.go#L39

rcaril (changes requested):
Solid implementation, we just need to review the usage of `OptionalWorkspaceID`.

pkg/text/list.go
anthony-gomez-fastly: sure, i'll add both

anthony-gomez-fastly: added in `6ff3587d`

pkg/commands/ngwaf/workspace/countrylist/list.go
anthony-gomez-fastly: I do the parsing in the function on line 70, figured it was better to do it in the generic function

anthony-gomez-fastly: you're right about required, i'll remove it

rcaril: > I do the parsing in the function on line 70, figured it was better to do it in the generic function

Ah I see that now, thanks for pointing that out! 

>you're right about required, i'll remove it

Thanks!

anthony-gomez-fastly: fixed in `4ea7d893`

pkg/commands/ngwaf/workspace/iplist/create.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
+	c := CreateCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create an workspace-level ip list").Alias("add")
```
rcaril:
```suggestion
	c.CmdClause = parent.Command("create", "Create a workspace-level ip list").Alias("add")
```
Small typo

pkg/commands/ngwaf/workspace/stringlist/create.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
+	c := CreateCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create an workspace-level string list").Alias("add")
```
rcaril:
```suggestion
	c.CmdClause = parent.Command("create", "Create a workspace-level string list").Alias("add")
```

pkg/commands/ngwaf/workspace/wildcardlist/create.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
+	c := CreateCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create an workspace-level wildcard list").Alias("add")
```
rcaril:
```suggestion
	c.CmdClause = parent.Command("create", "Create a workspace-level wildcard list").Alias("add")
```

pkg/commands/ngwaf/workspace/countrylist/create.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewCreateCommand returns a usable command registered under the parent.
+func NewCreateCommand(parent argparser.Registerer, g *global.Data) *CreateCommand {
+	c := CreateCommand{
+		Base: argparser.Base{
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create an workspace-level country list").Alias("add")
```
rcaril:
```suggestion
	c.CmdClause = parent.Command("create", "Create a workspace-level country list").Alias("add")
```

pkg/commands/ngwaf/workspace/iplist/delete.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/fastly/go-fastly/v12/fastly/ngwaf/v1/scope"
+
+	"github.com/fastly/cli/pkg/argparser"
+	"github.com/fastly/cli/pkg/commands/ngwaf/ngwaflist"
+	fsterr "github.com/fastly/cli/pkg/errors"
+	"github.com/fastly/cli/pkg/global"
+	"github.com/fastly/cli/pkg/text"
+)
+
+// DeleteCommand calls the Fastly API to delete an account-level ip list.
```
rcaril, line 17:
```suggestion
// DeleteCommand calls the Fastly API to delete an workspace-level ip list.
```

pkg/commands/ngwaf/workspace/countrylist/create.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Required.
+	c.CmdClause.Flag("entries", "Entries for the list. Can either a comma separated list or a path to a file.").Required().StringVar(&c.entries)
```
rcaril: While the other changes are specific, this one needs to be updated across a handful of files, hopefully a find a replace will make this an easy task. 
```suggestion
	c.CmdClause.Flag("entries", "Entries for the list. Can either be a comma separated list or a path to a file.").Required().StringVar(&c.entries)
```

rcaril (approved):
A few typos, but otherwise looks good to me!

pkg/commands/ngwaf/workspace/countrylist/create.go
anthony-gomez-fastly: fixed in `41c877a0`

PR #1611, opened 2026-01-12
Add flag conversion utilities for converting string flags to bools and checking ascending and desecnding flags

anthony-gomez-fastly:
### Change summary
Adds 2 new utilities, one for converting string command line flags to booleans and one for checking `asc` or `desc` ordering flags

 All Submissions:

### New Feature Submissions:


pkg/commands/alerts/list.go
```diff
@@ -114,12 +115,10 @@ func (c *ListCommand) constructInput() (*fastly.ListAlertDefinitionsInput, error
...
+	var err error
 	if c.order.WasSet {
-		switch c.order.Value {
-		case "asc":
-		case "desc":
-			sign = "-"
-		default:
+		sign, err = flagconversion.ConvertOrderFromStringFlag(c.order.Value)
+		if err != nil {
 			err := errors.New("'order' flag must be one of the following [asc, desc]")
```
kpfleming: Would you consider passing the flag name into the conversion function so that it can produce the Error object with the message in it? That would eliminate a lot more duplicated code.

pkg/flagconversion/flag_conversion.go
```diff
@@ -0,0 +1,29 @@
+package flagconversion
```
kpfleming: I'd just these into `argparser/flags.go`, I don't think there's much value in creating a new package for them and they are very closely related to the rest of the code in the existing package.

pkg/commands/alerts/list.go
anthony-gomez-fastly: the main reason i didn't is that it adds an arg only for the name of the flag which is only used in the error, which felt odd. i agree that it removes more code, so i may just swallow that odd feeling and do it.

anthony-gomez-fastly: also we'd have to pass in the command to do `c.Globals.ErrLog.Add(err)` so it's adding 2 args to eliminate 2 lines of code per instance. I think i'm more comfortable leaving the error handling at the command level unless you have strenuous objections

kpfleming: The problem is that the details of what is acceptable live inside the conversion function, and shouldn't be duplicated in the caller's code constructing the error message. I'd be fine with leaving the `c.Globals.ErrLog.Add` function call in the caller, but just eliminating the `errors.New` function call in the caller and passing along the error which came back from the conversion function.

As a somewhat contrived example, what if we decided to improve the 'boolean' flag function to accept true/false, yes/no, or 1/0? We'd have to change the error messages in all of its callers.

anthony-gomez-fastly: makes sense! implemented in `d6847272`

pkg/flagconversion/flag_conversion.go
anthony-gomez-fastly: ahh i forgot where all the utils went and missed this. moved in [8d50353](https://github.com/fastly/cli/pull/1611/commits/8d503533338b65e58f66d6ecd1c404d91ff70694)

PR #1612, opened 2026-01-12
CDTOOL-1260: Introduce 'service purge' command and 'purge' alias.

kpfleming:
### Change summary

This PR adds the 'service purge' command as a replacement for the 'purge' command. The 'purge' command is still available as an alias, but will issue a deprecation warning when it is used.

This involved a number of changes:

1. The existing 'pkg/commands/purge' package content was moved to 'pkg/commands/service/purge', and modified to work as expected in that location.
2. A new 'pkg/commands/alias' package was added to hold alias-only packages.
3. An alias was added in 'pkg/commands/alias/purge'. This alias uses `struct` embedding to avoid duplicating any of the code from the `pkg/commands/service/purge' package. The alias command is 'unlisted' and issues a deprecation warning when it is used.

This PR also fixes the `make fmt` target in the `Makefile` to work properly.

 All Submissions:

### New Feature Submissions:

### User Impact

New deprecation warning for the `purge` command.


pkg/commands/service/purge/purge.go
```diff
@@ -21,10 +21,28 @@ import (
 // CommandName is the string to be used to invoke this command.
 const CommandName = "purge"
 
-// NewRootCommand returns a new command registered in the parent.
-func NewRootCommand(parent argparser.Registerer, g *global.Data) *RootCommand {
-	var c RootCommand
-	c.CmdClause = parent.Command(CommandName, "Invalidate objects in the Fastly cache")
+// PurgeCommand calls the Fastly API to purge items from the cache.
+type PurgeCommand struct { //revive:disable:exported
```
kpfleming, line 25: Since this is a new source file Revive complains about this structure name, but I've suppressed the warning since the code is not new, it was just relocated, and this naming is consistent with other existing commands.

PR #1617, opened 2026-01-14
Add [setup.products] support to fastly compute deploy/publish

harmony7:
This PR adds first-class support for enabling Fastly Products during the initial `fastly compute publish`, driven by a new `[setup.products]` section in `fastly.toml`.

## What this enables

On first publish (service creation only), users can now declaratively enable products such as Fanout, API Discovery, NGWAF, etc., alongside other `[setup]` resources.

Example:

```toml
[setup.products.fanout]
enable = true

[setup.products.ngwaf]
enable = true
workspace_id = "my-workspace"
```

### New manifest schema
- Adds `SetupProducts` under `[setup]`
- Each product has a typed configuration
- NGWAF supports required parameters (`workspace_id`)
- Validation is product-specific and runs before activation

### New setup resource
- Introduces `setup.Products`, integrated into the existing setup lifecycle
- Runs only when creating a new service (consistent with other `[setup]` resources)

### Clean, table-driven implementation
- Both configuration and activation paths are table-driven
- Adding a new product is a small, localized change
- Avoids reflection; product-specific logic remains explicit and readable

### Activation via go-fastly product APIs
- Uses `fastly/products/*` enable APIs
- Activation happens after service creation
- Spinner output and error handling follow existing CLI patterns

### Safe defaults
- Products are enabled only when explicitly configured and `enable = true`
- Disabled or absent products are ignored
- Manifest round-tripping preserves `[setup.products]`

## Review notes

- Main logic: `pkg/commands/compute/setup/products.go`
- Manifest changes: `pkg/manifest/setup.go`
- Deploy flow wires products into the existing setup + creation phases
- Error handling and output are consistent with other setup resources

## Why this change

We already support declarative setup for backends, KV stores, secret stores, etc.  
Products are a natural extension of that model, and this removes the need for manual post-deploy steps or separate CLI invocations when bootstrapping a new service.


philippschulte (approved):
LGTM!

I would only suggest adding more tests:
* to trigger a validation failure
* that verifies disabled or omitted products are ignored

But I keep this up to you! Thanks!

pkg/commands/compute/setup/products.go
```diff
@@ -0,0 +1,391 @@
...
+	return &Product{Enable: true}
+}
+
+var _ ProductSettings = (*Product)(nil)
+
+func (p *Product) Enabled() bool {
+	return p != nil && p.Enable
+}
+
+type ProductNgwaf struct {
```
anthony-gomez-fastly: nit: I think we've had this written as `NGWAF` in most places in the code

pkg/commands/compute/deploy.go
```diff
@@ -1013,6 +1033,17 @@ func (c *DeployCommand) CreateServiceResources(
...
+	if err := sr.products.Create(); err != nil {
+		c.Globals.ErrLog.AddWithContext(err, map[string]any{
+			"Accept defaults": c.Globals.Flags.AcceptDefaults,
+			"Auto-yes":        c.Globals.Flags.AutoYes,
+			"Non-interactive": c.Globals.Flags.NonInteractive,
+			"Service ID":      serviceID,
+			"Service Version": serviceVersion,
+		})
+		return err
+	}
```
philippschulte, line 953: This fixes the test and passes CI:

```Go
if sr.products.Predefined() {
	if err := sr.products.Create(); err != nil {
		c.Globals.ErrLog.AddWithContext(err, map[string]any{
			"Accept defaults": c.Globals.Flags.AcceptDefaults,
			"Auto-yes":        c.Globals.Flags.AutoYes,
			"Non-interactive": c.Globals.Flags.NonInteractive,
			"Service ID":      serviceID,
			"Service Version": serviceVersion,
		})
		return err
	}
}
```

pkg/commands/compute/setup/products.go
```diff
@@ -0,0 +1,391 @@
...
+		if err := spec.configure(p.Stdout, &p.required, product); err != nil {
+			return fmt.Errorf("%s: %w", "setup.products."+spec.id, err)
+		}
+	}
+
+	return nil
+}
+
+// Create calls the relevant API to create the service resource(s).
+func (p *Products) Create() error {
```
philippschulte, line 374: At the very beginning I would call a helper function which manually checks and iterates specs and if nothing is enabled returns early. Something like:

```Go
func (p *Products) Create() error {
    if !anyEnabled(&p.required) {
        return nil
    }
    ...
}
```

pkg/manifest/setup.go
```diff
@@ -87,3 +92,60 @@ type SetupSecretStoreEntry struct {
...
+	for i := 0; i < rv.NumField(); i++ {
+		fv := rv.Field(i)
+		if fv.Kind() != reflect.Ptr || fv.IsNil() {
+			continue
+		}
+
+		if fv.Type().Implements(settingsT) {
+			return true
+		}
+	}
```
philippschulte: I think `SetupProducts.AnyDefined()` can be explicit and clearer:

```Go
func (p *SetupProducts) AnyDefined() bool {
    return p != nil && (
        p.APIDiscovery != nil ||
        p.BotManagement != nil ||
        ...
        p.Ngwaf != nil ||
        ...
    )
}
```

pkg/mock/client.go
```diff
@@ -58,6 +58,37 @@ func (c *HTTPClient) Do(r *http.Request) (*http.Response, error) {
...
+}
+
+// NewHTTPClientWithResponses returns a mock HTTP Client that returns stubbed response and
+// no errors, and saves requests.
+func NewHTTPClientWithResponses(res []*http.Response) *HTTPClient {
+	return NewHTTPClientDefault(res, make([]error, len(res)))
+}
+
+// NewHTTPClientWithErrors returns a mock HTTP Client that returns stubbed response and
+// no errors, and saves requests.
```
philippschulte: I don't think the comment matches the function's signature.

philippschulte (changes requested):
`TestDeploy_ActivateBeacon` fails with:

> want "error activating version:", have "failed to convert interface to a fastly client"

In `pkg/commands/compute/deploy.go`, you added:

```Go
if err := sr.products.Create(); err != nil { ... }
```
The problem is that this statement is **unconditional**!

pkg/mock/client.go
harmony7: nice catch, thanks

pkg/commands/compute/deploy.go
harmony7: I looked at it again and I think that's actually not the problem.  The problem is that in a test,
`p.APIClient` is not `*fastly.Client`, so this cast fails:

```
fc, ok := p.APIClient.(*fastly.Client)
```

For this particular test, the `anyEnabled` I'm adding next should fix it

philippschulte (approved):
Thank you!

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@
 - feat(compute/build): Block version 1.93.0 of Rust to avoid a wasm32-wasip2 bug. ([#1653](https://github.com/fastly/cli/pull/1653))
 - feat(service/vcl): escape control characters when displaying VCL content for cleaner terminal output ([#1637](https://github.com/fastly/cli/pull/1637))
 
+- feat(compute/deploy): Apply \[setup.products] for enabling products during initial deploy ([#1617](https://github.com/fastly/cli/pull/1617))
```
kpfleming:
```suggestion
- feat(compute/deploy): Apply \[setup.products] for enabling products during initial deploy or publish ([#1617](https://github.com/fastly/cli/pull/1617))
```

kpfleming (changes requested):
I'm satisfied with the level of testing currently present in the PR - I don't think we need to test to ensure that the correct API calls are issued (with the correct attributes), as we don't currently have a good way to do that and this functionality can invoke quite a number of API calls.

The only changes I'd make at this point are to add support for the `mode` parameter for DDoS Protection and the `contentguard` parameter for Bot Management, and then this should be ready to proceed.

harmony7:
I've updated to go-fastly 15 and added support for "mode" in DDOS Protection
(content guard doesn't seem to have a setup value in its enable function in go-fastly)

kpfleming:
That's correct, ContentGuard is enabled/disabled by using the `configure` operation after the `enable` operation has been completed. We hide this in the Terraform provider by making it appear all as one block, but the code invokes two API operations to make the changes.

pkg/commands/compute/setup/products.go
```diff
@@ -0,0 +1,427 @@
...
+	required ProductsMap
+}
+
+// ProductsMap represents the configuration parameters for enabling specified products
+// for a service.
+type ProductsMap struct {
+	APIDiscovery        ProductSettings
+	BotManagement       ProductSettings
+	BrotliCompression   ProductSettings
+	DdosProtection      ProductSettings
```
kpfleming: This is really a nit, but if we are going to maintain the capitalization of 'API' and 'NGWAF', then we should maintain it for 'DDOS' too.

pkg/manifest/setup.go
```diff
@@ -88,3 +91,59 @@ type SetupSecretStoreEntry struct {
 	// values are input during setup.
 	Description string `toml:"description,omitempty"`
 }
+
+type SetupProducts struct {
+	APIDiscovery        *SetupProduct      `toml:"api_discovery,omitempty"`
+	BotManagement       *SetupProduct      `toml:"bot_management,omitempty"`
+	BrotliCompression   *SetupProduct      `toml:"brotli_compression,omitempty"`
+	DdosProtection      *SetupProduct      `toml:"ddos_protection,omitempty"`
```
kpfleming:
```suggestion
	DdosProtection      *SetupProductDdosProtection      `toml:"ddos_protection,omitempty"`
```

pkg/manifest/testdata/fastly-viceroy-update.toml
```diff
@@ -75,3 +75,8 @@ file = "/path/to/other/secret.json"
 [[local_server.secret_stores.store_two]]
 key = "fourth"
 env = "ENV_FOURTH"
+
+[setup]
+[setup.products]
```
kpfleming, line 80: This should have blocks for NGWAF and DDOS Protection (and Bot Management once it has support for ContentGuard), so that the parsers for those products will be tested.

kpfleming (changes requested):

pkg/manifest/setup.go
harmony7: yeah thanks for catching that

pkg/manifest/testdata/fastly-viceroy-update.toml
harmony7: The viceroy-update test was only testing that `[local_server]` doesn't lose anything

so I added a new test that also tests `[setup]`

pkg/commands/compute/setup/products.go
harmony7: Thanks, I've applied this change.

pkg/manifest/testdata/fastly-viceroy-update.toml
harmony7: And if I had this test in there I would have caught the`SetupProductDdosProtection` item that I had missed above.

harmony7:
thanks for your review, please re-review

kpfleming (approved):

kpfleming:
I'm going to merge this so it can go into today's release, but a follow-up PR should add support for ContentGuard in the Bot Management product.

PR #1630, opened 2026-01-22
CDTOOL-1254 convert dictionary to service dictionary

anthony-gomez-fastly:
### Change summary
Convert the `dictionary` to `service dictionary` and add alias for deprecated `dictionary` command

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/pkg/commands/commands.go
+++ b/pkg/commands/commands.go
@@ -1374,7 +1374,6 @@ func Define( // nolint:revive // function-length
 		aliasBackendDescribe,
 		aliasBackendList,
 		aliasBackendUpdate,
-		aliasDictionaryRoot,
 		aliasDictionaryCreate,
 		aliasDictionaryDelete,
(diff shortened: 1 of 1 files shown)


pkg/commands/commands.go
```diff
@@ -1367,6 +1374,12 @@ func Define( // nolint:revive // function-length
 		aliasBackendDescribe,
 		aliasBackendList,
 		aliasBackendUpdate,
+		aliasDictionaryRoot,
```
rcaril: I believe we don't need the `aliasDictionaryRoot` root to be defined here. Since these are only aliases and running 'fastly dictionary' by itself doesn't actually do anything except tell the user to select a subcommand.

anthony-gomez-fastly: fixed in `24082ac9`

PR #1640, opened 2026-01-27
[CDTOOL-691] Better error reporting when JavaScript tools are missing

jedisct1:
### Change summary

 

Validate JavaScript toolchains before attempting a build, catching common setup issues early and providing accurate error messages instead of build failures.

As Bun becomes increasingly popular, it’s not ideal to ask users to install Node when they already have Bun. So, this PR also adds proper support for Bun as an alternative runtime.

The verification detects whether a project uses Node.js or Bun by checking for lockfiles, with support for Bun workspaces where the lockfile lives at the workspace root rather than in the subpackage.

When something is missing, the error message now explains exactly what's wrong and how to fix it, whether that's installing a runtime, running npm install, or adding the @fastly/js-compute package.
 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Use get better guidance and can use Bun as an alternative Runtime.

No breaking changes.


pkg/commands/compute/language_javascript.go
```diff
@@ -140,16 +161,16 @@ func (j *JavaScript) Dependencies() map[string]string {
 // Build compiles the user's source code into a Wasm binary.
 func (j *JavaScript) Build() error {
 	if j.build == "" {
-		j.build = JsDefaultBuildCommand
-		j.defaultBuild = true
-
-		usesWebpack, err := j.checkForWebpack()
-		if err != nil {
+		// Only verify toolchain when using default build (no custom [scripts.build])
```
kpfleming: This could be an issue, since most (possibly all) of our JavaScript starter kits include a `scripts.build` property, and that means users using those starter kits won't benefit from any of this verification logic.

It may be possible to do the verification even if `scripts.build` is set, as long as its value is exactly `npm run build` which is what we have in the starter kits.

pkg/commands/compute/language_javascript.go
```diff
@@ -254,3 +275,351 @@ type NPMPackage struct {
...
+			Inner: nodeErr,
+			Remediation: `Node.js is installed but npm is missing.
+
+Install npm (usually bundled with Node.js):
+  - Reinstall Node.js from https://nodejs.org/
+  - Or install npm separately: https://docs.npmjs.com/downloading-and-installing-node-js-and-npm
+
+Verify: npm --version
+
+Then retry: fastly compute build`,
```
kpfleming: In each case where a 'retry' suggestion is provided we may need to do something more complicated, because the user may not have run `fastly compute build` - they may have run `fastly compute publish` or `fastly compute serve`, both of which internally run `fastly compute build`.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - feat(service/ratelimit): moved the `rate-limit` commands under the `service` command, with an unlisted and deprecated alias of `rate-limit` ([#1632](https://github.com/fastly/cli/pull/1632))
 - feat(compute/build): Remove Rust version restriction, allowing 1.93.0 and later versions to be used. ([#1633](https://github.com/fastly/cli/pull/1633))
 - feat(service/resourcelink): moved the `resource-link` commands under the `service` command, with an unlisted and deprecated alias of `resource-link` ([#1635](https://github.com/fastly/cli/pull/1635))
+- feat(compute/build): improved error messaging for JavaScript builds with pre-flight toolchain verification including Bun runtime support
```
kpfleming: Please add the PR link to this line, following the example of the other CHANGELOG lines.

kpfleming (commented):
Only some comments for now, overall this looks like a nice improvement. I'll ask @harmony7 to also review this since he's our JavaScript runtime expert.

CHANGELOG.md
jedisct1: Added, thanks!

pkg/commands/compute/language_javascript.go
jedisct1: Indeed! The help messages have been updated to be more generic.

pkg/commands/compute/language_javascript.go
jedisct1: Users of the starter kits should now benefit from it!

kpfleming:
Pinging @harmony7, this is still waiting for a review.

pkg/commands/compute/language_javascript.go
```diff
@@ -83,13 +98,19 @@ type JavaScript struct {
 	manifestFilename string
 	// metadataFilterEnvVars is a comma-separated list of user defined env vars.
 	metadataFilterEnvVars string
+	// nodeModulesDir is the resolved path to node_modules (may be in parent dir for monorepos).
+	nodeModulesDir string
```
harmony7: In JavaScript, `node_modules` can be in any directory up the directory tree.

For example,

```
root/
  node_modules/
    @fastly/js-compute/ - Fastly Compute SDK
    shared_lib1/
  app1/
    node_modules/
      lib1/
    src/
      index.js - Entry point
```

In this case, when compiling `app1/src/index.js` the SDK will find `lib1`, `shared_lib1`, and `@fastly/js-compute` just fine, because JavaScript will "merge" node_modules like that.

Since you're using the `nodeModulesDir` to find `@fastly/js-compute`, you should be able to account for such a case. This might be better as not a single string but as an array or something.

pkg/commands/compute/language_javascript.go
```diff
@@ -254,3 +290,352 @@ type NPMPackage struct {
...
+
+	if j.verbose {
+		text.Info(j.output, "Found package.json at %s\n", pkgPath)
+		text.Info(j.output, "Found node_modules at %s\n", nodeModulesPath)
+	}
+	return nil
+}
+
+// verifyWebpackInstalled checks that webpack is installed if used.
+func (j *JavaScript) verifyWebpackInstalled() error {
```
harmony7: I didn't realize checking for webpack was a step in this validation. There was a time in the past when we did require webpack with our SDK, but that is no longer the case.

In most cases webpack is no longer necessary, and in the few cases it's desired, the application author adds it themself. They are free to use webpack, esbuild, or any desired bundler, and [we provide a documentation page](https://www.fastly.com/documentation/guides/compute/developer-guides/javascript/#module-bundling) on how to do this. Our starter kits also don't use webpack anymore except for one that illustrates how to add it.

So I think we can remove `verifyWebpackInstalled()`, as well as stop automatically running webpack when it's present. Running `webpack` should be done explicitly by the author as described in the docs, if it's desired.

pkg/commands/compute/language_javascript_test.go
```diff
@@ -0,0 +1,612 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	fsterr "github.com/fastly/cli/pkg/errors"
+)
+
+// createFakeRuntime creates a fake executable that outputs the given string.
+func createFakeRuntime(t *testing.T, dir, name, output string) {
```
harmony7, line 15: This is a neat trick! I like it =)

harmony7 (changes requested):
Thank you for putting this together!

I want to bring up a couple of points, I've placed comments inline. Since we're in here now improving this part, I think it's probably worth tidying these points up too.  Please let me know if you have any thoughts.

pkg/commands/compute/language_javascript.go
jedisct1: Good call. I removed all the webpack-specific logic. That simplifies things quite a bit :)

pkg/commands/compute/language_javascript.go
jedisct1: `findAllNodeModules` walks up the directory tree collecting every `node_modules` it finds.

A hoisted `@fastly/js-compute` in a parent `node_modules` dir will be found correctly.

pkg/commands/compute/language_javascript_test.go
jedisct1: :)

pkg/commands/compute/language_javascript.go
```diff
@@ -176,81 +197,331 @@ func (j *JavaScript) Build() error {
...
 }
 
-// NPMPackage represents a package.json manifest and its dependencies.
-type NPMPackage struct {
-	DevDependencies map[string]string `json:"devDependencies"`
-	Dependencies    map[string]string `json:"dependencies"`
+// getDefaultBuildCommand returns the appropriate build command for the detected runtime.
+func (j *JavaScript) getDefaultBuildCommand() string {
+	if j.runtime != nil && j.runtime.PkgMgr == "bun" {
+		return fmt.Sprintf("bunx js-compute-runtime ./src/index.js %s", binWasmPath)
```
harmony7: I would like to suggest we move this `fmt.Sprintf("bunx js-compute-runtime ./src/index.js %s", binWasmPath)` to a named constant like we have for `JsDefaultBuildCommand`.  As a reader of the code I think it's easier on the eyes to have the consistency between the two cases here.

harmony7 (approved):
The changes look great!

I have one last comment, which can be considered non-blocking. Please review it, make the change if you think appropriate, and then go ahead and merge.

Thanks again for this great PR.

pkg/commands/compute/language_javascript.go
jedisct1: That would indeed be cleaner and more consistent. Done!

anthony-gomez-fastly (approved):

PR #1654, opened 2026-02-13
Dockerfile-rust: use base image toolchain instead of reinstalling stable

backkem:
### Change summary

`rustup toolchain install stable` re-downloads the latest stable toolchain,
which may differ from the pinned `rust:` base image version. This seems
redundant and is somewhat non-obvious if you want to pin the Rust version.
(e.g. when pinning Rust 1.92 since the current latest CLI panics with stable:
"version '1.93.1' of Rust has not been validated for use with Fastly Compute.").

Since the base image already ships with the correct toolchain, just add
the wasm target directly.

All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/cli/pulls) for the
same update/change?

### User Impact

None — only affects the Docker build image, not the CLI itself.

No. Technically this Dockerfile still fails to run CLI v13.3.0 unless you pin the Rust version. This at least centralizes the latter.

### DIFF
--- a/Dockerfile-rust
+++ b/Dockerfile-rust
@@ -1,9 +1,8 @@
 FROM rust:latest
 LABEL maintainer="Fastly OSS <oss@fastly.com>"
 
-ENV RUST_TOOLCHAIN=stable
-RUN rustup toolchain install ${RUST_TOOLCHAIN} \
-  && rustup target add wasm32-wasip1 --toolchain ${RUST_TOOLCHAIN} \
+ENV RUSTUP_TOOLCHAIN=$RUST_VERSION
+RUN rustup target add wasm32-wasip1 \
   && apt-get update && apt-get install -y curl jq && apt-get -y clean && rm -rf /var/lib/apt/lists/* \
   && cargo install wasm-tools --locked \
   && export FASTLY_CLI_VERSION=$(curl -s https://api.github.com/repos/fastly/cli/releases/latest | jq -r .tag_name | cut -d 'v' -f 2) \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Bug Fixes:
+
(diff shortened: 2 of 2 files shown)


Dockerfile-rust
```diff
@@ -1,9 +1,8 @@
 FROM rust:latest
 LABEL maintainer="Fastly OSS <oss@fastly.com>"
 
-ENV RUST_TOOLCHAIN=stable
-RUN rustup toolchain install ${RUST_TOOLCHAIN} \
-  && rustup target add wasm32-wasip1 --toolchain ${RUST_TOOLCHAIN} \
+ENV RUSTUP_TOOLCHAIN=$RUST_VERSION
```
anthony-gomez-fastly, line 4: i'm still learning rust, does this var need to be set in the dockerfile or does it come from the host system?

backkem: `RUST_VERSION` is defined in the upstream Rust container image. E.g.:

https://github.com/rust-lang/docker-rust/blob/cc333dda42f949065ae7bf90f28b29e6044a4d44/stable/trixie/Dockerfile#L5-L8

`RUSTUP_TOOLCHAIN` is used to pin the toolchain so no other version is installed at runtime:
https://rust-lang.github.io/rustup/environment-variables.html?highlight=RUSTUP_TOOLCHAIN#environment-variables

> RUSTUP_TOOLCHAIN (default: none). If set, will [override](https://rust-lang.github.io/rustup/overrides.html) the toolchain used for all rust tool invocations. A toolchain with this name should be installed, or invocations will fail. This can specify custom toolchains, installable toolchains, or the absolute path to a toolchain.

Combined this ensures the entire build process remains pinned to the toolchain version of the upstream Rust container image. This ensures:
* No additional Rust toolchain is being downloaded at container image build time.
* We have controle over which toolchain is will actually be used at runtime to make sure it doesn't accidentally conflict with the `toolchain_constraint ` (causing flaky builds):
https://github.com/fastly/cli/blob/33f76bc65ded89b2a06e4629cd350615404f5e0b/.fastly/config.toml#L21

PR #1767, opened 2026-04-28
Add Tools `go-releaser` CI Job

rcaril:
### Change summary

Since we skip any release steps for all PRs that aren't tagged as a release, we need to add a CI job to test `goreleaser` tools builds against Dependabot PRs. This ensures that future dependency PRs opened will validate that they do not conflict with the current version of `goreleaser` to prevent scenarios like #1762.  

 All Submissions:

### New Feature Submissions:

### Notes:

We are only testing against `ubuntu-latest` as we are just looking for collision errors, which should be available on any architecture. So this should be sufficient for this use case.


anthony-gomez-fastly (approved):
TY!

PR #1773, opened 2026-05-01
Add support for cpp in for compute

anthony-gomez-fastly:
### Change summary
Adds ability to build C++ services and publish them

test service [here](https://manage.fastly.com/compute/services/RYOnZ5b5b3LaYeeIkeMmLp) 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:


.github/workflows/pr_test.yml
```diff
@@ -94,6 +94,27 @@ jobs:
...
+          tar xzf wasi-sdk-25.0-arm64-macos.tar.gz
+          echo "$(pwd)/wasi-sdk-25.0-arm64-macos/bin" >> $GITHUB_PATH
+        shell: bash
+      - name: "Install WASI SDK (Windows)"
+        if: matrix.platform == 'windows-latest'
+        run: |
+          Invoke-WebRequest -Uri "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-x86_64-windows.tar.gz" -OutFile "wasi-sdk.tar.gz"
+          tar -xzf wasi-sdk.tar.gz
+          echo "$PWD/wasi-sdk-25.0-x86_64-windows/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        shell: pwsh
```
kpfleming, line 186: If it's practical, please move these version numbers to workflow-level variables, so that changing them in the future will have lower risk of making a mistake.

pkg/commands/compute/build_test.go
```diff
@@ -662,6 +662,209 @@ func TestBuildJavaScript(t *testing.T) {
...
+		},
+		{
+			name: "unknown language",
+			args: args("compute build"),
+			fastlyManifest: `
+			manifest_version = 2
+			name = "test"
+			language = "foobar"`,
+			wantError: "unsupported language foobar",
+		},
```
kpfleming: These are not language-specific so probably shouldn't be duplicated here; if they are already duplicated for the other languages, maybe they should be moved to their own test function which is gated solely on `TEST_COMPUTE_BUILD`.

fastly-dev
kpfleming, line 1: What is this file?

kpfleming (dismissed):
One extraneous file, and a couple of 'nice to have' improvements. Otherwise this looks great.

philippschulte (dismissed):
Some of my findings are the same as Kevin's:

1. Please remove the committed `fastly-dev` binary from the repo root.
2. `cpp` is now exposed in `compute init`, but I do not see matching C++ starter-kit config. This could break the init flow if the starter-kit list is empty.
3. The new `config.CPP` defaults do not appear to be added to the static/generated config, so real users may get empty C++ config values.
4. `WasmWasiTarget` is defined but not used; either use it in the default build command or remove it for now.
5. Add an init test for `--language cpp`, especially covering the no-starter-kit case.

kpfleming:
> 2\. `cpp` is now exposed in `compute init`, but I do not see matching C++ starter-kit config. This could break the init flow if the starter-kit list is empty.

We don't have starter kits yet, that will happen next week. We do know what their URLs will be though, if those need to be included in the build.

fastly-dev
anthony-gomez-fastly: oh this was the sample i built, removing

pkg/commands/compute/build_test.go
anthony-gomez-fastly: done in `c5112ee6`

.github/workflows/pr_test.yml
anthony-gomez-fastly: fixed in `387422a7`

anthony-gomez-fastly:
@philippschulte 

1. Please remove the committed fastly-dev binary from the repo root.
Done in `d5685ea`
2. cpp is now exposed in compute init, but I do not see matching C++ starter-kit config. This could break the init flow if the starter-kit list is empty.
Kevin addressed this already, I have included a test that has the starter kit URL, which fails now but should pass once we're ready and the url is public.
3. The new config.CPP defaults do not appear to be added to the static/generated config, so real users may get empty C++ config values.
I added them to fastly.toml in `c1dc0443`
4. WasmWasiTarget is defined but not used; either use it in the default build command or remove it for now.
it is used in the build process from what i understand, as demonstarted in the config.
5. Add an init test for --language cpp, especially covering the no-starter-kit case.
done in `c632e09`

pkg/commands/compute/init_test.go
```diff
@@ -42,6 +42,13 @@ func TestInit(t *testing.T) {
 			Branch: "main",
 		},
 	}
+	skCPP := []config.StarterKit{
+		{
+			Name:   "Default",
+			Path:   "https://github.com/fastly/compute-starter-kit-cpp-default",
```
kpfleming, line 48: There will also be an `-empty` starter kit, as there is for the other languages.

kpfleming (approved):
One very small change and this is ready to go; the tests will fail until the starter kits are made public, so we can either just accept that for a couple of weeks or skip them for now.

philippschulte (commented):
Thanks for the updates. Point 4 still looks unresolved to me.

I see `wasm_wasi_target` is now present in `.fastly/config.toml` and mapped onto `config.CPP.WasmWasiTarget`, but I do not see that value being read anywhere in the C++ build path. The default build command still hardcodes `wasm32-wasip1`:

```go
var CPPDefaultBuildCommand = fmt.Sprintf("clang++ -O3 --target=wasm32-wasip1 -o %s main.cpp", binWasmPath)
```

and `Build()` assigns that constant directly when `[scripts.build]` is missing.

So unless I am missing another code path, `WasmWasiTarget` is configured but currently unused.

anthony-gomez-fastly:
@philippschulte you're right, i missed the hardcoded value not being used, added support in `02965bcb`

pkg/commands/compute/init_test.go
anthony-gomez-fastly: added in `02965bcb`

philippschulte (approved):
Thank you!

jedisct1 (approved):

PR #1774, opened 2026-05-05
CDTOOL-1328 list version refactor

anthony-gomez-fastly:
### Change summary

  Performance Optimization - Skip ListVersions API call for numeric versions                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                       
  Changes to pkg/argparser/flags.go:                                                                                                                                                                                                                                                   
                                                            
  1. OptionalServiceVersion.Parse() - Skip API call for numeric versions                                                                                                                                                                                                               
    - Added early return when user provides numeric version (e.g., --version 1)
    - Returns &fastly.Version{Number: n} directly without calling ListVersions API                                                                                                                                                                                                     
    - Saves 1 API call per command when numeric version is used                                                                                                                                                                                                                        
    - Non-numeric versions (e.g., "latest", "active") still call API as before                                                                                                                                                                                                         
  2. OptionalAutoClone.Parse() - Handle unknown version state                                                                                                                                                                                                                          
    - Added stateUnknown check: v.Active == nil && v.Locked == nil                                                                                                                                                                                                                     
    - Updated autoclone condition to include stateUnknown case                                                                                                                                                                                                                         
    - Effect: When --autoclone is provided with numeric version, always clones (defensive behavior) since state is unknown                                                                                                                                                             
    - Maintains safety while benefiting from optimization                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                       
  Changes to pkg/argparser/flags_test.go:                                                                                                                                                                                                                                              
  - Updated test "specific version not in list" - now expects success (version 5) instead of error                                                                                                                                                                                     
  - Added explicit Active/Locked fields to "version is editable" test      

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### DIFF
--- a/pkg/argparser/flags.go
+++ b/pkg/argparser/flags.go
@@ -166,8 +166,22 @@ func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fast
 			return fastly.ToValue(vs[i].Number) > fastly.ToValue(vs[j].Number)
 		})
 		return vs[0], nil
+	case "staged":
+		serviceDetails, err := client.GetServiceDetails(context.TODO(), &fastly.GetServiceDetailsInput{
+			ServiceID: sid,
+			Filters: []fastly.ServiceDetailsFilter{
+				{Key: "versions.staged", Value: true},
+			},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("error getting service details: %w", err)
+		}
+		if serviceDetails.Version == nil {
+			return nil, fmt.Errorf("no staged service version found")
+		}
+		return serviceDetails.Version, nil
 	default:
-		return nil, fmt.Errorf("invalid version value %q: must be a version number, \"latest\", or \"active\"", sv.Value)
+		return nil, fmt.Errorf("invalid version value %q: must be a version number, \"latest\", \"active\", or \"staged\"", sv.Value)
 	}
 }
 
--- a/pkg/argparser/flags.go
+++ b/pkg/argparser/flags.go
@@ -115,6 +115,13 @@ type OptionalServiceVersion struct {
 }
 
 // Parse returns a service version based on the given user input.
+//
+// Supported values:
+//   - Numeric version (e.g., "1", "2", "42"): Returns the specified version
+//   - "active": Returns the currently active version
+//   - "staged": Returns the currently staged version
+//   - "latest": Returns the highest version number (latest version)
+//   - Omitted (no flag provided): Returns active version, falls back to latest if no active version exists
 func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fastly.Version, error) {
 	// When no --version flag is provided (WasSet=false), default to "active" to preserve
 	// the original behavior of trying active version first, with fallback to latest.
--- a/pkg/argparser/flags.go
+++ b/pkg/argparser/flags.go
@@ -116,10 +116,10 @@ type OptionalServiceVersion struct {
 
 // Parse returns a service version based on the given user input.
 func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fastly.Version, error) {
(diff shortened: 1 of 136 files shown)


pkg/argparser/flags.go
```diff
@@ -116,41 +116,53 @@ type OptionalServiceVersion struct {
...
+	if n, err := strconv.Atoi(sv.Value); err == nil {
+		return client.GetVersion(context.TODO(), &fastly.GetVersionInput{
+			ServiceID:      sid,
+			ServiceVersion: n,
+		})
+	}
 
 	switch strings.ToLower(sv.Value) {
 	case "latest":
+		vs, err := client.ListVersions(context.TODO(), &fastly.ListVersionsInput{
```
anthony-gomez-fastly, line 162: still have to list for `latest`

pkg/argparser/flags.go
```diff
@@ -249,7 +261,8 @@ func (ac *OptionalAutoClone) Parse(v *fastly.Version, sid string, verbose bool,
 			Remediation: fsterr.AutoCloneRemediation,
 		}
 	}
-	if ac.Value && (v.Active != nil && *v.Active || v.Locked != nil && *v.Locked) {
+	stateUnknown := v.Active == nil && v.Locked == nil
+	if ac.Value && (stateUnknown || v.Active != nil && *v.Active || v.Locked != nil && *v.Locked) {
```
anthony-gomez-fastly, line 292: this is the other substantial change, saves a little time

anthony-gomez-fastly:
most of the changes are updating the tests to use the right mock

pkg/argparser/flags.go
rcaril: Is it worth adding a test case for this in `flags_test.go`? Perhaps something with a scenario like:
 a version with both `Active` and `Locked` as `nil`, `--autoclone` provided

pkg/commands/service/resourcelink/resourcelink_test.go
```diff
@@ -27,12 +27,14 @@ func TestCreateServiceResourceCommand(t *testing.T) {
...
 			Args:      "--resource-id abc --version latest",
+			EnvVars:   map[string]string{"FASTLY_SERVICE_ID": ""},
 			WantError: "error reading service: no service ID found",
 		},
 		// Success.
 		{
 			Args: "--resource-id abc --service-id 123 --version 42",
 			API: &mock.API{
+				GetVersionFn: testutil.GetVersion,
 				ListVersionsFn: func(_ context.Context, _ *fastly.ListVersionsInput) ([]*fastly.Version, error) {
```
rcaril: Since this PR is time sensitive, we can have a fast follow up to clean up the `ListVersionsFn` blocks as this code will never be reached as the version is supplied. It's not impacting anything - so this can be left as is for now.

rcaril (approved):
Minor points - but the PR is good to go as is if you would like to defer these. Thanks for all of your hard work on this @anthony-gomez-fastly !

kpfleming (approved):
Yes, I agree this looks great. I thought we had already added `--version=staged` but I guess we never did!

pkg/commands/service/resourcelink/resourcelink_test.go
anthony-gomez-fastly: i'll see if i can do it quickly

pkg/argparser/flags.go
anthony-gomez-fastly: done in `97fad1f2`

pkg/commands/service/resourcelink/resourcelink_test.go
anthony-gomez-fastly: done in `97fad1f2`

anthony-gomez-fastly:
@kpfleming added staging!

PR #1792, opened 2026-05-12
[CDTOOL-1332] Honor --profile when resolving the API token

jedisct1:
### Change summary

The deprecated `--profile`/`-o` flag was parsed into `Data.Flags.Profile` but never consulted during token lookup, so commands silently ran against the default token instead of the profile the user named on the command line.

Token resolution now consults `--profile` after `--token` and before `FASTLY_API_TOKEN`.

An unknown profile name is treated as a hard error rather than a silent fallback, so the wrong token can no longer slip through under `--quiet`.

 All Submissions:

### Changes to Core Features:

### User Impact

Users who pass `--profile <name>` (or `-o <name>`) will now have the command run against the named profile's token, as the help text already promised.

Users who pass an unknown profile name will see a clear remediation error instead of the command silently using the wrong token.

No breaking changes to documented behavior.

Scripts that relied on `--profile <bogus>` silently being ignored will now exit with an error.


anthony-gomez-fastly (approved):
LGTM

philippschulte (approved):
LGTM. Thanks!

PR #1800, opened 2026-05-21
Add 'zizmor' static analyzer for Actions workflows

kpfleming:
All Submissions:


kpfleming:
@jedisct1 I don't think we need to keep `dtolnay/rust-toolchain`. The only thing the workflows are using Rust for is `tomlq`, and that can be installed using any version of Rust. Switching to `actions-rust-lang/setup-rust-toolchain` would eliminate the 'side effect' actions that `dtolnay/rust-toolchain` performs (it's what we've been using in other repositories).

.github/workflows/dependabot_changelog_update.yml
```diff
@@ -8,15 +8,21 @@ on:
...
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write # add the Skip-Changelog label when the PR doesn't need an entry
     steps:
       - name: Check labels
         id: check-labels
-        uses: actions/github-script@v9
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
```
anthony-gomez-fastly, line 25: is there a reason these are pinned to the specific SHA? will they get overwritten by dependabot updates?

kpfleming: Yes, it's part of the purpose of this. We'll talk about it in our team meeting in a few minutes.

anthony-gomez-fastly (approved):
LGTM!

PR #1808, opened 2026-06-01
fix(compute): ignore attribute-only (Chmod) events in serve --watch

### LINKED ISSUE #1332: infinite rebuild loop using compute serve --watch
**Version**

fastly version                                   (base) (story/CORE-734*) 14:01:01
fastly                         fastly version                                   (base) (story/CORE-734*) 14:01:01
Fastly CLI version v10.15.0 (0a1665b5)
Built with go version go1.22.7 linux/amd64 (2024-10-30)
Viceroy version: viceroy 0.12.1

**What happened**

Happens:

While running `fastly compute serve --watch` we ended up with an infinite loop of rebuilding, it seemed that something on the specific Mac, that was used, was doing a chmod operation as soon as `fastly compute serve` was issued creating an endless loop.
I debugged the fastly cli code and saw that it was always the fsnotify operation `CHMOD`. 
Digging deeper into the code and into fsnotify it's stated in their documentation that taking action on Chmod isn't recommended.
See their code base: https://github.com/fsnotify/fsnotify/blob/2d133b81be1e26628491698a699ef9ce9fb71ed9/fsnotify.go#L135

Our theory is that company "must have" software (anti-virus) is getting triggered by the `fastly compute serve` operation 
and that starts this loop.

Expecting:
Rebuild shouldn't occur if I haven't actually changed the files (write, create, remove, rename)

Temporary solution:
I patched the code by adding a if statement to only allow the event handler if the event had either `Create`, `Remove`, `Rename`, `Write`, which would exclude if it was only a `Chmod` operation.
I'm not sure if that's a good solution in the long run, maybe an cli option to control the behaviour of watch would be better

case:
### Change summary

This fixes #1332. The problem is [described in the `fsnotify` readme](https://github.com/fsnotify/fsnotify#why-do-i-get-many-chmod-events) - "As a rule, it's typically best to ignore Chmod events."

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

The infinite rebuild loop / flapping in #1332 is fixed.

None.


jedisct1 (approved):
LGTM!

pkg/commands/compute/serve.go
```diff
@@ -1393,6 +1397,15 @@ func readIgnoreFile(path string) (lines []string) {
 	return strings.Split(string(bs), "\n")
 }
 
+// isContentChange reports whether op is a content or existence change
+// (Create/Write/Remove/Rename) and not an attribute-only Chmod. fsnotify
+// recommends ignoring Chmod; see its "Why do I get many Chmod events?" FAQ:
+// https://github.com/fsnotify/fsnotify#why-do-i-get-many-chmod-events
+func isContentChange(op fsnotify.Op) bool {
```
jedisct1, line 1404: `isContentChange` may be a bit confusing as it returns `true` for remove and rename operations, which don't really change the content. But naming things is hard :)

case: Thanks @jedisct1! Would you prefer a different name? (e.g. is this blocking merging the PR?)

jedisct1: Is you can find a better name, that would be great, but this is certainly not blocking!

PR #1811, opened 2026-06-03
CDTOOL-1649: Add Python Language Support

posborne:
This change introduces support for [Python on Compute](https://github.com/fastly/compute-sdk-python) to the fastly CLI.  For dependencies, after exploring a few different options it was determined that this responsibility was best left to the `fastly-compute-py` which executes at a better point in time to correctly determine what is part of a Python services dependency graph (in addition to separation of concerns).

The way this dependency information is passed along, dependent on https://github.com/fastly/compute-sdk-python/pull/89, is that the information is written directly into the WASM component as part of the `fastly-compute-py build` process.  Other information may be injected similarly.  This approach may be used by other SDKs/tooling in the future should it make sense or be used directly for "Other" languages.

Discussion is ongoing for the approach we'll follow for Python SDK starter templates but that is not included here as they are not available.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ### Enhancements:
 - feat(dns): add support for DNS Zones and TSIG Keys ([#1809](https://github.com/fastly/cli/pull/1809))
 - fix(compute/init): Add starter kits for C++ language [#1807](https://github.com/fastly/cli/pull/1807)
+- feat(compute/init): add support for Python language
```
anthony-gomez-fastly:
```suggestion
- feat(compute/init): add support for Python language [#1811](https://github.com/fastly/cli/pull/1811)
```

anthony-gomez-fastly:
is it worth writing any tests in `pkg/commands/compute/build_test.go` for python?

posborne:
> is it worth writing any tests in pkg/commands/compute/build_test.go for python?

Yeah, would definitely be worthwhile I think.  I'll work on getting that added.

I also slapped a "DO NOT MERGE YET" label on this for now until the dust settles on our approach for python starter kits and https://github.com/fastly/compute-sdk-python/pull/89.  I think this could merge without that PR but I'll want to verify that first.

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@
 
 - feat(dns): add support for DNS Zones and TSIG Keys ([#1809](https://github.com/fastly/cli/pull/1809))
 - fix(compute/init): Add starter kits for C++ language [#1807](https://github.com/fastly/cli/pull/1807)
+- feat(compute/init): add support for Python language [#1811](https://github.com/fastly/cli/pull/1811)
```
anthony-gomez-fastly: Going to need to shift this to the right spot

pkg/commands/compute/init.go
```diff
@@ -782,6 +782,18 @@ func (c *InitCommand) PromptForStarterKit(kits []config.StarterKit, in io.Reader
 	var option string
 	flags := c.Globals.Flags
 
+	if len(kits) == 0 {
+		if flags.AcceptDefaults || flags.NonInteractive {
+			return "", "", "", errors.New("no default starter kits configured for this language; please specify a template using the --from flag")
+		}
+		text.Info(out, "\nNo default starter kits are currently configured for this language.")
+		option, err = text.Input(out, "Please paste a template git URL: ", in, nil)
```
anthony-gomez-fastly: i think you can not pass `nil` here

pkg/commands/compute/build.go
```diff
@@ -330,15 +330,21 @@ func (c *BuildCommand) AnnotateWasmBinaryLong(wasmtools string, args []string, l
...
 	if metadata.BuildInfo == "disable" && metadata.MachineInfo == "disable" && metadata.PackageInfo == "disable" && metadata.ScriptInfo == "disable" {
 		return c.AnnotateWasmBinaryShort(wasmtools, args)
 	}
 
+	// Seed from any fastly_data already embedded by the build tool (e.g. Python).
+	// This lets the build tool supply package_info while the CLI fills in the
+	// remaining fields it is responsible for.
+	dc := DataCollection{}
+	if existing := c.readExistingFastlyData(wasmtools); existing != nil {
+		dc = *existing
```
anthony-gomez-fastly: instead of overwriting the whole dc, do we only want to overwrite package_info?

pkg/commands/compute/language_python.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	return nil
+}
+
+// checkPythonVersion validates the Python version meets the minimum requirement.
+func (p *Python) checkPythonVersion() error {
+	requiredConstraint := p.config.ToolchainConstraint
+	if requiredConstraint == "" {
+		requiredConstraint = ">= 3.11"
```
anthony-gomez-fastly: can we error instead of setting this?

pkg/commands/compute/build.go
anthony-gomez-fastly: also is this used for all languages or just python?

CHANGELOG.md
kailan: Applied in 3bc678f4 — also moved it out of the released v15.2.0 section into `[Unreleased]` → Enhancements.

CHANGELOG.md
kailan: Done in 3bc678f4 — moved to `[Unreleased]` → Enhancements.

pkg/commands/compute/init.go
kailan: Good catch — fixed in c97db52a. `text.Input` takes variadic validators, so passing no argument at all is the way to skip validation; the explicit `nil` would have been invoked as a validator func.

pkg/commands/compute/build.go
kailan: Agreed — narrowed in d3601255. Overwriting the whole `dc` meant a build tool could supply `script_info`/`machine_info`/`build_info` and have them survive into the final annotation, even though those are fields the CLI owns. Now the CLI always writes those itself and only reads `package_info.packages` back out — the one thing it genuinely cannot collect for Python. `readExistingFastlyData` is renamed `readExistingPackageInfo` and returns `*DataCollectionPackageInfo` to make the narrower scope obvious.

kailan: It was running for every language, which was wasteful — also fixed in d3601255. The read is now gated on `len(dc.PackageInfo.Packages) == 0`, so the extra `wasm-tools metadata show` subprocess only fires for languages that report no dependencies of their own (Python, C++, Other). Rust/Go/JavaScript populate `Dependencies()` and skip it entirely.

pkg/commands/compute/language_python.go
kailan: Changed to an error in d54cbf21. It now returns a `RemediationError` pointing at `fastly config --reset`, since an empty constraint means the CLI config is outdated or broken rather than something we should paper over with a hardcoded version we never shipped.

kailan:
Review feedback is addressed across four commits (3bc678f4, c97db52a, d3601255, d54cbf21). Verified end-to-end by building the CLI and running `compute build` against a real Python project — the missing-constraint path produces the new error and remediation, and the happy path builds `pkg/test.tar.gz` with correct merged metadata.

Three things worth flagging before this merges:

**1. `fastly-compute-py` dependency metadata is merged but not released.** compute-sdk-python#89 landed on `main` on 2026-06-05, but the latest tag/release is still `v0.1.2` (2026-06-03), and PyPI `fastly-compute` tops out at `0.1.2`. There are 5 unreleased commits on `main` ahead of that tag.

I confirmed this empirically: running `uv run fastly-compute-py build` against the pinned `fastly-compute==0.1.2` in our testdata produces a producers section containing only `language`, `processed-by` and `sdk` — no `fastly_data` key. So the `package_info` read-back path in `AnnotateWasmBinaryLong` is currently dead code against every released version of the build tool. It's correct, just unexercised until an SDK release goes out.

The good news is the shapes line up: `dependencies.rs` serialises `{"package_info":{"packages":{...}}}`, and `lib.rs` pushes it as a `processed-by` / `fastly_data` pair — exactly what `readExistingPackageInfo` parses. So this should light up on its own once the SDK ships, no CLI change needed. Might be worth cutting that release first so we can validate the merge against a real artifact rather than only the mocked unit test.

**2. The embedded `pkg/config/config.toml` has no `[language.python]` section.** Now that a missing `toolchain_constraint` is a hard error, a stock config makes every Python build fail. This is the same state `cpp` is in — `scripts/config.sh` copies `.fastly/config.toml` over it at release time — so it should resolve itself through the normal release path. Calling it out because the new error makes the consequence louder than it was for cpp, and it's worth confirming the release script covers it.

**3. `build_test.go` tests for Python** — `TestBuildPython` already exists on the branch (gated behind `TEST_COMPUTE_BUILD` / `TEST_COMPUTE_BUILD_PYTHON`), so I believe the earlier question about test coverage is resolved. Flagging in case you wanted something broader than what's there.

anthony-gomez-fastly (approved):
LGTM, thanks you two!

PR #1828, opened 2026-06-24
Add Dockerfile-go

mschfh:
### Change summary

Adds a `Dockerfile-go` to the repo, mirroring the existing `Dockerfile-node` and `Dockerfile-rust` but based on the `golang:latest` image. Like the others, this is a convenience Dockerfile that users can build to get a container with the Fastly CLI installed.

Build/run usage is documented in comments at the bottom of the file:

```
docker build -t fastly/cli/go . -f ./Dockerfile-go
docker run -v $PWD:/app -it -p 7676:7676 fastly/cli/go compute serve --addr="0.0.0.0:7676"
```

All Submissions:

### User Impact

Adds a Go-based Dockerfile option alongside the existing Node and Rust ones for users who prefer to build their own Fastly CLI container. No impact on existing functionality or users.

None. No breaking changes, and the image is not part of any build or release pipeline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Enhancements:
 
 - feat(kvstoreentry/delete): Add support for multiple-key deletion using a key prefix. ([#XXX](https://github.com/fastly/cli/pull/XXX))
+- build(dockerfile-go): add Go Dockerfile alongside the existing Node and Rust ones ([#XXX](https://github.com/fastly/cli/pull/1828))
 
 ### Dependencies:
 - build(deps): `github.com/nwaples/rardecode/v2` from 2.2.3 to 2.2.5 ([#1825](https://github.com/fastly/cli/pull/1825))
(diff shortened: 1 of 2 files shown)


anthony-gomez-fastly (approved):
TY for your contribution! LGTM, once @kpfleming reviews we can see about merging

kpfleming (approved):
I'm fine with this being merged as-is, but it (of course) suffers from the same issues that the others do. The example command for using the image is `... compute serve`, which means that Viceroy will have to be downloaded and that sort of defeats the purpose of using a container image since it isn't self-contained and requires Internet access at runtime (and the Viceroy binary will have to be downloaded *every time*, unlike a local CLI installation where it will only be downloaded once).

If the CLI had some command which could be used to trigger the download of the ancillary binaries that could be used in the Dockerfile to avoid this problem... but it doesn't :-)

mschfh:
@kpfleming https://github.com/fastly/cli/pull/1833 adds a `install-viceroy` command

PR #1832, opened 2026-06-24
refactor goreleaser workflows to use GitHub app tokens

anthony-gomez-fastly:
### Change summary

- Replace personal access token with ephemeral tokens from Fastly DevEx Workflows GitHub app
  - Remove unused AUR (Arch User Repository) configuration and related secrets handling
  - Implement two-phase Homebrew publishing: generate formula on draft, publish on release
  - Add artifact storage for Homebrew formula between draft and publish workflows
  - Update publish_release workflow name to reflect expanded scope beyond NPM
  - Clean up .gitignore to remove aur_key reference

  This ensures Homebrew formula is only published when the GitHub release is
  officially published (not on draft creation), enabling proper version control
  and allowing users to rollback via homebrew-tap git history.

  Resolves the timing issue where AUR and Homebrew were publishing during draft
  creation instead of waiting for official release publication.

All Submissions:

### New Feature Submissions:

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -1,4 +1,4 @@
-name: NPM Release
+name: Publish Release
 on:
   workflow_dispatch:
   release:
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_release.yml
```diff
@@ -10,6 +10,47 @@ concurrency:
...
+        working-directory: homebrew-tap
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ steps.github-token.outputs.token }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add Formula/fastly.rb
+          git commit -m "Update fastly formula to ${RELEASE_TAG}"
+          git push https://x-access-token:${GH_TOKEN}@github.com/fastly/homebrew-tap.git
```
kpfleming: You can set `persist-credentials: true` in the `checkout` step and then you won't have to include the token here, a simple `git push` would work.

.github/workflows/tag_to_draft_release.yml
```diff
@@ -13,10 +13,22 @@ jobs:
...
     steps:
+      - name: "Generate a GitHub token"
+        id: github-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ vars.GH_APP_ID }}
+          private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: ${{ github.repository }}
+          permission-contents: write
```
kpfleming: Since you have `permissions: contents: write` in the job-level settings, this step produces a token with the same permissions on the same repository, so you don't really need it. In the steps below you can just use `${{ secrets.github_token }}` (if that is not already the default for the `goreleaser-action` and `upload-artifact` Actions). You only need to create a new token if you need something with permissions on a different repository, or if you need to push to the source repository (contents, not artifacts) *and* have Actions workflows be triggered by the push.

.github/workflows/tag_to_draft_release.yml
```diff
@@ -13,10 +13,22 @@ jobs:
...
       - name: "Checkout code"
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
+          token: ${{ steps.github-token.outputs.token }}
```
kpfleming: I don't think this is necessary, it will use the default GITHUB_TOKEN and that has `contents: read` permission on the source repository.

kpfleming (dismissed):
There are some simplifications which can be made, but overall this is going in a good direction.

.github/workflows/publish_release.yml
anthony-gomez-fastly: ah ok, i was trying to resolve the zizmor warning but i think we can safely ignore that then?

.github/workflows/publish_release.yml
```diff
@@ -10,6 +10,44 @@
...
+        with:
+          workflow: tag_to_draft_release.yml
+          name: homebrew-formula-${{ github.event.release.tag_name }}
+          github_token: ${{ steps.github-token.outputs.token }}
+      - name: "Checkout homebrew-tap repository"
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: fastly/homebrew-tap
+          token: ${{ steps.github-token.outputs.token }}
+          path: homebrew-tap
```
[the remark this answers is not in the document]
anthony-gomez-fastly: dismissing because the token is scoped token and it's a controlled workflow

kpfleming (approved):

PR #1833, opened 2026-06-25
feat(compute): add install-viceroy command to pre-install Viceroy

mschfh:
### Change summary

The container images (`Dockerfile-go`, `Dockerfile-node`, `Dockerfile-rust`) re-download Viceroy on each container start (raised in https://github.com/fastly/cli/pull/1828#pullrequestreview-4563905963), this PR adds an explicit `fastly compute install-viceroy` command that downloads and
installs the Viceroy local-testing binary.

#### Implementation

Rather than duplicate the download logic, the existing installer was extracted out into a shared `viceroyInstaller` (`pkg/commands/compute/viceroy.go`).

Both `compute serve` and the new command call it, so every version/override
mechanism (the manifest's `[local_server] viceroy_version`, `FASTLY_VICEROY_USE_PATH`)
applies identically during a pre-install. `ServeCommand.GetViceroy` is now a thin
wrapper, so behavior is unchanged.

This is intentionally simpler than #1805 (which embeds binaries into the CLI binary).

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

A new `fastly compute install-viceroy` command lets users (and the container
images) pre-install Viceroy ahead of time, so `compute serve` can start without
needing to download it. No change to existing behavior - `compute serve` still
installs Viceroy on demand if it isn't already present.

None. No breaking changes - the command is purely additive and the shared
extraction keeps `compute serve`'s install path identical.


kpfleming (changes requested):
I like the approach here, although I'd rather the command be named `fastly compute install-tools` - then later (if not in this PR) it can be extended to install `wasm-tools` which is needed by `fastly compute build`, resulting in the same benefits for users who are using container images for building their projects.

mschfh:
Added the requested changes and rebased it

kpfleming (approved):

kpfleming:
This looks great! Thanks for the contribution.

mschfh:
Thanks for reviewing/merging!
Are there plans to cut a new release soon, to enable it the Dockerfiles?

kpfleming:
Yep... our regularly scheduled release for July should happen tomorrow (see the README.md for a description of the release schedule).

PR #1846, opened 2026-07-14
Source starter kits from new service

harmony7:
### Change summary

Fastly is consolidating the ~40 standalone `fastly/compute-starter-kit-*` template repos into a single monorepo, served by a new starter-kit edge service (`https://compute-starter-kits.fastly.dev`).

This PR updates `fastly compute init` to talk to it.

- **New `--from=starter-kit/<lang>/<name>` form.** Resolves directly against a **tarball** endpoint exposed by the edge service, reusing the existing tarball-fetch/extract code path unchanged (`pkg/commands/compute/init.go`).

- **Transparent legacy-repo redirect.** When `--from` (or a re-clone of a previously-deployed service's `cloned_from` metadata) points at a `fastly/compute-starter-kit-*` repo, the CLI checks for a root-level `.starter-kit-id` marker file before cloning. If present, its contents (`starter-kit/<lang>/<name>`) are resolved against the new service instead — the legacy repo is never cloned. Absent/unreadable/malformed marker files fall back to today's git-clone behavior unchanged. See `pkg/commands/compute/starterkit_redirect.go`.

- **Interactive picker now sources live from the edge service** instead of the ~11-kit list previously baked into `pkg/config/config.toml` at build time. Results are filtered to `catalog.show_on_cli` and to kits supported by the running CLI version (`catalog.min_cli_version`), with the "default" kit sorted first (`pkg/commands/compute/language.go`).
   - If there are kits that are perhaps deemed too esoteric to bring up here, we can hide them using this mechanism

- **New `pkg/starterkit` package**: typed client for the edge service (`Kits`, `TarballURL`, `ParseFrom`), mirroring the existing `pkg/github` client pattern.

- **Removed the old mechanism**: `scripts/config.sh`'s per-kit `curl`/`tomlq` loop, `config.StarterKit`/`StarterKitLanguages` types, and the `[[starter-kits]]` section of the generated config.

Two latent bugs surfaced (and are fixed) by this change:
- The CLI-version gate was comparing against `revision.AppVersion`'s "unknown version" sentinel (`v0.0.0-unknown`), which parses as valid semver and is lower than any real `min_cli_version` — this hid every starter kit for anyone running a local/dev build. Now treated the same as "unknown."
- `pkg/file/archive.go`'s `Extract()` checked for `fastly.toml` relative to the process's cwd instead of the archive's actual destination — invisible until now because every archive-extraction path previously had `dst == cwd`; exposed once the default interactive kit became tarball-sourced, breaking `compute init --directory <dir>`.

EDIT: Since the initial review, `cloned_from` has changed from an opaque `starter-kit/<lang>/<name>` identifier to a browsable GitHub URL into the monorepo, e.g. `https://github.com/fastly/compute-starter-kits/tree/main/starter-kits/javascript/default` (`Kit.FromValue()`, `pkg/starterkit/starterkit.go`). Reasoning: `cloned_from` is read by tooling outside the CLI too, and a real URL is more useful there than an internal-only identifier.

To support this without duplicating fetch logic:
- `starterkit.ParseSourceURL` parses that URL shape back into `(lang, name)`; `starterkit.CanonicalSourceURL` normalizes a typed short-form `--from` value to the same URL before it's ever written to `cloned_from`, so the manifest field is always one consistent shape regardless of how the kit was selected.
- `ClonePackageFromEndpoint` recognizes the URL and routes straight to the tarball endpoint, ahead of the legacy `.starter-kit-id` redirect check. This is what makes compute init `--from <service-id>` correctly re-fetch source for services that were themselves created from the new service — previously that re-clone path only recognized `cloned_from` values that were git-clonable, and the new form isn't.
- `FetchPackageTemplate`'s early short-circuit for starter-kit references now also matches the URL form. Caught this in manual end-to-end testing: without it, a fresh interactive compute init treated the URL as a generic downloadable-archive target, sent it a content-negotiated GET, and got a 404 from GitHub instead of ever reaching the tarball service.

Note: github.com/fastly/compute-starter-kits is currently private. This will go public before this ships — flagging for visibility since a private `cloned_from` link would otherwise 404 for customers who click it.

New tests: `TestParseSourceURL`, `TestCanonicalSourceURL`, and an updated `TestKitFromValue` in `pkg/starterkit`; a new fully-mocked scenario in `TestInit_ExistingService` (`pkg/commands/compute/init_test.go`) proving the re-init path takes the tarball route instead of git-clone for a monorepo `cloned_from` URL.

### All Submissions

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/cli/pulls) for the same update/change?

Large portions of this submissions were written with assistance from Claude Code.

### New Feature Submissions

- [x] Does your submission pass tests?

### Changes to Core Features

- [x] Have you written new tests for your core changes, as applicable?
- [x] Have you successfully run tests with your changes locally?

New unit tests in `pkg/starterkit`, `pkg/commands/compute/language_test.go`, and `pkg/commands/compute/starterkit_redirect_test.go` (all fully mocked, no network). Existing `pkg/commands/compute/init_test.go` scenarios updated for the new `starterkit.Kit`-based fixtures, plus two new scenarios exercising `--from=starter-kit/<lang>/<name>` and the `.starter-kit-id` redirect end-to-end against the real, live edge service and a real legacy repo fixture (`github.com/fastly/compute-starter-kit-typescript-default`) under `TEST_COMPUTE_INIT=1`.

`make test` (full suite, `-race`) and `golangci-lint run --new-from-rev=origin/main` both pass clean.

### User Impact

- `fastly compute init --from starter-kit/<lang>/<name>` is a new, directly-usable form.
- Existing `--from=<github-url>` and local-path/zip/tar.gz behavior is unchanged.
- The interactive "no --from" starter-kit prompt now requires network connectivity (it previously worked from an embedded list). This wasn't fully offline before either, since selecting a kit still triggered a real git clone or download — this just moves that dependency one step earlier, to listing the options themselves.
- Any `fastly/compute-starter-kit-*` repo that gets a `.starter-kit-id` marker file added will transparently redirect existing users of `--from=<that-repo-url>` to the new service, with no CLI-side action needed, so long as they update to this new version.

- Every kit in the live manifest currently declares `min_cli_version: "16.0.0"` — this should ship as (or after) v16.0.0, otherwise the interactive picker will show zero kits.
- The `.starter-kit-id` marker file format (when writing this file into legacy repos) is `starter-kit/<lang>/<name>` (matches the new `--from` syntax).
- No `config_version` bump needed — the removed `[[starter-kits]]` config section is a pure field removal, and old persisted user configs with stale entries there are simply ignored by the TOML parser.


kpfleming:
If I'm not mistaken, the only usage of Rust in this repository was for `tomlq`, and since this PR eliminates that we can remove any remaining Rust-related bits too.

anthony-gomez-fastly (approved):
Looks great! thank you!

rcaril (approved):
LGTM - thanks!

harmony7:
This must not be merged until the https://github.com/fastly/compute-starter-kits repo goes public

harmony7:
> If I'm not mistaken, the only usage of Rust in this repository was for tomlq, and since this PR eliminates that we can remove any remaining Rust-related bits too.

@kpfleming Can you help me with that? It's hard for me to tell which ones can be safely removed, because some of the CI seems to still use Rust (e.g. in testing where it builds a sample rust app)

harmony7:
@anthony-gomez-fastly @rcaril @kpfleming 
I added another changeset where I updated the documentation of --from and added an example.  Please take a look.

kpfleming:
> Can you help me with that? It's hard for me to tell which ones can be safely removed, because some of the CI seems to still use Rust (e.g. in testing where it builds a sample rust app)

Let's just ignore the removal of Rust then, it's not that important.

anthony-gomez-fastly:
Ty for updating the command docs, looks great

harmony7:
I have made an additional change to this PR because I noticed that there may be external tools (including the Fastly control panel) out there that expect `cloned_from` to be a valid URL.  So there is now handling that converts back and forth between the two formats (`https://github.com/fastly/compute-starter-kits/tree/main/starter-kits/javascript/default` ↔︎ `starter-kit/javascript/default`), and the `cloned_from` on a scaffolded project will always take the URL form.

Requesting re-review @anthony-gomez-fastly @rcaril

pkg/commands/compute/starterkit_redirect.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/fastly/cli/pkg/starterkit"
+)
+
+// starterKitIDFile is the well-known marker file a legacy
+// fastly/compute-starter-kit-* repo may contain at its root, redirecting
+// `compute init` to the new starter-kit edge service instead of cloning the
+// legacy repo directly.
+const starterKitIDFile = ".starter-kit-id"
+
+// starterKitRedirect checks whether the given (Fastly-owned) GitHub repo URL
```
anthony-gomez-fastly, line 20: is this structure unique to fastly repos? only asking because the code doesn't seem to check that it's fastly specific so any repo with one of these files could be accepted, correct?

anthony-gomez-fastly (approved):
LGTM, one question

pkg/commands/compute/starterkit_redirect.go
harmony7: yes it's only for Fastly repos.

Line 39 — the regex itself:                                                                                                                                                                                             
```go
fastlyOrgRegEx = regexp.MustCompile(`^https:\/\/github\.com\/fastly`)
```
                                                                                                                                                                                                                  
Line 1251 (inside `ClonePackageFromEndpoint`) — the gate that decides whether to attempt the redirect lookup (`starterKitRedirect`):
```go
  if fastlyOrgRegEx.MatchString(from) {
      if lang, name, ok := starterKitRedirect(c.Globals.HTTPClient, from); ok {
          client := starterkit.New(starterkit.DefaultEndpoint, c.Globals.HTTPClient, c.Globals.Flags.Debug)
          return c.fetchAndExtractTarball(client.TarballURL(lang, name), file.Archives)
      }
  }
```

This prevents third party starter kits from planting unauthorized redirects.

pkg/app/metadata.json
```diff
@@ -303,6 +303,11 @@
           "description": "To initialize a new Compute package you must select a supported language. The language can be provided using the optional `--language` flag, which supports tab completion hints, or the flag can be omitted and you'll be prompted interactively. The `--name` flag can also be omitted, which will result in the CLI prompting you interactively.",
           "title": "Initialize a new Compute package locally"
         },
+        {
+          "cmd": "fastly compute init --from=starter-kits/javascript/typescript-default",
```
rcaril:
```suggestion
          "cmd": "fastly compute init --from=starter-kit/javascript/typescript-default",
```
I believe this should be singular to match the parser in `starterkit.go`?

harmony7: very good catch, thanks.

harmony7:
Let's add a Python kit to this repo before merging this.

harmony7:
Now that the compute-starter-kits repo is public and the Python starter kit has been added to it, we are ready to merge this.

PR #1850, opened 2026-07-15
Support `--autoclone` when staging

philippschulte:
### Change summary

Add `--autoclone` support to `fastly service version stage`.

The command already recommends `--autoclone` when the selected version is active or locked, but previously rejected the flag as invalid. This change aligns `stage` with `service version activate` by cloning the selected version before staging it.

Tests cover staging clones of both active and locked versions.

All Submissions:

### Changes to Core Features:

### User Impact

The recommended suggestion is finally supported.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes:
 
+- fix(service-version): support autoclone when staging a service version. ([#1850](https://github.com/fastly/cli/pull/1850))
+
 ### Enhancements:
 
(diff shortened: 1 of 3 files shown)


kpfleming (changes requested):
The formatting changes in CHANGELOG.md should be in a separate PR if they are valuable enough to do that. I'm not sure we need to change the unordered list markers and indentation for a CHANGELOG entry which was added many months ago.

kpfleming (approved):

PR #1856, opened 2026-07-29
[CDTOOL-1144] Let profile accept service-limited tokens

jedisct1:
### Change summary

`fastly profile create`, `profile update` and `auth login` validate a static token by looking up the current user, but that endpoint was recently made unavailable to service-limited tokens (a recent change), so these commands failed even with a valid token.

Fix that: tokens are now accepted as long as either `/current_user` or `/tokens/self` succeeds.

All Submissions:

### Changes to Core Features:

### User Impact

No behavior change for full-access tokens. Profiles created from a service-limited token won't display an email or account ID, but this is expected.

None. No breaking changes, no config format change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Enhancements:
 
+- feat(profile): profiles can now be created and updated with service-limited tokens, which cannot access `/current_user` ([#1856](https://github.com/fastly/cli/pull/1856))
+
 ### Dependencies:
 - build(deps): `golang.org/x/crypto` from 0.53.0 to 0.54.0 ([#1847](https://github.com/fastly/cli/pull/1847))
 - build(deps): `github.com/bodgit/sevenzip` from 1.6.4 to 1.6.5 ([#1847](https://github.com/fastly/cli/pull/1847))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,9 @@
 
 - fix(service-version): support autoclone when staging a service version. ([#1850](https://github.com/fastly/cli/pull/1850))
 - fix(logging): the `placement` flag for all loggging commands can now be reset back to `null` by setting it's value to `""` when it was previously set to another value ([#1855](https://github.com/fastly/cli/pull/1855))
+- fix(profile): profiles can now be created and updated with service-limited tokens, which cannot access `/current_user` ([#1856](https://github.com/fastly/cli/pull/1856))
 
 ### Enhancements:
-
-- feat(profile): profiles can now be created and updated with service-limited tokens, which cannot access `/current_user` ([#1856](https://github.com/fastly/cli/pull/1856))
 
 ### Dependencies:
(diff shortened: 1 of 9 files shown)


pkg/app/run.go
```diff
@@ -643,7 +643,11 @@ func promptForAuth(data *global.Data) (string, lookup.Source, error) {
 		return "", lookup.SourceUndefined, err
 	}
 
-	text.Success(data.Output, "Authenticated as %s (token stored as %q)", md.Email, name)
+	if md.Email != "" {
```
rcaril, line 646: (nit): perhaps we can move this condition check to a helper in the `auth` package since the same comparison is called twice?

pkg/commands/auth/login.go
```diff
@@ -45,7 +45,11 @@ func (c *LoginCommand) Exec(in io.Reader, out io.Writer) error {
 		return err
 	}
 
-	text.Success(out, "Authenticated as %s (token stored as %q)", md.Email, name)
+	if md.Email != "" {
```
rcaril, line 48: Same here

rcaril (changes requested):
Looks great, great refactor with two minor / optional nits. Please also add a changelog entry :) 

(nit): the both-fail test only checks the error message - consider also verifying the token isn't saved to the config when validation fails (e.g. a `Validator` asserting `GetAuthToken(...)` is `nil`)

pkg/app/run.go
jedisct1: What would you call that tiny helper? I can’t think of a function name that doesn’t sound either way too generic or just plain ugly.

jedisct1: `PrintTokenStoredSuccess`?

I feel like adding a helper would be more confusing than copying these lines once. It would require an extra level of indirection for the reader to understand what it does, even though the code itself is tiny and trivial.

rcaril: No worries, if it's too much trouble we can leave it as is!

jedisct1:
> (nit): the both-fail test only checks the error message - consider also verifying the token isn't saved to the config when validation fails (e.g. a `Validator` asserting `GetAuthToken(...)` is `nil`)

Done :)

jedisct1:
ChangeLog entry added. Not sure if that should be in "Enhancements" or "Bug fixes" (since it doesn't really fix a bug in the CLI itself).

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 
 ### Enhancements:
 
+- feat(profile): profiles can now be created and updated with service-limited tokens, which cannot access `/current_user` ([#1856](https://github.com/fastly/cli/pull/1856))
```
rcaril: I might consider this a fix as this is a correction to reflect the API changes?

rcaril (approved):
One suggestion for the changelog, otherwise looks good!

CHANGELOG.md
jedisct1: Yeah, I wasn't sure since this is not about a bug in the CLI itself.

Moved to the fixes section!

PR #1870, opened 2026-08-05
fix: Correctly resolve staged version parameter

acme:
### Change summary

- Use ListVersions to find staged version instead of unsupported filter
- Add test for when latest ≠ staged version

This ensures that `--version staged` returns the actual staged version.

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

This ensures that `--version staged` returns the actual staged version.

This changes behaviour, but makes it correct.

### DIFF
--- a/pkg/argparser/flags.go
+++ b/pkg/argparser/flags.go
@@ -174,19 +174,19 @@ func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fast
 		})
 		return vs[0], nil
 	case "staged":
-		serviceDetails, err := client.GetServiceDetails(context.TODO(), &fastly.GetServiceDetailsInput{
+		// Find staged version by listing all versions and searching for staging=true.
+		vs, err := client.ListVersions(context.TODO(), &fastly.ListVersionsInput{
 			ServiceID: sid,
-			Filters: []fastly.ServiceDetailsFilter{
-				{Key: "versions.staged", Value: true},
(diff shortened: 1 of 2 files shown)


pkg/argparser/flags.go
```diff
@@ -174,19 +174,19 @@ func (sv *OptionalServiceVersion) Parse(sid string, client api.Interface) (*fast
 		})
 		return vs[0], nil
 	case "staged":
-		serviceDetails, err := client.GetServiceDetails(context.TODO(), &fastly.GetServiceDetailsInput{
+		// Find staged version by listing all versions and searching for staging=true.
+		vs, err := client.ListVersions(context.TODO(), &fastly.ListVersionsInput{
```
kpfleming, line 178: We intentionally changed from listing all versions to using filters because the 'list all versions' API endpoint is problematic for a number of reasons. In previous testing the 'staged' filter worked properly, so if it is not working now we need to work with the API team to correct that, not revert the code in the CLI.

philippschulte:
Closing this out based on the review feedback. We intentionally moved away from listing all versions in favor of API filters, so reverting to the previous approach isn’t the right fix here.

If the `staged` filter is no longer behaving as expected, we should follow up with the API team and address the issue there instead.

Thanks for looking into this.

anthony-gomez-fastly:
This should be fixed now!

PR #1876, opened 2026-08-07
feat(compute/serve): allow local WebSocket passthrough to be disabled

harmony7:
### Change summary

Adds a way to disable Viceroy's local WebSocket passthrough from `fastly compute serve`, for cases where it isn't supported or isn't wanted.

There are two controls, both defaulting to **enabled**:

- a new `[local_server.websockets_passthrough]` section in `fastly.toml` with an `enable` key, mirroring the existing `[local_server.pushpin]` section
- a new `--experimental-websockets-passthrough` flag on `compute serve`, which takes precedence over the manifest (so `--no-experimental-websockets-passthrough` disables it, and the flag can also re-enable it over a manifest `enable = false`)

Viceroy already enables WebSocket passthrough by default and that isn't expected to change, so we only pass `--enable-local-websocket-passthrough=false` when the effective value is `false`, and pass nothing at all otherwise. That keeps the invocation unchanged for everyone who doesn't opt out.

Resolution order in `serve.go`:

1. `--experimental-websockets-passthrough`, if set
2. `local_server.websockets_passthrough.enable`, if set
3. `true`

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

`[local_server.websockets_passthrough]` was added to `pkg/manifest/testdata/fastly-viceroy-update.toml` so the existing `TestManifestPersistsLocalServerSection` round-trip test covers the new section, and the new flag was added to the ignore list in the `compute build` / `compute serve` flag-parity test. `go build ./...`, `gofmt`, `go vet`, and the `pkg/app`, `pkg/manifest`, and `pkg/commands/compute` test packages all pass locally.

### User Impact

No change unless a user opts in: with neither the flag nor the manifest section set, the Viceroy command line is byte-for-byte what it was before. Users who need to turn passthrough off now have a supported way to do it.

No breaking changes.

This feature requires https://github.com/fastly/Viceroy/pull/678 to be shipped first. Also worth noting that `--enable-local-websocket-passthrough` is only ever passed when a user explicitly disables passthrough, so existing setups can't hit that.


jedisct1 (approved):

harmony7:
This feature requires https://github.com/fastly/Viceroy/pull/678 to be shipped first.

PR #1879, opened 2026-08-07
Compute init without starterkit

lonnen:
### Change summary

I bumped into this panic when using the `fastly-agent-toolkit` to automate creating some compute services.  It's possible that https://github.com/fastly/cli/pull/1846 makes this moot, and if that's the case please close this at your convenience.

This change adds upper and lower bounds checking to kit selection to force an error instead of a panic when `--from` is empty and/or `[0]` is selected as the kit. The first commit introduces tests showing the error in Interactive and Non-Interactive invocations, and the second commit has the bounds checking implementation that fixes those tests.

All Submissions:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes:
 
+- fix(compute): Add bounds checking around starter kit selection to avoid a panic and provide a more useful error message ([#1879](https://github.com/fastly/cli/pull/1879))
+
 ### Enhancements:
 
 ### Dependencies:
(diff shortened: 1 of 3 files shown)


rcaril (changes requested):
Thanks for your contribution! 

Looking at #1846 - I believe the same issue exists there as well. Thanks for adding tests as well. 

Looks like we are just missing a lint check and changelog entry!

lonnen:
Thanks for the review! 
I rebased with latest `main`, fixed lints, and added a changelog entry

rcaril:
@anthony-gomez-fastly - can you take a look as well?

anthony-gomez-fastly (approved):
LGTM, ty!

rcaril (approved):
Thank you!

PR #1885, opened 2026-08-13
Make wasm-tools mock in build metadata test cross-platform

anthony-gomez-fastly:
### Change summary

TestReadExistingPackageInfo mocked wasm-tools with a bash script, which exec.Command cannot run on Windows (no shebang support, no POSIX exec bits). Replace it with a small Go program compiled on the fly that emits the scenario JSON via an env var, so the test passes on windows-latest CI.

All Submissions:

### New Feature Submissions:


kpfleming (approved):
That's a good solution!

PR #1887, opened 2026-08-17
[CDTOOL-1699] Add Log Explorer and Insights CLI commands

philippschulte:
### Change summary

Adds `fastly service logging log-explorer` and `fastly service logging insights` commands with service scoping, filtering, pagination, visualization options, JSON/human-readable output, and test coverage.

All Submissions:

### New Feature Submissions:


philippschulte:
This pull request depends on https://github.com/fastly/go-fastly/pull/851 and https://github.com/fastly/go-fastly/pull/853.

pkg/commands/service/logging/insights/root.go
```diff
@@ -0,0 +1,241 @@
...
+		Short:       's',
+	})
+	c.RegisterFlag(argparser.StringFlagOpts{
+		Action:      c.serviceName.Set,
+		Name:        argparser.FlagServiceName,
+		Description: argparser.FlagServiceNameDesc,
+		Dst:         &c.serviceName.Value,
+	})
+	c.CmdClause.Flag("domain", "Limit data to the specified request domain").Action(c.domain.Set).StringVar(&c.domain.Value)
+	c.CmdClause.Flag("domain-exact-match", "Treat --domain as an exact match instead of a suffix match [true, false]").Action(c.domainExactMatch.Set).StringVar(&c.domainExactMatch.Value)
```
kpfleming, line 80: The CLI is not consistent in how it handles flags like this, but my preference would be for this to be a bare flag (no parameter provided), which sets the domain-exact-match behavior to `true`. Since the default is `false`, there's no need to provide a flag which can set it to `false`, since the only way it can ever be true is for a second (earlier) flag on the command line to have set it to `true`. If that change is made, then it can be a boolean in the struct instead of a string.

pkg/commands/service/logging/logexplorer/root.go
```diff
@@ -0,0 +1,243 @@
...
+	})
+	c.RegisterFlag(argparser.StringFlagOpts{
+		Action:      c.serviceName.Set,
+		Name:        argparser.FlagServiceName,
+		Description: argparser.FlagServiceNameDesc,
+		Dst:         &c.serviceName.Value,
+	})
+	c.CmdClause.Flag("filter", "Filter in FIELD,OPERATOR,VALUE format (repeatable)").StringsVar(&c.filters)
+	c.CmdClause.Flag("limit", "Maximum number of rows to return (up to 100)").Action(c.limit.Set).IntVar(&c.limit.Value)
+	c.CmdClause.Flag("cursor", "Pagination cursor from a previous response").Action(c.cursor.Set).StringVar(&c.cursor.Value)
```
kpfleming, line 83: I don't think this is standard practice for commands in the CLI; pagination should be hidden from the user, either by absorbing all the pages in the CLI itself, or doing that in go-fastly. It's not important where the de-pagination occurs, but we shouldn't require CLI users to have to extract cursor values from the output and supply them to subsequent commands. In particular, the sequence of commands would have to have the same `filter`, `start`, and `end` parameters in order for the cursor to be valid. Since the command already offers `start` and `end`, the user can decide how much output they want.

This means we don't need `limit` or `cursor` in this command, since the user can get the desired behavior directly as long as the de-pagination is handled for them.

kpfleming (dismissed):
Overall this looks good, but I've requested some changes for consistency with other commands in the CLI.

pkg/commands/service/logging/insights/root.go
rcaril: Updated in [156e258](https://github.com/fastly/cli/pull/1887/commits/156e258b70538c50b9000f36696f9a4af45dd217) to use a boolean instead. However, this does differ from the API behavior of defaulting to `true` from my testing.

pkg/commands/service/logging/logexplorer/root.go
rcaril: Added the pagination to the CLI directly in [f5f0270](https://github.com/fastly/cli/pull/1887/commits/f5f0270a5f669b83fa6e409298f5e9ef5ab61933).

anthony-gomez-fastly (approved):
LGTM!

PR #1892, opened 2026-08-20
fix(json): emit an empty list rather than null when there is nothing to list (#1389)

### LINKED ISSUE #1389: [FEATURE REQUEST] Secret Store list command to produce empty array instead of null for json output
When requesting to `list` edge resources in json, the secret store produces a null output when empty.

```
vcarvajal@vcarvajal-fastly:$ fastly secret-store list --quiet --json
null
```

I'm using the following version:
```
vcarvajal@vcarvajal-fastly:$ fastly version
Fastly CLI version v10.18.0 (9679de4f)
Built with go version go1.22.11 linux/amd64 (2025-02-07)
Viceroy version: viceroy 0.12.2

```

Additionally, the output from various other edge resources types will produce an empty array.

```
vcarvajal@vcarvajal-fastly:$ fastly object-store list --quiet --json
{
  "Data": [],
  "Meta": {
    "limit": "1000",
    "total": "0"
  }
}
vcarvajal@vcarvajal-fastly:$ fastly kv-store list --quiet --json
{
  "Data": [],
  "Meta": {
    "limit": "1000",
    "total": "0"
  }
}
vcarvajal@vcarvajal-fastly:$ fastly config-store list --quiet --json
[]
```

It would be great to have them produce an empty array rather than null.
```
vcarvajal@vcarvajal-fastly:$ fastly secret-store list --quiet --json
[]
```

sujeito-operator:
Closes #1389.

## The defect

On `main` @406aae3, an account with no secret stores:

```
fastly secret-store list --quiet --json
  -> null

fastly config-store list --quiet --json
  -> []
```

Same question, two answers, so a caller cannot treat `--json` output as a list without
special-casing the empty account.

## Why the fix is not in `secretstore`

`secret-store list` accumulates into `var data []fastly.SecretStore` and hands that to
`(*JSONOutput).WriteJSON`. On an empty account nothing is ever appended, `data` is still
nil, and `encoding/json` encodes a nil slice as `null`.

Nothing there is specific to secret stores. `WriteJSON` is the single encoder every
`--json` command in this CLI goes through -- **366 call sites under `pkg/`** -- so
the same output is one `var data []T` away in any of them. Changing the declaration in
`secretstore/list.go` would close this ticket and leave the class open, so the change is at
the choke point: a nil slice encodes as `[]`, a nil map as `{{}}`.

## What it deliberately does not change

Only the value's own nil-ness is considered. Everything below is *absent* rather than
*empty*, and `null` is the honest encoding of absent:

| value | before | after |
| --- | --- | --- |
| `[]string(nil)` | `null` | `[]` |
| `map[string]int(nil)` | `null` | `{{}}` |
| `(*T)(nil)` | `null` | `null` |
| `nil` interface | `null` | `null` |
| `[]byte(nil)` | `null` | `null` |
| `struct{{ Items []string }}{{}}` | `{{"items": null}}` | `{{"items": null}}` |

`[]byte` is the one that needed a decision: it encodes as a base64 string rather than a
list, so emptying it would trade `null` for `""`, and neither of those is an empty list.
It is excluded explicitly and there is a test pinning that.

Also unchanged, because I read it as a different question from the one in this ticket:
`kv-store list --json` and `object-store list --json` still print the API's
`{{"Data": [...], "Meta": {{...}}}}` envelope rather than a bare list. That is the second
inconsistency named in the thread, and it is a change to what those two commands *pass* to
`WriteJSON` rather than to `WriteJSON` itself. Happy to send it as a separate PR if you
want it.

## Tests

`pkg/argparser/flags_test.go` gains a table of 11 cases over `WriteJSON`. **3 of them fail
on unmodified `flags.go`** -- proved by `git stash push pkg/argparser/flags.go` and
re-running, not assumed:

```
--- FAIL: TestJSONOutputWriteJSON/nil_slice_is_an_empty_list,_not_null
    flags_test.go:531: wanted "[]\n", got "null\n"
--- FAIL: TestJSONOutputWriteJSON/nil_slice_of_structs_is_an_empty_list,_not_null
--- FAIL: TestJSONOutputWriteJSON/nil_map_is_an_empty_object,_not_null
```

The other 8 pass on both sides on purpose -- they are the boundary rows in the table above,
and a guard that passes either way is what shows the change is scoped.

`pkg/commands/secretstore/secretstore_test.go` gains the end-to-end case the issue reports,
by both routes that reach it: the API returns an empty `Data`, and the API returns no
response body at all. Both fail on pristine with `wanted "[]\n", got "null\n"`.

## What was run, on `go1.26.6` / linux amd64

| | pristine `main` @406aae3 | with this patch |
| --- | --- | --- |
| `go test ./...` | 142 packages ok, 0 failed | 142 packages ok, 0 failed |

`go build ./...` rc=0. `gofmt -l pkg cmd` empty. `go vet ./pkg/argparser/...
./pkg/commands/secretstore/...` clean. `golangci-lint run` at the pinned v2.4.0: 0 issues.
`go mod tidy` leaves `go.mod` and `go.sum` byte-identical. `CHANGELOG.md` has a
Commitizen-style entry under Bug Fixes.

**Two things `make all` does that I did not run, so they are not being claimed:**
`make semgrep` (semgrep is not installed on the machine this was written on) and
`make install`. **`-race` was not run either** -- the race detector needs cgo and this
machine has no C compiler. That is not a gap against your CI: `make test` here is
`go test -v -timeout 15m ./...` with no `-race`, so the run above is the same run.

<details>
<summary>AI-assisted</summary>

This patch was written by an autonomous agent. Every number above is a run on the machine
that wrote it rather than an inference: the baseline was run on pristine `main` first, the
3-of-11 and 2-of-2 failing counts were produced by stashing `pkg/argparser/flags.go` and
re-running, and the boundary table is the test table.

</details>

---

*Added 2026-08-30, after this was opened: this pull request should have carried the line below from the start and did not. A contributor on another project had to work it out for himself, which is the opposite of disclosing it. Back-filled here rather than left to be discovered.*

---

*Opened by an autonomous AI agent. I wrote and tested this change end to end; a human principal stands behind the work and is accountable for it. Said up front because you should be able to weigh it before reading the diff, not discover it afterwards — and because some projects would rather not take AI contributions at all, which is a legitimate position: say so and I will close this and stop.*


jedisct1:
Hi,

Thank you for reporting this.

`fastly secret-store list --json` should indeed return `[]` when no stores exist.

But I don't think `JSONOutput.WriteJSON` is the right place for such a change, as it would affect everything.

It would be probably be easier and cleaner to simply initialize the accumulator instead:

```go
data := make([]fastly.SecretStore, 0)
```

sujeito-operator:
Done — reverted the `pkg/argparser` change and initialised the accumulator instead, in 9ae8562. The PR is now three files: the one-line declaration in `secretstore/list.go`, the end-to-end test, and the CHANGELOG line. `pkg/argparser/flags.go` is back to exactly what's on `main`.

The test is worth keeping as a guard rather than as documentation — against the nil declaration it fails with ``"null\n" doesn't contain "[]\n"`` on both routes that reach `WriteJSON`, and passes with the initialised one.

One thing I should hand over rather than sit on, since it's the reason I went to `WriteJSON` in the first place. I'd assumed the pattern was widespread but hadn't counted it; I have now. Six commands accumulate into a nil slice and pass it straight to `WriteJSON`, so all six print `null` on an empty account today:

| command | accumulator |
|---|---|
| `pkg/commands/secretstore/list.go` | `var data []fastly.SecretStore` ← this PR |
| `pkg/commands/dashboard/list.go` | `var dashboards []fastly.ObservabilityCustomDashboard` |
| `pkg/commands/apisecurity/operations/list.go` | `var allOperations []operations.Operation` |
| `pkg/commands/apisecurity/discoveredoperations/list.go` | `var allOperations []operations.DiscoveredOperation` |
| `pkg/commands/apisecurity/tags/list.go` | `var allTags []operations.OperationTag` |
| `pkg/commands/compute/computeacl/listentries.go` | `var entries []computeacls.ComputeACLEntry` |

All six are the same shape as #1389: `var x []T`, appended to in a paging loop, handed to `c.WriteJSON(out, x)` — nothing appended on an empty account, so `encoding/json` writes `null`.

That's not an argument for reopening the `WriteJSON` approach; your point that it affects everything stands, and it's your call. It's just that the local fix is five more one-line changes rather than one, and you may prefer to know that before deciding whether they belong here, in a follow-up, or nowhere. Happy to add the other five to this PR if you want them — it's the same line each time and I'd add the same table-test case per command. Otherwise I'll leave this one scoped to #1389.

pkg/commands/secretstore/list.go
```diff
@@ -45,7 +45,10 @@ func (c *ListCommand) Exec(in io.Reader, out io.Writer) error {
 		return fsterr.ErrInvalidVerboseJSONCombo
 	}
 
-	var data []fastly.SecretStore
+	// Initialised rather than declared nil so that an account with no secret
+	// stores encodes as `[]` instead of `null`: encoding/json writes a nil
+	// slice as null, and `--json` output should be a list either way.
```
jedisct1, line 50: This comment looks AI-generated and unnecessary. Could you remove it?

PR #1901, opened 2026-09-03
[CDTOOL-1682] Add Support for ARC endpoints

rcaril:
### Change summary

This PR adds support for AI Runtime Control operations. 

All Submissions:

### New Feature Submissions:

### Changes to Core Features:


pkg/commands/airuntimecontrol/providerconnection/create.go
```diff
@@ -0,0 +1,83 @@
...
+			Globals: g,
+		},
+	}
+	c.CmdClause = parent.Command("create", "Create a provider connection").Alias("add")
+
+	// Required.
+	c.CmdClause.Flag("name", "Human-readable name of the provider").Required().StringVar(&c.name)
+	c.CmdClause.Flag("models", "Comma-separated list of allowed AI model identifiers").Required().StringsVar(&c.models, kingpin.Separator(","))
+	c.CmdClause.Flag("base-url", "Base URL for the provider's API").Required().StringVar(&c.baseURL)
+	c.CmdClause.Flag("api-key", "Provider's secret key for authentication (falls back to FASTLY_ARC_API_KEY)").Action(c.apiKey.Set).StringVar(&c.apiKey.Value)
```
kpfleming: What's the motivation for the environment variable fallback here? It's only going to be used once, in the `create` command, so providing it in a variable instead of the command line doesn't provide a reusability benefit.

pkg/commands/airuntimecontrol/providerconnection/update.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	c.CmdClause = parent.Command("update", "Update a provider connection")
+
+	// Required.
+	c.CmdClause.Flag("id", "Alphanumeric string identifying the provider connection").Required().StringVar(&c.id)
+
+	// Optional.
+	c.CmdClause.Flag("models", "An updated comma-separated list of allowed AI model identifiers").StringsVar(&c.models, kingpin.Separator(","))
+	c.CmdClause.Flag("base-url", "An updated base URL for the provider's API").Action(c.baseURL.Set).StringVar(&c.baseURL.Value)
+	c.CmdClause.Flag("api-key", "An updated provider secret key for authentication").Action(c.apiKey.Set).StringVar(&c.apiKey.Value)
```
kpfleming, line 48: This has an environment variable fallback in `create`, but not here, so if we keep that fallback it will need to be implemented here too for consistency.

kpfleming (changes requested):
This looks fantastic! Very nice work. I only have one pair of questions, related to the environment variable fallback for provider API keys.

pkg/commands/airuntimecontrol/providerconnection/create.go
rcaril: My understanding is that making a provider connection is per model. So one may find themselves using this command pretty often, especially as newer models are releaesd periodically.

kpfleming: So during the initial setup there may be a small batch of `create` commands, but after that there will be just one or two on an occasional basis. Given that, putting the key into an environment variable just so it can be used by one or two commands doesn't seem to be a benefit. Persons managing provider connections are not likely (I think) to just put that key into their environment and leave it there all the time.

I'd prefer to drop the environment variable fallback (which means quite a number of small changes in this PR), and revisit it if we get feedback from users that having to supply the API key on the command line is a burden for them.

rcaril: Fair enough! Removed in [eefecde](https://github.com/fastly/cli/pull/1901/commits/eefecde9f970df0f635a01f8e942cd66fbd9c2ae).

pkg/commands/airuntimecontrol/providerconnection/update.go
rcaril: Removed env support in [eefecde](https://github.com/fastly/cli/pull/1901/commits/eefecde9f970df0f635a01f8e942cd66fbd9c2ae) as per https://github.com/fastly/cli/pull/1901#discussion_r3927392013.

kpfleming (approved):

PR #1903, opened 2026-09-05
fix(text): prevent uint underflow in WrapIndent when limit is less than or equal to indent

Tyagiquamar:
This PR prevents a potential uint underflow in WrapIndent when limit <= indent.


kpfleming:
This looks like a reasonable fix but the tests are failing :-)

Tyagiquamar:
> This looks like a reasonable fix but the tests are failing :-)

Thanks for catching that. The failing cases were in the new test expectations: `wordwrap.WrapString` keeps `hello world` on one line when the fallback width is `1`. Updated the expectations in `27a36486`; the focused test should now match the implementation.

kpfleming (approved):

kpfleming:
Thanks for the contribution!

PR #1886, opened 2026-08-14
Remove dynamic build of configuration file

kpfleming:
### Change summary

Commit 468c63b39557ab22583b76dd4169446a6cdfe10c changed the CLI to use the compute-starter-kits.fastly.dev service to obtain starter kit information at runtime (when 'fastly compute init' is executed), so there is no need to have any build- or release-time construction of a static file.

All Submissions:

* [X] Have you followed the guidelines in our Contributing document?
* [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/cli/pulls) for the same update/change?

PR #1909, opened 2026-09-14
Remove Dangling Compute Starter Kit Related Make Config Code

rcaril:
### Change summary

This PR is a follow up to #1886, which was intended on removing the dynamic build of configuration files relating to compute starter kits. The remaining `make config` call remained, resulting in errors when a new tag action was run:

https://github.com/fastly/cli/actions/runs/34853540122/job/104007188117

All Submissions:


.github/workflows/tag_to_draft_release.yml
```diff
@@ -27,14 +27,6 @@ jobs:
         with:
           go-version: '1.26.x'
           cache: false
-      - name: "Install Rust"
```
rcaril, line 30: This is not called anywhere else and seemed to be only used for the Compute starter kit related actions.

anthony-gomez-fastly (approved):

